PR #13, opened 2025-06-22
Use the `blobUrlTemplate` value provided on the editor element

flavorjones:
Use the `blobUrlTemplate` value provided on the editor element instead of hardcoding it. This will allow the use of Active Storage path variants, including Basecamp-style account slugs.


jorgemanrubia:
Will merge to avoid merge conflicts (I'm modifying this very same method in [other PR](https://github.com/basecamp/actiontext-lexical/pull/14))

jorgemanrubia:
Merged but still not shipped, as I need to fix something related to attachments first. I'll keep you posted

PR #16, opened 2025-06-23
Add empty class to the list item instead of the list

andyra:
A tiny suggestion: @jorgemanrubia, does it work for you if we place the empty class on the item instead of the list? That's the thing we're actually styling, so it would be a little more direct on the CSS side.


jorgemanrubia:
@andyra this is updated in fizzy/main (not deployed)

andyra:
Beautiful, thanks!

PR #26, opened 2025-07-14
Support for unfurling links

flavorjones:
On the editor element, set a `data-unfurl-url` attribute to a path or URL that can be used to try to unfurl pasted URLs.

---

This is rough and doesn't have tests yet, I'm looking for feedback.

Open questions about the existing implementation:

- Is using `data-unfurl-url` acceptable? If not, please let me know what you want the API to look like!
- Is there a better way to traverse the nodes in the editor to find the node(s) to replace? What I have works but looks ugly and probably also is too liberal in how it find nodes.
- Placing the cursor after the updated text node seems to only be necessary if the node is the last node in the editor; what the code is doing right now is racy if someone is typing when the unfurling happens. Is there a better way?
- I'm not sure whether I should be wrapping the unfurling in an async/await or if what I'm doing with promises is fine. It feels awkward but my JS chops are still not good enough to discern how to better structure it.

Code that needs to be written before this should go into production

- The unfurl fetch should only happen for a small set of allowlists URLs, which the RichLink gem can provide. I'm thinking about another attribute, `data-unfurl-allowlist` which declares an endpoint that returns an array of strings or regexes that must be matched before the request is made.
- Only a simple title is supported at the moment, if we want better embeds that's going to be more work.
- Tests.


src/elements/editor.js
```diff
@@ -83,6 +83,10 @@ export default class LexicalEditorElement extends HTMLElement {
     return this.getAttribute("attachments") !== "false"
   }
 
+  get unfurlUrl() {
+    return this.dataset.unfurlUrl
```
jorgemanrubia, line 87: since we own the element, I think we could use a property like `unfurl-url` without the `data-*` prefix.

src/editor/clipboard.js
```diff
@@ -33,10 +33,27 @@ export default class Clipboard {
...
+          this.#fetchUnfurl(this.editorElement.unfurlUrl, text)
+            .then(response => response.json())
+            .then(data => this.contents.unfurlLink(text, data))
+        }
       }
     })
   }
 
+  #fetchUnfurl(unfurlUrl, linkUrl) {
+    return fetch(unfurlUrl, {
```
jorgemanrubia, line 46: The API makes sense but I wonder if it would make sense to design the API so that the server can reply with an arbitrary HTML fragment that will replace the link. That could serve for these links and, also, to render embedded previews of other links (e.g: youtube), images, etc. or just any arbitrary HTML. We don't need to support these now, but I think it would be a nice flexibility to have.

For the case of unfurlink links, this approach allows the server to render the link with the title and URL directly. The response could be a regular HTML response, no need for JSON.

What do you think? I think the API you proposed makes sense too.

src/editor/clipboard.js
```diff
@@ -33,10 +33,27 @@ export default class Clipboard {
         this.contents.createLinkWithSelectedText(text)
       } else {
         this.#pasteMarkdown(text)
+        if (isUrl(text) && this.editorElement.unfurlUrl) {
```
jorgemanrubia, line 36: After pasting the markdown, if it contains a link, you will have a link node with the pasted URL. What do you think of doing something like:

```js
...
const html = await response.text()
this.contents.replacelLinkBeforeCursorWith(newHtml)
```

`unfurlLinkBeforeCursorWith` would act on the link introduced right before where the cursor is. I have some methods like that in contents, and the code is quite verbose and awkard 😅.

O3 suggests this code that I haven't tested:

```js
/**
 * Replace the <a> element that is immediately before (or under) the caret
 * with the DOM represented by `html`.  Does nothing if no link is found.
 *
 * @param {string} html – raw HTML string to insert
 */
replaceLinkBeforeCursorWith(html) {
  this.editor.update(() => {
    const selection = $getSelection();
    if (!selection || !selection.isCollapsed()) return;

    // ────── helper: walk up to the nearest LinkNode ──────
    const findEnclosingLink = (node) => {
      while (node && !$isLinkNode(node)) node = node.getParent();
      return $isLinkNode(node) ? node : null;
    };

    let anchorNode   = selection.anchor.getNode();
    let linkNode     = findEnclosingLink(anchorNode);

    // If the caret is not inside a link, look at previous siblings,
    // skipping back until we bump into a LinkNode or run out of nodes.
    for (let cur = anchorNode; !linkNode && cur; cur = cur.getPreviousSibling()) {
      linkNode = findEnclosingLink(cur);
    }
    if (!linkNode) return;                 // nothing to replace

    // Turn the HTML string into Lexical nodes
    const newNodes = $generateNodesFromDOM(this.editor, parseHtml(html));
    if (newNodes.length === 0) return;

    // Replace the link with the first new node, chain-insert the rest
    let prev = newNodes[0];
    linkNode.replace(prev);                // swap in the first node
    for (let i = 1; i < newNodes.length; i++) {
      prev.insertAfter(newNodes[i]);
      prev = newNodes[i];
    }

    // Keep the caret right after the inserted content
    prev.select();
  });
}
```

I think it makes sense to do something more predictive here (target the last link you know exists) instead of reprorcessing all the links in the editor on each operation.

Regarding allow-listing, I think that's great for links to other hosts. For links to the same hosts I think we could allow by default. For our use case I think we can rely on "same hosts only", and leave the setting for external hosts as a future improvement.

jorgemanrubia:
Also forgot to mention that, for testing:

- There is a functional "dummy" app in `/test/dummy` that you can extend to test this manually with an endpoint or whatever.
- I'm using capybara system tests (versus JS unit tests). I think we could extend those to test this too.

flavorjones:
Closing this. Thanks for the feedback @jorgemanrubia, I will probably come back with an improved PR.

PR #46, opened 2025-09-02
Prepare official release+css polish

andyra:
This adds some CSS polish to the basic CSS included with Lexxy.

I didn't go too crazy with adding variables for customization. To allow full UI customization entirely through custom properties, we'd have to add hundreds of CSS variables. Instead, I opted for using very low specificity (via the [:where() pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/:where)), to make it easy to override.

I also opted out of adding dark mode styles. It seemed like a nice thing to include at first until I realized that anyone who has an existing color system will likely want to hook that in here. Much easier to do that if they don't have to also override dark mode colors as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ By default, the partial to render the attachment will be looked up in `app/views
 <em><%= person.name %></em> (<strong><%= person.initials %></strong>)
 ```
 
-On the editor side, let's start with the *inline* approach:
+On the editor side, let's start with the *inline* approach by rendering all the prompt items inside the `<lexxy-prompt>` element:
 
 ```erb
 <%= form.rich_text_area :body do %>


app/assets/stylesheets/lexxy-content.css
```diff
@@ -1,258 +1,205 @@
...
+  color: var(--lexxy-color-text-subtle);
   font-size: var(--lexxy-text-small);
-  flex-wrap: wrap;
-  line-height: 1.3;
-  column-gap: 0.5ch;
-  justify-content: center;
-  margin: var(--lexxy-block-space-half) var(--lexxy-inline-space);
-  text-align: center;
 
   .input {
```
jzimdars: Should use just the plain `input` element here instead of Fizzy's `.input` class. Easy oversight.

app/assets/stylesheets/lexxy-editor.css
```diff
@@ -1,135 +1,94 @@
...
-    background-color: var(--lexxy-color-canvas);
-    border: 1px solid var(--lexxy-color-ink-lighter);
-    border-radius: 0.5em;
-    box-shadow: var(--lexxy-shadow);
-    color: var(--lexxy-color-ink);
-    padding: var(--lexxy-block-space) calc(var(--lexxy-inline-space) * 1.5);
-    position: absolute;
-    z-index: 1;
-
-    .input[type="url"] {
```
jzimdars, line 114: Should use just the plain `input` element here instead of Fizzy's `.input` class.

app/assets/stylesheets/lexxy-editor.css
```diff
@@ -155,40 +109,85 @@ lexxy-toolbar {
...
+
+  .lexxy-dialog-actions {
+    display: flex;
+    font-size: var(--lexxy-text-small);
+    flex: 1 1 0%;
+    gap: 1ch;
+    margin-block-start: 1ch;
+  }
+
+  .input[type="url"],
```
jzimdars: Should use just the plain `input` element here instead of Fizzy's `.input` class.

app/assets/stylesheets/lexxy-editor.css
```diff
@@ -155,40 +109,85 @@ lexxy-toolbar {
...
+  button {
+    line-height: 1.5lh;
+    padding-inline: 1ch;
+  }
+
+  .input[type="url"] {
```
jzimdars: Should use just the plain `input` element here instead of Fizzy's `.input` class.

PR #56, opened 2025-09-04
Fix README example for lexxy-prompt trigger example

jespr:
<lexy-prompt> should be defined inside of form.rich_text_area :body block


jorgemanrubia (approved):
Thanks

PR #65, opened 2025-09-05
Set aria attributes properly

brunoprietog:
There are currently some issues with the way aria attributes are set:
- [Here](https://github.com/basecamp/lexxy/blob/461bfd58f017efb629353704d6c3d6bd10ca4967/src/elements/editor.js#L31), the textbox role is being set using internals, but the result of this is that the wrapper is now interpreted as an editable field, which is incorrect. The actual editable field is the one created afterwards.
- This means that if the element has an associated label, at least in NVDA on Windows, when passing through the wrapper using browse mode, the screen reader attempts to read the label and succeeds, since internals automatically associates them, but because it has a textbox role and is not contenteditable, it fails to read the content of the field located in the wrapper's child. In other words, it does not read the content of the editor.
- Likewise, since `lexxy-editor` is merely a container, setting aria attributes does not produce any of the expected effects on the actual contenteditable.

So the idea is:
- Set the base role of the `lexxy-editor` wrapper element to `presentation`, so that the accessibility tree ignores it and simply continues to the children. Instead, the role `textbox` goes in the contenteditable, as well as the `aria-multiline` attribute that was missing.
- This makes it possible to set aria attributes in `lexxy-editor` and have the browser ignore them, so that we can simply copy the attributes to the contenteditable without having to delete them.
- Set a default `aria-label` with the content of the associated label text, so that the screen reader reads the label content when passing through the contenteditable. If `lexxy-editor` has an `aria-label`, it takes priority over the label content, as is the case with a native input.
- Additionally, I am adding `toolbar` as the base role to the `lexxy-toolbar` element.


jorgemanrubia:
Amazing @brunoprietog, thank you so much!

PR #66, opened 2025-09-05
Update README.md

joeinnes:
https://37signals.com/02


jorgemanrubia:
Thanks @joeinnes. No problem with *battle* here from my side.

joeinnes:
Ok!

PR #69, opened 2025-09-05
Lexxy is visible when toolbar is disabled

zoltanhosszu:
When toolbar is set to false, Lexxy was displayed with 0 opacity. This PR removes that code.


zoltanhosszu:
#69 nice 😂

PR #103, opened 2025-09-05
Allow toolbar to scroll on the x-axis

### LINKED ISSUE #63: lexxy-toolbar mobile responsiveness
The toolbar is not currently mobile responsive. Where I'm using lexxy, I added `overflow-x: scroll` as a temporary solution but that may not be what the maintainers have in mind.

Image

andyra:
Allows toolbar to scroll on the x-axis when there's not enough room for the buttons.

@jorgemanrubia, this is a functional improvement, but I think we can do better here by using JS to tuck buttons into an overflow menu when there's not enough room for them. I'll take a look!

Fixes #63


andyra:
Actually, maybe hold on a review for the time being, @jorgemanrubia. Looks like overflow isn't a great solution since it clips the link dialog.

andyra:
@jorgemanrubia, I just pushed up a dynamic JS version that folds buttons that don't fit into an overflow menu. Right now it only happens on init, but it would probably be nice to listen for window resize events. Feel free to rearrange things as you see fit!

https://github.com/user-attachments/assets/0a3a49d2-80e3-4d52-ad4b-ab51eedbe918

jorgemanrubia:
Thanks @andyra 🙏, I moved the logic to the "toolbar" class and also made it trigger it on window resizing.

Issue #59, opened 2025-09-04
Lexxy events


lylo:
I figure this will be planned, but I thought I'd create an issue anyway while I was trying out lexxy. 

I have a number of customisations to Trix in my app which I'm using Trix events for, e.g. `trix-initialize` and `trix-change` (autosaving, for example), so it would be great to have `lexxy-change` etc equivalents. I also make use of `trix-attachment-add` and `trix-file-accept` to prevent drag and drop, plus limit to specific file types.

jorgemanrubia:
`lexxy:change` is already in place, just not documented. I aggree that we need more standard events in place!

UPDATE: Added note to the README

jorgemanrubia:
I'll close this one. I think `lexxy:change` is what will be needed 99% of the time. We should also add an event for canceling attachments but we can add those down the road as an specific issue, instead of keeping a generic "events" one hanging.

maprinz:
> I think lexxy:change is what will be needed 99% of the time

I'd give it a 75% 😃 I think there are several use cases, where you want to listen to certain editor events. For me, I'd like to build a Notion-like floating toolbar that pops up once a text is selected (and hides when text is deselected). So the not documented `lexxy:node-selected` event might be something that people are interested in?

Keep up the great work ❤️

lylo:
@jorgemanrubia I have some real-world use cases for other events. I can try and work on these if you'd be amenable? Let me explain what I was using the Trix ones for:

* `trix-file-accept`: I used this to restrict file types. I only want images or videos so I'd check the `file.type` value of the event. Perhaps Lexxy has another way to do this?
* `trix-initialize`: I use this to detect when the editor has loaded so I can load a draft from local storage if something has changed in there.

I think these are legit, no?

jorgemanrubia:
Yes you are right. I've reopened this! Thanks for the feedback!

MSILycanthropy:
Taking a bit of a lay of the land here to make this a bit easier to digest. Please correct me if I'm missing anything.

**Currently Existing Events**
`lexxy:change`:  Fired when there's any change to the content of the editor

The next few are related to the internals of node selection. Which is currently only related to attachments. Should these be exposed to users, or somehow marked as internal for now?
`lexxy:move-to-next-line`: Fired when a user moves to the next line from a selected node
`lexxy:node-selected`: Fired when a user selected a selectable node
[...]

PR #106, opened 2025-09-06
Improve README and events doc with Stimulus integration examples

AhmedNadar:
- Adds events doc cover 4 lexxy events and addressing #59
- Add Stimulus controller examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,24 +1,48 @@
 # Lexxy
 
-A modern rich text editor for Rails.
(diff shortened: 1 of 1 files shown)


jorgemanrubia:
I appreciate the work @AhmedNadar, but this PR is too large and it includes many changes to the README I don't want to include. 

Re: events, the only public event is `lexxy:change` for now. I'd be happy to merge a PR documenting only that if you want. Thanks 🙏.

PR #182, opened 2025-09-08
Add details to lexxy:change event documentation

AhmedNadar:
Document when the event fires (typing, pasting, deleting, etc.)


jorgemanrubia:
Thanks @AhmedNadar. I prefer not to include stimulus examples in general, since both are separated libraries. I think just documenting the event is enough here.

Issue #172, opened 2025-09-07
Couple Icon issues


MSILycanthropy:
**Issue Observed**
Booted up the sandbox to play around a bit and noticed a handful of the icons are slightly cutoff. Namely the Quote and Code icons. This seems to not be the case in any other content I've seen about Lexxy.

Key to note, only seeing this on Windows Chrome (139.0.7258.157). Tested on my Mac and the icons display just fine. Potentially an issue with the svg viewbox? 

The link icon seems to be fully missing on both devices as well.

**Screenshots**
<img width="1248" height="327" alt="Image" src="https://github.com/user-attachments/assets/8beb6df4-b6f7-4cc4-a3e3-8f03ccbe2e33" />

jorgemanrubia:
Thanks for the report @MSILycanthropy 
cc @andyra

andyra:
Good instinct, @MSILycanthropy; looks like the SVG viewboxes are a couple pixels off for a couple icons. I'll get that patched up. Thanks for the keen eye!

PR #183, opened 2025-09-08
Fix link button SVG being invisible

excid3:
The link button's SVG isn't visible because the selector only targets `button[data-command]` and `summary` elements. 

This adds `button[data-dialog-target]` as a selector which fixes the issue. It should also be helpful in the future if other buttons open dialogs.

Reported in #172


jorgemanrubia:
Hey thanks @excid3 🙏. @andyra addressed this issue and more on https://github.com/basecamp/lexxy/pull/185

PR #185, opened 2025-09-08
Toolbar icon polish

### LINKED ISSUE #172: Couple Icon issues

andyra:
- Fix missing icon link
- Redraw icons
  - Equal size viewboxes to avoid clipping
  - More visually balanced
- Add classes to buttons so we don't need multiple selectors to style similar elements
- Remove extra space caused by `lexxy-link-dialog` element

|Before|After|
|--|--|
|CleanShot 2025-09-08 at 15 08 27@2x|CleanShot 2025-09-08 at 15 07 58@2x|

Fixes #172


eserna27:
🥳

PR #190, opened 2025-09-08
Rename internal events and add initialization and file acceptance events

MSILycanthropy:
Closes https://github.com/basecamp/lexxy/issues/59

- Refactors internal events to use the `lexxy:internal` prefix
- Fire `lexxy:initialized` in `connectedCallback`
- Fire cancelable `lexxy:file-accept` event before uploading files


src/elements/editor.js
```diff
@@ -40,6 +40,8 @@ export default class LexicalEditorElement extends HTMLElement {
 
     CommandDispatcher.configureFor(this)
     this.#initialize()
+
+    requestAnimationFrame(() => dispatch(this, "lexxy:initialized"))
```
MSILycanthropy: There's one small-ish caveat with this in a Stimulus-centered world. Since web components initialize very early on, it's likely that Stimulus controllers won't have connected by the time this event is emitted.

That's at least what I ran into when looking into writing a quick test for this using the `event_log_controller`.

Might warrant some documentation, or another solution if there is one.

brunoprietog:
How about just calling it `lexxy:initialize`? Events shouldn't have verbs in the past tense.

brunoprietog:
The same applies to other events.

jorgemanrubia (approved):
Fantastic work @MSILycanthropy! Thanks 🙏

jorgemanrubia:
Could you please rebase and run `yarn build` and push the assets to resolve the conflicts.

test/system/attachments_test.rb
```diff
@@ -53,4 +53,24 @@ class AttachmentsTest < ApplicationSystemTestCase
 
     assert_no_attachment content_type: "image/png"
   end
+
+  test "allows valid attachments" do
```
jorgemanrubia: Could you move these two tests to events_test.rb and include `lexxy:file-accept` in the title? E.g: `use lexxy:file-accept to allow valid attachments`

PR #195, opened 2025-09-10
Fix issue with frozen nodes in dev with esbuild

### LINKED ISSUE #175: Issue with active storage uploads.
Hello and thanks for the great work.

We are trying the new editor in one of our internal rails 8 projects. We are having some trouble uploading attachments though.

In the web console we see the following error:

`TypeError: Attempted to assign to readonly property. triggered here:`

https://github.com/basecamp/lexxy/blob/b5d385efe4a0ea6d7a50f052a8c5ceb74678f80c/src/nodes/action_text_attachment_upload_node.js#L108

We are using js-bunding(esbuild)

MSILycanthropy:
Fixes #175 

Lexical apparently freezes nodes in development so you can catch unwanted mutations that happen outside of `editor.update` that might cause unwanted behavior. The main issue comes from this line in their source code [here](https://github.com/facebook/lexical/blob/main/packages/lexical/src/LexicalReconciler.ts#L254).

This PR just moves calculating the `src` property to the `replace` call a few lines later so the bad mutation is avoided and everything happens as lexical expects.


jorgemanrubia (approved):
Oh nice one @MSILycanthropy! I'm curious why I can't reproduce, but the patch makes sense. Thanks!

PR #204, opened 2025-09-12
Use collection partial rendering

KonnorRogers:
was poking through how Lexxy handles prompts and noticed it using loop rendering instead of collection rendering for the example.

I know its "just an example", but probably good to steer people away from rendering in loops especially on something like `<Model>.all`


jorgemanrubia:
Yes, that is better! Thanks @KonnorRogers

PR #206, opened 2025-09-12
Toggle link instead of replacing selection

tonysm:
### Changed

- Use `$toggleLink` instead of replacing the selection with a link ([docs](https://lexical.dev/docs/api/modules/lexical_link#togglelink-1))

---

The previous code was replacing the selected text with a new link node, but this removes any other inline markup decoration such as bold or italic.

Instead of doing that, we could use the `$toggleLink` helper from Lexical, which will keep the other markup in place, so we can combine bold, italic and links.

---

### Markup difference

**Before:**

Before pasting the link:

```html
<p>
  Hello <b><strong>World</strong></b>
</p>
```

After pasting the link:

```html
<p>Hello <a href="http://localhost:3000/">World</a></p>
```

**After:**

Before pasting the link:

```html
<p>
  Hello <b><strong>World</strong></b>
</p>
```

After pasting the link:

```html
<p>
  Hello
  <a href="http://localhost:3000/"
    ><b><strong>World</strong></b></a
  >
</p>
```


jorgemanrubia:
Hey the test is failing @tonysm, is it working for you locally?

tonysm:
@jorgemanrubia I think it's because we need to run `yarn build`. I didn't run it because I had other PRs, so there will be conflicts. Let me run it.

PR #207, opened 2025-09-12
Fix bold and italic not working together

### LINKED ISSUE #201: Bold + Italic not working?
If you select text and mark it as both bold and italic it only appears as bold. Here's a screenshot of me marking text as bold and italic. You can see that both toolbar buttons are pressed so Lexxy appears to know that B + I is active.

Image

However the text in the editor is only bold. Here's the markup that is generated:

```
<p dir="ltr">
  <strong data-lexical-text="true">This</strong>
  <span data-lexical-text="true"> is a </span>
  <em data-lexical-text="true">test</em>
  <span data-lexical-text="true">.</span>
</p>
```

I've tried this in my app as well as a basic sandbox and it's apparent in both.

tonysm:
### Changed

- Allows combining bold, italic, and underline

---

closes #201 

I found this solution in https://github.com/facebook/lexical/discussions/4381 

However, I don't know about backward compatibility with the markup generated by Trix... I think it's fine since Lexical will add the classes when ingesting the content before displaying. I tested in the dummy app creating a post before the changes, then applying the changes and editing the post, it seems to work for me.

---

<details>
<summary>Markup difference</summary>

I'm using screenshots because there's a difference between the markup in the Sandbox Output element and the one generated inside the editor.

**Before:**

image

As you can see, the output box has an `<i>` that has `<b>` and `<strong>` inside, but the actual editor markup doesn't have those.

**After:**

image

When adding the custom classes, the editor HTML is the same, but now it uses the defined classes so the styles apply. The classes are only added to the editor HTML and are not part of the output HTML.

</details>


test/test_helpers/editor_handler.rb
```diff
@@ -74,6 +74,14 @@ def within_contents(&block)
     page.within content_element, &block
   end
 
+  def toggle_command(command, toolbar_selector = "lexxy-toolbar")
+    find("#{toolbar_selector} [data-command=\"#{command}\"]").click
+  end
+
+  def content
```
jorgemanrubia: could you please rename this method to `html`? *content* is used for other editor parts and would like to distinguish

jorgemanrubia:
Test failing for this one too

test/test_helpers/editor_handler.rb
tonysm: I tried `html` but it failed for some reason I didn't quite understand. I renamed it to `inner_html` and it works. What do you think?

tonysm:
> Test failing for this one too

Fixed the other bold/italic tests (they also needed the classes now) and compiled the assets. All tests are passing for me locally.

test/test_helpers/editor_handler.rb
jorgemanrubia: That is good, thanks!

jorgemanrubia:
Hey @tonysm I will merge this but I want to play a bit with it first. I would like to not having to serve the class attributes as part of the value. Would love to keep those as a mere internal editor thing.

tonysm:
@jorgemanrubia you mean to keep the CSS classes as only part of the editor, but not as part of the final render?

abuisman:
It would be nice if the resulting html would be something like: `Some<strong>test<em>italic</em>strong</strong>`. 

If you go to the [lexical playground](https://playground.lexical.dev/) you can see that it also splits text up into pieces with identical styling.

tonysm:
@jorgemanrubia @abuisman I've updated the code so now we're removing any inline styles or CSS classes from the exported HTML.

image

Here's the docs about the [HTML export config](https://lexical.dev/docs/concepts/serialization#html-property-for-import-and-export-configuration) and [there's a playground linked there with a similar implementation](https://stackblitz.com/github/facebook/lexical/tree/main/examples/react-rich?embed=1&file=src%2FApp.tsx&terminalHeight=0&ctl=1&showSidebar=0&devtoolsheight=0&view=preview), so I based this on that one.

The result is that those CSS classes will only be used inside the editor and not part of the output HTML.

jorgemanrubia:
Thanks @tonysm. [We are using DOMPurify to sanitize the HTML](https://github.com/basecamp/lexxy/blob/10595cbf2ad054b8a9a35cb453c2eba7fbca46b5/src/helpers/html_helper.js#L57-L65). I would try to configure it to do this here.

I would try to do something targeted. `style` attributes are already stripped out, I would remove `class` attributes only for these nodes, not in general. ChatGPT suggests something like this could work:

```js
DOMPurify.addHook("uponSanitizeElement", (node, data) => {
  if (data.tagName === "strong" || data.tagName === "em") {
    node.removeAttribute('class');
  }
});
```

We could place that code in `src/config/dom_purify.js`, and require the file from `index.js`. What do you think?

tonysm:
@jorgemanrubia Ahh, via DOM Purify makes sense too. Okay, done. 👍🏼

jorgemanrubia (approved):
Wonderful @tonysm, thank you so much! I'll release a new version with all these shortly!

PR #208, opened 2025-09-13
Remember the initial value and handle resetting the form state

tonysm:
### Changed

- Remembers the initial value and implements the `formResetCallback` by resetting the editor state with whatever set in the initial value


src/elements/editor.js
```diff
@@ -25,6 +25,8 @@ export default class LexicalEditorElement extends HTMLElement {
 
   static observedAttributes = [ "connected" ]
 
+  #initialValue = ''
```
jorgemanrubia: Could you please use double " for strings here

I should totally configure eslint here, but I haven't yet 🙏!

tonysm: done!

PR #217, opened 2025-09-17
Make ApplicationMailer work without ActionMailer dependency

eelcoj:
This aims to solve the issue Lexxy currently has with Rails apps without ActionMailer. 

```
NameError: uninitialized constant Actiontext::Lexical::ActionMailer (NameError)

    class ApplicationMailer < ActionMailer::Base
                              ^^^^^^^^^^^^
Did you mean?  ActionCable
```

(Fixed it, but currently unsure where/how ActionMailer would be used).


jorgemanrubia:
Oh we could totally remove that! Could you just adapt the PR to nix the application mailer completely? That was just generated automatically when I bootstrapped the gem with the engine skeleton.

eelcoj:
Even better! ✌️

PR #220, opened 2025-09-19
Replace marked library with Lexical's native markdown parser

robzolkos:
## Summary

Replaces the external `marked` library with Lexical's built-in `@lexical/markdown` parser for clipboard markdown handling. This change improves performance and provides better consistency with the editor's existing markdown functionality.
  
## Changes Made

- **Updated clipboard.js**: Replace `marked()` with `$convertFromMarkdownString()` using Lexical's native transformers
- **Removed dependency**: Eliminated `marked` from package.json and yarn.lock
- **Rebuilt assets**: Updated compiled JavaScript files to reflect the new implementation

## Benefits

- **Reduced bundle size** by removing external markdown parser dependency
- **Better performance** with native Lexical integration instead of HTML conversion
- **Improved consistency** with existing markdown shortcuts already using Lexical transformers
- **Simplified dependency tree** by leveraging existing `@lexical/markdown` package

## Testing

The functionality remains identical for users - markdown content pasted into the editor will continue to work exactly as before, but now uses the more robust Lexical parser.

Test cases:

- [x] Basic markdown formatting (bold, italic, headers)
- [x] Lists and nested lists
- [x] Code blocks and inline code
- [x] Links and blockquotes
- [x] Bundle builds successfully
- [x] No breaking changes to existing functionality

## Verification

- No references to `marked` library remain in built assets
- `pasteMarkdown` function now uses `editor.update()` with Lexical transformers
- All existing markdown shortcuts continue to work as expected

## Before / After

Rendering and generated HTML of pasted markdown is identical

### Before
before

### After
after

### Comparison with Lexical Playground
image

Note the removal of duplicate bullets - this is a css concern - Lexical playground injects css classes into li elements that have child ul to assist with removing duplicate bullets.  That is out of scope for this PR.


jorgemanrubia (approved):
I love this change from every angle. It fixes bug, simplifies and removes dependencies. Thank so much @robzolkos 👏!!!

src/editor/clipboard.js
```diff
@@ -40,8 +40,9 @@ export default class Clipboard {
   }
 
   #pasteMarkdown(text) {
-    const html = marked(text)
-    this.contents.insertHtml(html)
+    this.editor.update(() => {
+      $convertFromMarkdownString(text, TRANSFORMERS)
+    })
```
jorgemanrubia, line 45: Hm, I just noticed this after merging. Update creates an update context for content-changing operations, but I don't think this line is modifying the content at all. I'll confirm separately.

jorgemanrubia: My mistake: it works as intended. I thought that function was just returning the transformed string, but it is handling the insertion as expected, so that is great!

PR #239, opened 2025-09-25
Fix: Nil selection list prevents further editing

Tonksthebear:
On a fresh page load, clicking either numbered or bullet list **before** clicking on the editor causes an error due to selection being nil in the following code

```js
    const anchorNode = selection.anchor.getNode()
```

It seems to mess up the state of lexxy because toolbar usage after the error results in no changes.

Also, even though system tests seem properly set up to raise errors on js errors, it wasn't actually throwing a test failure on my end. For instance, I can get a test to fail by unning

```ruby
page.execute_script("throw new Error('Deliberate test error')")
```

but the console error being thrown by the mentioned bug is not failing the test.

I added a helper method that asserts no console messages whatsoever and that is correctly picking up on the bug.

Let me know if you want me to break up the tests and/or if you don't want to rely on the test helper.


jorgemanrubia (approved):
Thank so much @Tonksthebear 🙏

test/application_system_test_case.rb
```diff
@@ -9,6 +9,13 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
...
 
   teardown do
+    assert_no_console_messages
     # Add any teardown code needed for all system tests
   end
+
+  def assert_no_console_messages(level = nil)
+    logs = page.driver.browser.logs.get(:browser)
+    logs = logs.select { |log| log.level == level } if level
+    assert logs.empty?, "Unexpected console messages: \n" + logs.map(&:message).join("\n\n")
```
jorgemanrubia, line 19: Good idea 👍

PR #281, opened 2025-09-29
Expose link created events

monorkin:
This PR adds a new `lexxy:link-created` event that fires whenever someone pastes a link.  
With this event, you can build things like link unfurling, previews, or embeds.

Example:
```erb
<%= form.rich_textarea :content, 
          data: { 
            controller: "unfurl-link",
            unfurl_link_url_value: unfurl_link_path,
            action: "lexxy:link-created->unfurl-link#unfurl" } %>
```
```javascript
import { Controller } from "@hotwired/stimulus";
import { post } from "@rails/request.js"

export default class extends Controller {
  static values = {
    url: String,
  }

  unfurl(event) {
    this.#unfurlLink(event.detail.url, event.detail)
  }

  async #unfurlLink(url, callbacks) {
    post(
      this.urlValue,
      {
        body: JSON.stringify({ url }),
        headers: {
          "Content-Type": "application/json",
          "Accept": "application/json"
        }
      })
      .then(({ response }) => response.json())
      .then((data) => this.#insertUnfurledLink(data, callbacks))
  }

  #insertUnfurledLink(metadata, callbacks) {
    callbacks.replaceLinkWith(this.#renderUnfurledLinkHTML(metadata))
    // callbacks.insertBelowLink(this.#renderUnfurledLinkHTML(metadata), { attachment: true })
  }

  #renderUnfurledLinkHTML(metadata) {
    return `<a href="${metadata.canonical_url}">${metadata.title}</a>`
  }
}

```

### Event detail

The event provides

- `url`: The URL of the link that was created
- a set of callbacks for rendering unfurled content:
    - `replaceLinkWith(html, options)`: replaces the pasted link with the given HTML
    - `insertBelowLink(html, options)`: inserts the given HTML in a new line right below the link

Both callbacks accept an `options` object with:

- attachment: can be a boolean or an object; if it’s an object, it may include an sgid attribute.

Using an attachment lets you insert arbitrary HTML into the editor. Without an attachment, Lexical will sanitize the HTML (e.g. strip out DIVs and some attributes).

Example: Replacing a link with HTML

https://github.com/user-attachments/assets/0fcb89f9-391f-4302-8538-4589d6ede75a

Example: Inserting an attachment below a link

https://github.com/user-attachments/assets/d4a0c51b-cfc8-41c8-8442-9649f02c8e2a

---

I focused on the use-case of pasting a bare link, but this could be extended to all links by passing a context string like `paste`, `paste-with-selected-text` or `command`.


src/editor/clipboard.js
```diff
@@ -33,6 +34,9 @@ export default class Clipboard {
     item.getAsString((text) => {
       if (isUrl(text) && this.contents.hasSelectedText()) {
         this.contents.createLinkWithSelectedText(text)
+      } else if (isUrl(text)) {
+        const nodeKey = this.contents.createLink(text)
+        dispatchLinkEvent("created", this.editorElement, nodeKey, { url: text })
```
jorgemanrubia: Minor: I think this should be a regular private method in the clipboard, or maybe make the helper a generic `events_helper.js`. These helper tend to be generic in nature and cross-cutting.

src/editor/contents.js
```diff
@@ -256,6 +275,87 @@ export default class Contents {
...
+        const firstReplacementNode = htmlNodes[0]
+        node.replace(firstReplacementNode)
+
+        if (htmlNodes.length > 1) {
+          let previousNode = firstReplacementNode
+          htmlNodes.slice(1).forEach(htmlNode => {
+            previousNode.insertAfter(htmlNode)
+            previousNode = htmlNode
+          })
+        }
```
jorgemanrubia: I would consider extracting two methods for the two branches, to keep the original method easier to parse and understand. Also, I think there is some duplication you could get rid of: the node replacement and selection. Something like:

```js
const node = options.attachment ? this.#createCustomAttachmentNodeWithHtml(html) : this.#createHtmlNodeWith(html)
node.replace(customNode)
if (wasSelected) {
  node.selectEnd()
}
```

jorgemanrubia (commented):
Looks great @monorkin 👏

It would be good to document the new event in the readme. We should add entries for the callback methods you can invoke too with it. I would also include the example of the stimulus controller. As a minor thing, in the example you can tighten up things a bit with async/await:

```js
export default class extends Controller {
  static values = {
    url: String,
  }

  unfurl(event) {
    this.#unfurlLink(event.detail.url, event.detail)
  }

  async #unfurlLink(url, callbacks) {
    const { response } = await post(this.urlValue, {
      body: JSON.stringify({ url }),
      headers: {
        "Content-Type": "application/json",
        "Accept": "application/json"
      }
    })

    const link = await response.json()
    this.#insertUnfurledLink(link, callbacks)
  }

  #insertUnfurledLink(link, callbacks) {
    callbacks.replaceLinkWith(this.#renderUnfurledLinkHTML(link))
    // callbacks.insertBelowLink(this.#renderUnfurledLinkHTML(link), { attachment: true })
  }

  #renderUnfurledLinkHTML(link) {
    return `<a href="${link.canonical_url}">${link.title}</a>`
  }
}
```

src/editor/clipboard.js
monorkin: Agreed, changed it to a private method.

src/editor/contents.js
monorkin: You're right. I extracted the private methods in https://github.com/basecamp/lexxy/pull/281/commits/34f8df2d167a79ba4bcf22ee6ec525d2ef83cd5c

monorkin:
@jorgemanrubia thank you for the review :raised_hands: 
I updated the PR and expanded the events section of the readme in https://github.com/basecamp/lexxy/pull/281/commits/a440966d01550d74a6cf5ca3ca9e34db3f3a1f6c

brunoprietog:
Hi @monorkin! Events should be in present tense. How about calling the event `lexxy:autolink`?

jorgemanrubia (approved):
👏

jorgemanrubia:
+1 to change the event name as @brunoprietog suggests. Maybe `lexxy:insert-link` (since not necessariliy autolinking here)?

monorkin:
Agreed on the name change, it's `lexxy:insert-link` now.
Autolink reminds me of the AutoLink node, so I went with insert-link

app/assets/javascript/lexxy.js
```diff
@@ -7125,12 +7305,27 @@ class Clipboard {
...
+      } else if (isUrl(text)) {
+        const nodeKey = this.contents.createLink(text);
+        this.#dispatchLinkCreatedEvent(nodeKey, { url: text });
       } else {
         this.#pasteMarkdown(text);
       }
     });
   }
 
+  #dispatchLinkCreatedEvent(nodeKey, payload) {
```
brunoprietog, line 7318: It'd probably be worth changing it here too for consistency.

```suggestion
  #dispatchLinkInsertEvent(nodeKey, payload) {
```

src/editor/clipboard.js
```diff
@@ -33,12 +34,27 @@ export default class Clipboard {
...
+      } else if (isUrl(text)) {
+        const nodeKey = this.contents.createLink(text)
+        this.#dispatchLinkCreatedEvent(nodeKey, { url: text })
       } else {
         this.#pasteMarkdown(text)
       }
     })
   }
 
+  #dispatchLinkCreatedEvent(nodeKey, payload) {
```
brunoprietog: It'd probably be worth changing it here too for consistency:

```suggestion
  #dispatchLinkInsertEvent(nodeKey, payload) {
```

PR #285, opened 2025-09-30
Fix: pasting markdown will clear editor contents

### LINKED ISSUE #233: Pasting from another application replaces all Lexxy content
If you copy and paste text from within the Lexxy editor itself, it works. If you copy text (even plain text) from another application, pasting into Lexxy replaces all the content.

I've replicated this on Firefox, Chrome (Mac and Linux) as well as Safari. Here's a demo video.

https://github.com/user-attachments/assets/a42c4413-ae87-4b12-aad9-931c3294e720

jorgemanrubia:
E.g pasting a URL will clear the full editor contents.

Introduced in https://github.com/basecamp/lexxy/pull/220

@robzolkos I had to revert that one. Maybe we should apply the trasnformers manually to leverage Lexxy markdown conversion support, I couldn't find an API to convert markdown strings on the fly. For now, I will just revert.

fix #233


robzolkos:
Thanks @jorgemanrubia will take a look.

PR #293, opened 2025-09-30
Add support for `ActionText::Content`

romansklenar:
Refactored the HTML rendering logic for `ActionText` to support also `ActionText::Content` alongside `ActionText::RichText` without silently hiding errors.

This allows usage of `ActionText`'s content handling without the dedicated polymorphic tables by serializing the content directly in existing model's field. This is a much simpler approach if you want to keep your data in the same table and don't need additional logic that comes from storing rich texts in dedicated model `ActionText::RichText` (like attachments or handling N+1 loading).

Example:
```ruby
class Post < ApplicationRecord
  # Serialize the field to store ActionText::Content instead of calling `has_rich_text :content`
  serialize :content, coder: ActionText::Content # Table `posts` contains `content` field

  ...
end
```

I haven't added any tests since this comment indicates that this part of code + related parts in `ActionText` will have to be refactored. But suggestions are welcomed how and where to test this.

### DIFF
--- a/lib/lexxy/rich_text_area_tag.rb
+++ b/lib/lexxy/rich_text_area_tag.rb
@@ -31,7 +31,7 @@ module Lexxy
         else
             value.to_s
         end
-  
+
         if html.presence
           Nokogiri::HTML.fragment(html).tap do |fragment|
             fragment.css("action-text-attachment").each do |node|


lib/lexxy/rich_text_area_tag.rb
```diff
@@ -21,9 +21,16 @@ def lexxy_rich_textarea_tag(name, value = nil, options = {}, &block)
...
     private
-      # Tempoary: we need to *adaptarize* action text
+      # Temporary: we need to *adaptarize* action text
       def render_custom_attachments_in(value)
-        if html = value.try(:body_before_type_cast).presence
+        html = case value
+        when ActionText::Content
+            value.to_html
+        when ActionText::RichText
+            value.body_before_type_cast
```
romansklenar, line 30: Not sure if we want to have a generic fallback but added in [6858783](https://github.com/basecamp/lexxy/pull/293/commits/6858783aabd2ebede9495ff8821d6004a9bc6f03)

romansklenar:
RDY for review @jorgemanrubia (I can select only Copilot as reviewer).

jorgemanrubia:
I think this change does not belong to Lexxy, since it is specific to Action Text. It is a nice trick I had never seen before! It is also a departure of how Action Text is designed to store these rich text records, so I would keep it as something app-specific. If it serves, I have never seen problems with using action text polymorphic table in practice. Why would you want to do this?

That Copilot reviewer is annoying! TIL it exists 😅

PR #298, opened 2025-10-02
Adds support for `required` attribute

jorgemanrubia:


src/elements/editor.js
```diff
@@ -358,6 +359,14 @@ export default class LexicalEditorElement extends HTMLElement {
     return !this.editorContentElement.textContent.trim() && !containsVisuallyRelevantChildren(this.editorContentElement)
   }
 
+  #validateRequired() {
+    if (this.hasAttribute("required") && this.#isEmpty) {
+      this.internals.setValidity({ valueMissing: true }, "Please fill out this field.", this.editorContentElement)
```
brunoprietog, line 364: Hey @jorgemanrubia, this message is not localized. We can use this trick to apply the generic browser message instead:

```suggestion
      input = document.createElement("input")
      input.required = true
      this.internals.setValidity({ valueMissing: true }, input.validationMessage, this.editorContentElement)
```

brunoprietog: #299

PR #299, opened 2025-10-02
Use browser validation message

klenis:
I want to share a small hack that displays the same validation message for Lexxy as native input elements. These messages vary between browsers and depend on the user's language. As far as I know, there's no better way to retrieve this value from the browser.


src/elements/editor.js
```diff
@@ -31,6 +31,7 @@ export default class LexicalEditorElement extends HTMLElement {
     super()
     this.internals = this.attachInternals()
     this.internals.role = "presentation"
+    this.validationMessage = Object.assign(document.createElement("textarea"), { required: true }).validationMessage
```
brunoprietog: I think this should be a definition of a getter, with the required based on the presence of the attribute or not using `this.required`. We could have a textarea as a private property, in which messages change depending on the presence of attributes, so as not to recreate the element multiple times.

jorgemanrubia: Yes, good suggestion 👍

There is also a `createElement` in `html_helpers` that you can use for this in a slightly more concise way.

klenis: Good points! I haven't tried this before but here is an attempt to delegate the basic validity to the textarea in order to get native validation messages and also support for other attributes. I've tested it with text and images and adding/removing the required attribute.

There is still room for adding validations specific to the Lexxy (rich content) layer in `#setValidity`.

PR #308, opened 2025-10-05
Fix documentation for syntax highlighting Stimulus controller

davidalejandroaguilar:
Trying Trixx, I'm using Vite and this is the correct import.

So far so good, great job!

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ To apply syntax highlighting to rendered Action Text content, you need to call t
 
 ```javascript
 import { Controller } from "@hotwired/stimulus"
-import { highlightAll } from "lexxy"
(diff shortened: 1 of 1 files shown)


lylo:
I believe the documentation is currently correct with `import { highlightAll } from "lexxy"`

jorgemanrubia:
@davidalejandroaguilar thanks. The current documentation is correct if you are using importmaps. Would you mind modifying your PR to add a note to the README about how to import if using JS bundlers?

davidalejandroaguilar:
Thanks guys, you're right! I just amended my commit.

Think this small change will help prevent confusion. Again, great job @jorgemanrubia!

jorgemanrubia (approved):
Thanks @davidalejandroaguilar

PR #311, opened 2025-10-07
Render attachments with Action Text utilities

seanpdoyle:
This commit includes changes that implement the private `#render_custom_attachments_in` method in terms of existing `ActionText`-provided classes like `Fragment` and `Attachment`.

Namely, this commit replaces `Nokogiri`-scope classes and methods with `ActionText`-scoped alternatives. The content-encoding approach that Lexxy takes shares similarities with the
`.fragment_by_converting_trix_attachments` class method defined by the [ActionText::Attachments::TrixConversion][] concern:

```ruby
def fragment_by_converting_trix_attachments(content)
  Fragment.wrap(content).replace(TrixAttachment::SELECTOR) do |node|
    from_trix_attachment(TrixAttachment.new(node))
  end
end
```

It constructs `<action-text-attachment>` nodes from `<figure data-trix-attachment="…">` nodes. Lexxy is implemented to circumvent that transformation. Given the similarity in approach (find elements, replace or modify their content), I wonder what will constitute the divergence in behavior between the Trix Action Text adapter and the Lexxy Action Text adapter.

If this were implemented in an object (rather than a private view helper method), I wonder what the interface might entail?

```ruby
class TrixEditor
  def replace_attachment(content)
    Fragment.wrap(content).replace("[data-trix-attachment]") do |node|
      trix_attachment = TrixAttachment.new(node)

      Attachment.from_attributes(trix_attachment.attributes)
    end
  end
end

class LexxyEditor
  def replace_attachment(content)
    Fragment.wrap(html).replace(Attachment.tag_name) do |node|
      if node["url"].blank?
        attachment = Attachment.from_node(node)
        node["content"] = render_action_text_attachment(attachment).to_json
      end

      node
    end
  end
end
```

[ActionText::Fragment#replace]: https://github.com/rails/rails/blob/v8.0.3/actiontext/lib/action_text/fragment.rb#L41-L48
[ActionText::Attachments::TrixConversion]: https://github.com/rails/rails/blob/v8.0.3/actiontext/lib/action_text/attachments/trix_conversion.rb#L13-L17

### DIFF
--- a/lib/lexxy/rich_text_area_tag.rb
+++ b/lib/lexxy/rich_text_area_tag.rb
@@ -24,14 +24,13 @@ module Lexxy
       # Tempoary: we need to *adaptarize* action text
       def render_custom_attachments_in(value)
         if html = value.try(:body_before_type_cast).presence
-          Nokogiri::HTML.fragment(html).tap do |fragment|
-            fragment.css("action-text-attachment").each do |node|
-              if node["url"].blank?
-                attachment = ActionText::Attachment.from_node(node)
-                node["content"] = render_action_text_attachment(attachment).to_json
-              end
+          ActionText::Fragment.wrap(html).replace(ActionText::Attachment.tag_name) do |node|
+            if node["url"].blank?
+              attachment = ActionText::Attachment.from_node(node)
+              node["content"] = render_action_text_attachment(attachment).to_json
             end
-          end.to_html
+            node
+          end
         end
       end
   end
--- a/test/fixtures/action_text/rich_texts.yml
+++ b/test/fixtures/action_text/rich_texts.yml
@@ -3,3 +3,9 @@ hello_world:
   name: body
   body:
     <p>Hello everyone</p>
+
+hello_james:
+  record: hello_james (Post)
+  name: body
+  body:
+    <p>Hello <%= ActionText::FixtureSet.attachment("people", :james) %></p>
--- a/test/fixtures/posts.yml
+++ b/test/fixtures/posts.yml
@@ -3,3 +3,6 @@ empty:
 
 hello_world:
   title: "Hello World"
+
+hello_james:
+  title: "Hello James"
--- a/test/helpers/lexxy/tag_helper_test.rb
+++ b/test/helpers/lexxy/tag_helper_test.rb
@@ -1,0 +1,20 @@
+require "test_helper"
+
+class Lexxy::TagHelperTest < ActionView::TestCase
+  helper ActionText::ContentHelper
+
+  test "#lexxy_rich_textarea_tag renders <action-text-attachment> elements" do
+    render inline: <<~ERB, locals: { post: posts(:hello_james) }
+      <%= lexxy_rich_textarea_tag :body, post.body %>
+    ERB
+
+    assert_dom "lexxy-editor", count: 1 do |lexxy_editor, *|
+      assert_dom fragment(lexxy_editor["value"]), "div" do |div|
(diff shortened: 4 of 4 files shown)


seanpdoyle:
I'm not satisfied with the `#replace_attachment` method name, but I think the concept makes sense as one of the few Action Text "Editor" adapter methods.

The purpose of `#replace_attachment` (and the code that it aims to extract) is for transforming an `<action-text-attachment>` into HTML that's ready to be edited by the WYSIWYG editor.

What would be a suitable name for the *other* direction: a hook to transform WYSIWYG-provided content into `<action-text-attachment>` elements? Right now, Lexxy appears to speak `<action-text-attachment>` on both ends. Trix does not yet do that, but instead submits `<figure data-trix-attachment="…">` elements.

There is an opportunity to *make* Trix speak `<action-text-attachment>`. Would it be more appropriate for Trix to:

* transform `<figure data-trix-attachment="…">` elements to `<action-text-attachment>` elements (by configuring trix.js to do it through something like a `trix/actiontext` module?), so that they're already `<action-text-attachments>` by the time that Action Text receives them
* transform `<figure data-trix-attachment="…">` elements to `<action-text-attachments>` elements on the server in a `TrixEditor` lifecycle hook-like method?

jorgemanrubia:
@seanpdoyle I think that, at a high level, Action Text should let you configure:

1. How to transform editor-specific HTML into Action Text canonical HTML, when saving Action Text content in the server.
2. How to transform Action Text canonical HTML into editor-specific HTML, at editor rendering time.

I think this should essentially be the editor adapter interface. Maybe something like `to_action_text_html(editor_html)` and `to_editor_html(action_text_html)`.

Let's see how this could look like with Trix and Lexxy:

Trix, currently:

1. The server transforms Trix HTML to the canonical form when bootstrapping content objects (`fragment_by_converting_trix_attachments`...).
2. Exposes a Trix-specific method `to_trix_html` that is used when rendering the editor field. 

Lexxy:

1. Does not need any transformation because it already sends canonical HTML to the server.
2. It supports canonical Action Text HTML directly, but it needs the the server-side rendering of those attachments, to represent those in the editor. That is why I injected the `content` attribute in the JSON.

So, with the proposed interface"

- Trix adapter would move the `fragment_by_converting_trix_attachments` stuff to the `to_action_text_html` method, and the `to_trix_html` stuff to `to_editor_html`.
- Lexxy would only need to implement `to_editor_html` to inject the content. But I would argue that needing the HTML representation of attachments at editing time is a common need, so I would move this injection to Action Text itself. That would mean that the Lexxy adapter would not really have to perform any transformation at the adapter level.

I know that your WIP PR for Action Text, from pre-Lexxy times, is already totally aligned with this direction btw.

I would leave the interface working at the HTML/string level for maximum flexibility. You are free to parse the HTML and manipulate it as you want to. I wouldn't add the attachment-replacements method to the interface for now. Your example of how to replace attachments here using the existing APIs is great to include in the adapters or docs, but I would like to keep the interfaces as generic and broad as possible for now.

> There is an opportunity to make Trix speak <action-text-attachment>.

Indeed, but this change looks secondary right now. I mean, we need to support the current Trix version, so I would start with building an adapter for that. But I agree, that would be a nice evolution for Trix.

Thanks for your work here 🙏

jorgemanrubia:
To clarify, I think it is better to keep the interface of the adapter working with strings so that different adapters can do whatever transformation they need, without being limited by the APIs for handling nodes exposed by Action Text

PR #315, opened 2025-10-09
Hide upload button from default toolbar when attachments is set to false

### LINKED ISSUE #307: attachments="false" do not hide the image button in toolbar
Hi,

When `attachments: false` on a `rich_textarea`: `<%= form.rich_textarea :content, attachments: false %>` we get these behaviours:

* No images appear in the editor on upload
* Drag and drop is disabled
* Pasting images is disabled
* Dev console to warn: `This editor does not support attachments (it's configured with [attachments=false])`

From that I would also assume the upload icon should not show up as well

**Actual behaviour:**
The image button on the toolbar is displayed. 

**Expected behaviour:**  
The image button on the toolbar is removed. 

Happy to provide a PR if this is considered a bug, and contributions are welcome.

## Screenshots

Image

AlexB52:
Fixes #307 

Based on the discussion on issue #307, this also works and could be enough? 
Happy to provide a change for either approach

```css
:where(lexxy-editor) {
  &[attachments="false"] button[name="upload"] {
    display: none;
  }
}
```


test/system/toolbar_test.rb
```diff
@@ -105,4 +105,10 @@ class ToolbarTest < ApplicationSystemTestCase
 
     assert_no_selector "lexxy-toolbar"
   end
+
+  test "disable attachments" do
+    visit edit_post_path(posts(:empty), attachments_disabled: true)
+
+    assert_no_selector "lexxy-toolbar button[name=upload]"
+  end
```
kevcamel: Could it be worth testing that the button is still present when false or blank?

```ruby
test "enable attachments" do
  visit edit_post_path(posts(:empty), attachments_disabled: [false, nil].sample)

  assert_no_selector "lexxy-toolbar button[name=upload]"
end
```

AlexB52: Sounds good. I've updated the tests.

Note: The form on the page `edit_post_path(posts(:empty), attachments_disabled: true)` also needed a change for the test to work. To some extent, we're more testing the query parameter than the actual `attachments` attribute with these new tests.

AlexB52:
I've rebased onto `basecamp/lexxy/main` to update the js builds to the latest version.

test/dummy/app/views/posts/_form.html.erb
```diff
@@ -11,7 +11,7 @@
   <div>
     <%= form.label :body, style: "display: block" %>
     <%= form.rich_text_area :body, placeholder: "Write something...",
-          attachments: params[:attachments_disabled] ? "false" : nil,
+          attachments: (params[:attachments_disabled] != "true"),
```
AlexB52, line 14: This is a bit annoying as this change now diverges from the other query parameter disabling option logic on L15: `params[:toolbar_disabled] ? "false" : nil`

Issue #51, opened 2025-09-04
Add dark-mode colors


mikker:
I propose just overriding the necessary variables in a `@media` enclosure.

```css
@media (prefers-color-scheme: dark) {
  --lexxy-color-ink: white; /* ish */
}
```

Would be happy to contribute unless you have something in the works or would like to be more settled before adding more styles?

Merovex:
Well, those are variables, so it's very amenable to different methods. Here's what works for my apps, so YMMV.

```css
  :root[data-theme="dark"] {
    /* Colors - Inverted lightness values */
    --lexxy-color-ink: oklch(90% 0 0);
    --lexxy-color-ink-medium: oklch(75% 0 0);
    --lexxy-color-ink-light: oklch(65% 0 0);
    --lexxy-color-ink-lighter: oklch(35% 0 0);
    --lexxy-color-ink-lightest: oklch(15% 0 0);
    --lexxy-color-ink-inverted: oklch(10% 0 0);

    --lexxy-color-accent-dark: oklch(70% 0.19 260);
    --lexxy-color-accent-medium: oklch(75% 0.196 258);
    --lexxy-color-accent-lightest: oklch(25% 0.026 254);

    --lexxy-color-red: oklch(70% 0.15 27);
    --lexxy-color-green: oklch(70% 0.15 145);
    --lexxy-color-blue: oklch(75% 0.196 258);
    --lexxy-color-purple: oklch(70% 0.15 305);

    /* Code syntax colors - adjusted for dark background */
    --lexxy-color-code-token-att: #f97583;
    --lexxy-color-code-token-comment: #8b949e;
    --lexxy-color-code-token-function: #d2a8ff;
    --lexxy-color-code-token-operator: #ff7b72;
    --lexxy-color-code-token-property: #79c0ff;
    --lexxy-color-code-token-punctuation: #c9d1d9;
    --lexxy-color-code-token-selector: #7ee83f;
    --lexxy-color-code-token-variable: #ffa657;

    --lexxy-color-canvas: var(--lexxy-color-ink-inverted);
    --lexxy-color-text: var(--lexxy-color-ink);
    --lexxy-color-text-subtle: var(--lexxy-color-ink-medium);
    --lexxy-color-link: var(--lexxy-color-accent-dark);
    --lexxy-color-selected: var(--lexxy-color-accent-lightest);
    --lexxy-color-selected-dark: var(--lexxy-color-blue-medium);
[...]

PR #377, opened 2025-10-13
Dark mode for Sandbox

zoltanhosszu:
- Support for dark mode
- Added default content for Sandbox editor


jorgemanrubia:
Nice 🙏👏👏!! Cc @adjogima who has authored most of Lexxy CSS here so far.

jorgemanrubia:
Sorry I meant to mention @andyra here!

andyra:
The CSS looks great, @zoltanhosszu, thanks for taking the initiative. 👏

Adding dark mode styles does, indeed, make Lexxy more capable out of the box, but requires more work when implementing  into an app.

1. If your app isn't dark mode capable, Lexxy will be the only part that adapts to dark mode and will look out of place.
2. If your app does adapt to dark mode, you likely already have a set of CSS color tokens. By adding dark mode styles to Lexxy, developers will need to override both Lexxy's light and dark mode color tokens.

**Instead of this…**
```css
--my-text-color: black;
--lexxy-color-text: var(--my-text-color);

@media (prefers-color-scheme: dark) {
  --my-text-color: white;
  --lexxy-color-text: var(--my-text-color);
  /* ☝️ Dark mode override required */
}
```

**Developers can do this…**
```css
--my-text-color: black;
--lexxy-color-text: var(--my-text-color);

@media (prefers-color-scheme: dark) {
  --my-text-color: white;
}
```

Here's a related issue: https://github.com/basecamp/lexxy/issues/51#issuecomment-3259355723

zoltanhosszu:
Thank you @andyra for the feedback, after some thought I tend to agree! What I would do in this case then, is move these styles to the sandbox. That would give an example of how Lexxy could support your styles in implementation overrides. I'll also update the readme to include some examples for this.

andyra:
Putting the dark mode styles in the sandbox is a great idea, thank you!

PR #378, opened 2025-10-14
Horizontal divider

zoltanhosszu:
This PR adds support for `<hr>` tag
- The tag gets inserted inside a figure for the editor, but is handled as a pure `<hr>` for exporting
- 3 tests were added for adding and deleting


zoltanhosszu:
@jorgemanrubia would you please take a look? This is a bit more complex than strikethrough, not sure I did everything correctly. Thanks!

app/assets/javascript/lexxy.js
```diff
@@ -5911,6 +5915,74 @@ class ActionTextAttachmentUploadNode extends ActionTextAttachmentNode {
...
+      }
+    }
+  }
+
+  constructor(key) {
+    super(key);
+  }
+
+  createDOM() {
+    const figure = createElement("figure", { className: "horizontal-divider" });
```
jorgemanrubia, line 5958: Why did you go with using a figure for representing these in the editor instead of using an `hr` too?

zoltanhosszu: Good question! I looked at how Trix did it and wanted to follow the same pattern. But maybe it's unnecessary. Let me reconsider...

jorgemanrubia: I see. Ideally we should use the same markup that action text saves in the database from trix. I can help checking that later

samuelpecher: We're using the entire attachment flow to render an `<hr>`. It looks like it was made to be extensible, but 8 years later we might be able to simplify if it's backwards compatible.

_edit:_ I had to quote the `<hr>` as github allows the html directly in markdown.

We're saving `<bc-attachment content-type="application/vnd.basecamp.horizontal-rule.html"></bc-attachment>` into the db. Trix creates a figure with a `data-trix-attachment` attribute containing the details
```html
<figure contenteditable="false" data-trix-attachment="{&quot;content&quot;:&quot;&lt;hr&gt;&quot;,&quot;contentType&quot;:&quot;application/vnd.basecamp.horizontal-rule.html&quot;}" data-trix-content-type="application/vnd.basecamp.horizontal-rule.html" data-trix-id="129172" class="attachment attachment--content"><hr><figcaption class="attachment__caption"></figcaption></figure>
```

We then use [an ActiveModel instance](https://github.com/basecamp/bc3/blob/master/app/models/rich_text/attachables/content_attachment.rb) to render it from a partial with the content:
```
<%= tag.hr -%>
```

This renders in a view as:
```html
<figure class="attachment attachment--content attachment--horizontal-rule">
  <hr>
</figure>
```

lylo: Philosophically, wouldn't it make sense to save a raw `<hr>`? Lexxy is already breaking Trix conventions by using `<p>` tags so there doesn't seem to be a need to port a Trix approach for this. 

Backwards compatibility can be accomplished by preprocessing Trix content before loading into Lexxy if required.

jorgemanrubia: We are interested in keeping out-the-box compatibility with Trix content. So that is a good reason to try to respect the markup that Trix generates, as long as it makes sense. The paragraph one, as you mention, is a good departure. Better semantics for a core text component (and also handled by Lexical automatically 😅).

In this case, the horizontal rule is not a Trix component, but an extension we add in Basecamp. So there is no such a strong motivation to reuse the same markup. In any case, after playing with it, I like using the `<figure>` wrapper here for the editing representation. It adds some consistency to how we render selectable special elements.

I pushed some tweaks and will merge next. Great one @zoltanhosszu!

jorgemanrubia:
Before merging I need to tweak a bit the ENTER behavior, I will do today.

UPDATE: I will tackle that separately, since it is unrelated to this change.

zoltanhosszu:
Thank you @jorgemanrubia for the fixes and the merge! 🥳🙌

Issue #375, opened 2025-10-12
Toolbar responsiveness breaks when in a dialog


Jeremy-Walton:
I noticed this while experimenting with https://webawesome.com/docs/components/drawer but confirmed it happens in a vanilla `<dialog>` element as well. The toolbar seems to not respect the size of the window. The `connectedCallback()` is getting called because resizing does fix it, but something about the initial load is not working. Not sure if is specific to `<dialog>`  or slots or the #top-layer.

My attempted workaround was the wrap it in a stimulus controller with a slight delay, then find all toolbars and call the resize function, but since `#refreshToolbarOverflow()` is private , that won't work.

<img width="389" height="847" alt="Image" src="https://github.com/user-attachments/assets/305617e1-f251-4549-aaad-adac012996ce" />
<img width="462" height="400" alt="Image" src="https://github.com/user-attachments/assets/783eee08-839a-427d-a8c5-85b321ec80d6" />

zoltanhosszu:
Hi @Jeremy-Walton I believe this will be fixed in #379 here: https://github.com/basecamp/lexxy/pull/379/files#diff-6fa89d2c0ed6d78ab9585e522cbee4494670ce43e86a12b5b074e1b9f39204c3R40

You can try adding that one line in your instance to fix it before #379 gets merged. Hope it helps!

zoltanhosszu:
#379 was merged, this should be fixed. Can you confirm @Jeremy-Walton ?

Jeremy-Walton:
@zoltanhosszu This doesn't seem to be fixed. The toolbar is still breaking out of its container. Seems to happen after navigating away then then back. Disabling turbo completely does seem to resolve the issue so it must be a strange interaction between `<dialog>`, lexxy, and turbo.

## Initial page load
<img width="377" height="247" alt="Image" src="https://github.com/user-attachments/assets/3c65f2b8-5153-44f2-b8d0-ec2104bb219a" />

## After navigating away and then back
<img width="399" height="255" alt="Image" src="https://github.com/user-attachments/assets/6ae8e853-1d8e-4442-8085-00071fd17fe7" />

zoltanhosszu:
@Jeremy-Walton can you check now? :) Thanks!

Jeremy-Walton:
@zoltanhosszu This is working now! Thanks a bunch for seeing that through!!

PR #379, opened 2025-10-14
Undo/redo

zoltanhosszu:
This PR adds undo+redo buttons to the toolbar, while fixing a small bug too.

**Undo/redo**
- The actions are displayed according to the available history actions (disabled state for when you can't undo/redo)
- A spacer element was added so these buttons are placed towards the right side of the toolbar

**Bug fix**
Also there was a bug with the overflow menu, it rearranged the toolbar items on resizing. Here is the current behavior on main. Notice how the buttons are appearing in a random order:

https://github.com/user-attachments/assets/b7ed4964-2af0-40b9-8815-a3ed5ec1a197

This PR fixes this bug too.


src/elements/toolbar.js
```diff
@@ -109,6 +112,34 @@ export default class LexicalToolbarElement extends HTMLElement {
...
+  #monitorHistoryChanges() {
+    this.editor.registerUpdateListener(() => {
+      this.editor.getEditorState().read(() => {
+        const historyState = this.editorElement.historyState
+        if (historyState) {
+          const canUndo = historyState.undoStack.length > 0
+          const canRedo = historyState.redoStack.length > 0
+          
+          this.#setButtonDisabled("undo", !canUndo)
+          this.#setButtonDisabled("redo", !canRedo)
```
jorgemanrubia: These 2 lines could be replaced with `#updateUndoRedoButtonStates()`

src/elements/toolbar.js
```diff
@@ -199,7 +236,15 @@ export default class LexicalToolbarElement extends HTMLElement {
 
   #resetToolbar() {
     while (this.#overflowMenu.children.length > 0) {
-      this.insertBefore(this.#overflowMenu.children[0], this.#overflow);
+      var child = this.#overflowMenu.children[0]
+
+      // Move undo and redo after spacer, and the rest before spacer
+      var to = this.#spacer
```
jorgemanrubia: Could you move the spacer to the menu to represent it there too? I would love if we could remove the spacer-specific rearrangement logic, and we just moved elements from one place to another, in a generic way. We should let you use multiple spacers in the menu and fully configure the Lexxy menu at you want, so I would rather avoid expecting a special anchoring spacer element if possible. Also, I think these spacers can make sense in the collapsed menu too, right?

src/elements/toolbar.js
```diff
@@ -278,6 +322,16 @@ export default class LexicalToolbarElement extends HTMLElement {
         <svg viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><path d="M16 8a2 2 0 110 4 2 2 0 010-4z""/><path d="M22 2a1 1 0 011 1v18a1 1 0 01-1 1H2a1 1 0 01-1-1V3a1 1 0 011-1h20zM3 18.714L9 11l5.25 6.75L17 15l4 4V4H3v14.714z"/></svg>
       </button>
 
+      <div class="lexxy-editor__toolbar-spacer"></div>
```
jorgemanrubia: We should add  a [`role="separator"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/separator_role)

jorgemanrubia (commented):
Looking great!!

src/elements/toolbar.js
zoltanhosszu: This is a great idea! I reworked the code a bit to allow the spacer to move into the overflow as well.

zoltanhosszu:
@jorgemanrubia thanks for the valid feedback! I reworked some files (+ rebased), hope they'll be okay now :)

zoltanhosszu:
This PR also fixes https://github.com/basecamp/lexxy/issues/375

PR #410, opened 2025-10-30
Toolbar resizing with ResizeObserver()

zoltanhosszu:
The aim is to make sure the toolbar gets overflow calculated correctly when displayed in a dynamic environment (e.g. a Turbo frame). The PR will hopefully also fix https://github.com/basecamp/lexxy/issues/375


src/elements/toolbar.js
```diff
@@ -19,12 +19,17 @@ export default class LexicalToolbarElement extends HTMLElement {
...
+    requestAnimationFrame(() => this.#refreshToolbarOverflow());
+
+    this._resizeObserver = new ResizeObserver(() => this.#refreshToolbarOverflow());
+    this._resizeObserver.observe(this);
   }
 
   disconnectedCallback() {
-    window.removeEventListener("resize", this.#refreshToolbarOverflow)
+    if (this._resizeObserver) {
+      try { this._resizeObserver.disconnect(); } catch (_) {}
```
jorgemanrubia: could we remove the empty try/catch? Why is that in place?

zoltanhosszu: well, claude wrote that lol :D

jorgemanrubia: A global catch is an anti-pattern in my book, and needs to be REALLY justified. I would nix until we find the justification.

Also, we need to get rid of the semicolons. We should totally add eslint here to grab thore. But we are not using ; for JS code in general.

PS: Claude is full of anti-patterns, as it can't be otherwise since most of the code out there is not great.

zoltanhosszu: Thanks @jorgemanrubia for the explanation! I removed the try-catch block around the disconnect call

PR #414, opened 2025-10-31
Change default heading to level 1

LukBrezina:
Hello, currently the headings cycle between `h2`-`h4`. You can still create `h1`, but only when entering through markdown with `#` . Seems bit inconsistent to me, I'd expect the editor to output `h1` as default heading. Original trix seems to be toggling `h1` as well.

Sorry for spam if the current behaviour is intentional!


jorgemanrubia:
Hey @LukBrezina this is a reasonable expectation but I would rather keep the current default and make this configurable. We will have a configuration API soon, we are working on it. Thanks for the PR

Issue #428, opened 2025-11-05
Lexxy toolbar overflow menu is cut off on mobile


woochaq:
**Version:** lexxy-0.1.18.beta 
**Context:** Mobile viewport.

When I click the 3-dot (ellipsis) menu, the toolbar opens with position: absolute and is not fully visible on small screens; it gets cut off.

<img width="569" height="380" alt="Image" src="https://github.com/user-attachments/assets/974df8ef-b54c-4604-ba12-d0a6d220729e" />

It seems this could be solved by displaying the entire toolbar (not just the dropdown) in a container with overflow-x-auto enabled. This is the same approach used in the official Lexical demo on https://playground.lexical.dev/.

<img width="634" height="421" alt="Image" src="https://github.com/user-attachments/assets/4d312b2a-e154-4afa-ac57-a39d977b29eb" />

y-dashev:
@woochaq Hey, I created a PR that solves this issue while keeping the overflowing menu where buttons are pushed into the menu. #429

PR #429, opened 2025-11-05
fixing toolbar overflow menu cut

y-dashev:
This is regarding this reported issue: #428 
It looks like the easiest way for fixing this.

### DIFF
--- a/app/assets/stylesheets/lexxy-editor.css
+++ b/app/assets/stylesheets/lexxy-editor.css
@@ -167,6 +167,13 @@
   inset-inline-end: 0;
   padding: var(--lexxy-toolbar-gap);
   position: absolute;
+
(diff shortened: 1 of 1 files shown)


zoltanhosszu:
@y-dashev thank you for looking into it! I would much rather have the overflow menu wrap, instead of horizontal scroll. That behavior sort of makes the overflow menu unnecessary as we could've just made the whole toolbar scroll. Like this: https://github.com/basecamp/lexxy/commit/903d23df42d312fc21aba90cf3ddcccc564fb039

y-dashev:
hey @zoltanhosszu Thanks for your reply! I agree with you about the horizontal scroll. Your solution looks pretty neat as well. I guess we can close this PR.

PR #444, opened 2025-11-12
Use textarea for captions instead of input

andyra:
We currently use an `input` element for attachment captions, which don't support wrapping. By using `textarea` instead, along with the `field-sizing: content;` CSS property, captions can wrap as needed without having an explicit row size set on the element.

@jorgemanrubia I simply changed the element and updated the test to look for the textarea; not sure if anything else needs to be done to seal the deal.


jorgemanrubia (commented):
Nice @andyra. I will see how to stabilize that test and merge. CC @zoltanhosszu in case he has feedback on the approach.

PR #445, opened 2025-11-13
Prompt trigger space

zoltanhosszu:
This PR changes how prompts are triggerred and fixes a small bug:
- Previously prompt was triggerred whenever the character was detected. This can cause issues when a common phrase contains the prompt trigger character (e.g. @ for email addresses, etc). This change limits prompt triggering only if the previous character is a space or a new line.
- Previously if a prompt option was selected with the spacebar, 2 spaces were added after the inserted item. This change fixes that.


jorgemanrubia:
Thanks @zoltanhosszu 👏

PR #447, opened 2025-11-15
Add a rendered output section

5andu:
This PR adds a rendered output display to the sandbox page to make it easier to visualize what it would look like:
rendered output@2x
It also makes the HTML output full width so we can see more code without scrolling.

### DIFF
--- a/test/dummy/app/assets/stylesheets/sandbox.css
+++ b/test/dummy/app/assets/stylesheets/sandbox.css
@@ -8,7 +8,7 @@
     --sandbox-background-color: oklch(15% 0 0);
     --sandbox-text-color: oklch(96% 0 0);
     --sandbox-code-color: #f7fafc;
-  
+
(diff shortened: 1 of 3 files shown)


jorgemanrubia:
Thanks for this one @5andu but I prefer to keep the current views as they are. I think the real value in the sandbox is to see the generated HTLM side by side. Since Lexxy is WYSIWYG, seeing the rendered HTML is not as valuable. Also, you can go to `/posts` and view/edit posts there to check the action text rendering.

5andu:
@jorgemanrubia We can quickly see if there are some visual inconsistencies between the Editor WYSIWYG and the Rendered output section which is also why I thought having "Rendered Output" section might be valuable:
CleanShot 2025-11-20 at 01 23 32@2x
In this example having the Rendered Output section allows us to see that the url color is inconsistent, and that the code snippet is not formatted properly.

PR #451, opened 2025-11-18
Expose name attribute

samuelpecher:
Expose the name attribute to the javascript API

@jorgemanrubia eventually I'd like to expose all the expected attributes to duck-type this as a text area; what do think?


jorgemanrubia (approved):
👍

PR #452, opened 2025-11-18
Color highlighting tool

zoltanhosszu:
**This PR introduces the color highlighting tool**
- This solution relies on Lexical's built in FORMAT_TEXT_COMMAND with "highlight", resulting in a `<mark>` wrap around the highlighted area
- The coloring is done using style attribute and CSS variables (`--highlight-1`, `--highlight-bg-1`, etc)
- Styles are imported by extending `TextNode` conversion logic
- Trix styling is also converted into the appropriate Lexical nodes

These hopefully enable a pretty robust way of highlighting with color 🤞


test/system/toolbar_test.rb
```diff
@@ -165,4 +200,13 @@ class ToolbarTest < ApplicationSystemTestCase
     click_on "Redo"
     assert_equal_html "<p>Hello World</p>", find_editor.value
   end
+
+  private
+    def apply_highlight_option(attribute)
+      click_on "Color highlight"
+
+      within "lexxy-color-dialog dialog[open] [data-button-group='#{attribute}']" do
+        first(".lexxy-color-button").click
```
jorgemanrubia: you could pass a `button_index` as the second param and here do:

```ruby
all(".lexxy-color-button")[button_index - 1].click # So 1 is the first button.
```

So that it is clear which button you are invoking when you invoke the helper

test/system/toolbar_test.rb
```diff
@@ -24,6 +24,41 @@ class ToolbarTest < ApplicationSystemTestCase
...
+  end
+
+  test "background color highlighting" do
+    find_editor.select("everyone")
+    apply_highlight_option("background-color")
+
+    assert_equal_html "<p>Hello <mark style=\"background-color: var(--highlight-bg-1)\">everyone</mark></p>", find_editor.value
+  end
+
+  # TODO: Why is this failing?
```
jorgemanrubia: this is passing for me now? TODO to remove?

samuelpecher:
@jorgemanrubia this is ready for your eyes. 
- `highlighter` object has been extracted from dispatcher and tightened up
- button update for dialog moved into the dialog element
- state uncoupled from dialog (buttons now pass "toggle" message and state is handled by highlighter)
- DOMPurify configuration centralized in `dom_purify.js`

zoltanhosszu:
Awesome work @samuelpecher, thank you for your help!!

z-of-nz:
I follow this topic very closely because I already build something like that with lexicals Mark and storing the marks with a custom node class:

```js

/**
 * ActionTextAttachmentMarkNode
 * - This model references meta elements (like comments) in rich texts. The actual meta content will be saved and updated in RoR backend
 * - MarkNode from lexical is used to spread meta references over multiple html nodes of the rich text
 * - Adds an attribute (data-selection-group) for grouping selection before rails creates instance
 * - Adds an attribute (data-create-meta-content) for triggering the creation and storage of the meta user input as some ActiveStorage instance in backend
 * - Adds an attribute (data-delete-meta-content) for triggering the deletion of the referenced ActiveStorage instance in backend
 * - After creation the sgid is used for identification
 * - Updating of the meta content (UX and storage) should be provided by the parent application
 *
 */

```

Screenshot 2025-11-21 at 09 23 15


I made a PR and a thread in _Discussions_ 
see: https://github.com/z-of-nz/lexxy/tree/comments3

Is this highlighting tool leading into a way for saving the highlighting in some backend? even with a commenting on the highlighting that the user made?

jorgemanrubia (approved):
Super nice work folks 👏💯. Great work tightening things here @samuelpecher, this looks fantastic.

jorgemanrubia:
> Is this highlighting tool leading into a way for saving the highlighting in some backend? even with a commenting on the highlighting that the user made?

We are going with something simpler and based on text nodes for highlighted text. This plays nicely with combining with other styles and also with inline editing of the highighted contents, which happens naturally. The markup looks like 

```html
<mark style="background-color: var(--highlight-bg-2);">hey</mark>
```

That is what we will persist in the server and interpret at editor-bootstrapping time.

PR #460, opened 2025-11-24
Hide "Choose files" input created for attachments

samuelpecher:
* Append the element to lexxy-editor rather than body
* use `display: none;`

cc @jzimdars

### DIFF
--- a/src/editor/command_dispatcher.js
+++ b/src/editor/command_dispatcher.js
@@ -177,6 +177,7 @@ export class CommandDispatcher {
     const input = createElement("input", {
       type: "file",
       multiple: true,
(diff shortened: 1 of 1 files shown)


jzimdars:
Thanks for this, @samuelpecher! Do you think this should have a CSS class so it can be handled differently depending on the situation/app? I can't tell from the code comment if this is actually the element that gets triggered or if it's just an implementation wart. 

I'm also slightly concerned about `display: none`. In the past I've run into trouble when completely hiding inputs that need to be triggered programmatically. Usually we have to do some kind of `position:absolute` shenanigans to keep it iteractive.

Thoughts?

PR #461, opened 2025-11-25
Dialog updates

zoltanhosszu:
This PR introduces a few changes to how the dialogs are handled and structured in code.
- Extracted Dialog class for Color and Link dialog usage
- Dialog is now positioned relative to the trigger button
- Dialog now closes on clicking outside of it
- Better responsive behavior for dialogs
- Color dialog now only shows "Remove all coloring" if selection has highlights
- Added dark mode changes to sandbox for color highlights


src/editor/selection.js
```diff
@@ -145,6 +145,19 @@ export default class Selection {
...
+
+    for (const node of selection.getNodes()) {
+      if ($isTextNode(node) && node.hasFormat("highlight")) {
+        return true
+      }
+    }
+
+    return false
+  }
+
```
samuelpecher: Is this still needed?

src/elements/color_dialog.js
```diff
@@ -67,42 +52,38 @@ export class ColorDialog extends HTMLElement {
     const attribute = button.dataset.style
     const value = button.dataset.value
 
-    this.#editor.dispatchCommand("toggleHighlight", { [attribute]: value })
+    this.editor?.dispatchCommand("toggleHighlight", { [attribute]: value })
```
samuelpecher: Unless I've missed the intent, I think `?` will hide an error state. If there's no editor and the command doesn't fire, that feels like an error state which should log an error.
```suggestion
    this.editor.dispatchCommand("toggleHighlight", { [attribute]: value })
```

src/elements/dialog.js
samuelpecher, line 1: Nice extraction to get the common base between the Lexxy dialogs 👏

src/elements/dialog.js
```diff
@@ -0,0 +1,77 @@
...
+    this.dialog = this.querySelector("dialog")
+    this.addEventListener("keydown", this.#handleKeyDown.bind(this))
+  }
+
+  disconnectedCallback() {
+    this.#removeClickOutsideHandler()
+  }
+
+  show(triggerButton) {
+    if (!this.dialog) return
```
samuelpecher: Same as with `editor?` above, I wouldn't defend against a missing dialog: we'd want to see errors, unless an empty dialog is a valid state?

```suggestion
```

src/elements/dialog.js
```diff
@@ -0,0 +1,77 @@
...
+  get toolbar() {
+    return this.closest("lexxy-toolbar")
+  }
+
+  get editor() {
+    return this.toolbar?.editor
+  }
+
+  #positionDialog() {
+    if (!this.dialog || !this.triggerButton || !this.toolbar) return
```
samuelpecher: same as above

```suggestion

```

src/elements/color_dialog.js
```diff
@@ -1,37 +1,22 @@
...
 
-    this.#toolbar.registerUpdateButtonStatesCallback(this.#updateColorButtonStates.bind(this))
-  }
-
-  #handleKeyDown(event) {
-    if (event.key === "Escape") {
-      event.stopPropagation()
-      this.close()
-    }
+    this.toolbar?.registerUpdateButtonStatesCallback(this.#updateColorButtonStates.bind(this))
```
samuelpecher:
```suggestion
    this.toolbar.registerUpdateButtonStatesCallback(this.#updateColorButtonStates.bind(this))
```

samuelpecher (commented):
This is a nice improvement, especially the common platform in `ToolbarDialog`. I've left a few comments, but I think that being defensive against missing required elements will hide errors which need to be addressed.

src/elements/dialog.js
```diff
@@ -0,0 +1,71 @@
...
+    return this.toolbar.editor
+  }
+
+  #positionDialog() {
+    const left = this.triggerButton.offsetLeft
+    this.dialog.style.insetInlineStart = `${left}px`
+  }
+
+  #setupClickOutsideHandler() {
+    if (this.clickOutsideHandler) return
```
brunoprietog, line 40: The dialog already uses the `closedby` attribute set to `any`, which means that it automatically closes when you click outside or press escape, etc. This is fairly new and is not yet supported in Safari, although closing the dialog with escape is already a default feature of the dialog element when opening it in `modal` mode, which is not the case here.

https://github.com/basecamp/lexxy/blob/d46e6b78695688507eb1b027ca8406d11bd84870/src/elements/toolbar.js#L290

https://github.com/basecamp/lexxy/blob/d46e6b78695688507eb1b027ca8406d11bd84870/src/elements/toolbar.js#L304

https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/closedBy

```suggestion
    if ("closedBy" in HTMLDialogElement.prototype || this.clickOutsideHandler) return
```

It would be better to delegate this to the browser natively when available.

src/elements/dialog.js
```diff
@@ -0,0 +1,71 @@
...
+    if (this.clickOutsideHandler) {
+      document.removeEventListener("click", this.clickOutsideHandler, true)
+      this.clickOutsideHandler = null
+    }
+  }
+
+  #handleKeyDown(event) {
+    if (event.key === "Escape") {
+      event.stopPropagation()
+      this.close()
```
brunoprietog, line 67: Same here:

```suggestion
      if (!("closedBy" in HTMLDialogElement.prototype)) this.close()
```

src/elements/dialog.js
samuelpecher: Good spot! The lack of support in Safari explains a bug not in Safari: the toolbar buttons no longer close the dialog (or rather, they close and immediately reopen it).

We'll need to hook into the [close event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/close_event) to handle clean-up and restoring editor focus.

jorgemanrubia (approved):
Great one @zoltanhosszu thanks!

src/elements/dialog.js
samuelpecher: Addressed in #465

PR #462, opened 2025-11-25
Nested list handling

zoltanhosszu:
This PR changes how nested lists are handled (introduced https://github.com/basecamp/lexxy/commit/6dc1e30d4482ac97f6fa2e84124802c564d0f3d0), in a more conventional Lexical solution: [theming](https://lexical.dev/docs/getting-started/theming).

### Reasoning
Trix's nested items are part of the parent `<li>`:
```html
<li>Parent
  <ul>
    <li>Child</li>
  </ul>
</li>
```

Lexical wraps nested items in their own `<li>`:
```html
<li>Parent</li>
<li class="lexxy-nested-listitem">
  <ul>
    <li>Child</li>
  </ul>
</li>
```
With this change, the extra `lexxy-nested-listitem` class is applied to nested `<li>` tags.

This theme based change allows both of these markups in the content, while fixing the visual issue mentioned in https://github.com/basecamp/lexxy/commit/6dc1e30d4482ac97f6fa2e84124802c564d0f3d0. At the end it's a better solution if someone is replacing Trix for Lexxy, because the extra class allows to style the Lexxy based nested items appropriately.

PS The [Lexical Playground](https://playground.lexical.dev) also uses this approach.


samuelpecher (approved):
@jorgemanrubia I think this is good to go. I worked on this with @zoltanhosszu; it's a nice compact fix to a thorny issue with Lexical's list nodes. It won't impact existing Trix markup (which is more semantic) and requires no markup processing.

jorgemanrubia:
Folks fine to rebase, rebuild assets and merge 👍

PR #465, opened 2025-11-26
Dialog update follow-up

samuelpecher:
This follows-up on #461 by @zoltanhosszu, with @brunoprietog's suggestion.

We're brought a lot of toolbar dialog refinements and created some smooth interactions 

- Uses `<dialog>`'s native `closedBy` where available
- Keep button pressed look on open dialog (thanks @zoltanhosszu for the CSS method)
- Sync link input `required` with open state to prevent form validation issues
- Stop quote toggle changing selection
- Cleaned up dialog state updating callbacks
- Consistent naming of components (e.g. color => highlight)

image

Thanks @zoltanhosszu for the macOS / Safari testing!


src/elements/toolbar.js
samuelpecher, line 1: Moved the dialogs in front to style the buttons while open with the `+` next sibling selector

src/elements/toolbar_dialog.js
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  disconnectedCallback() {
+    this.#removeClickOutsideHandler()
+  }
+
+  updateStateCallback() { }
+
+  show(triggerButton) {
+    if (this.preventImmediateReopen) { return }
```
samuelpecher, line 17: This is required as the event which closes the dialog natively fires _before_ the click event on the button and we don't have a way of cancelling the native `cancel/close`, nor trapping the `click` event from the `close/cancel` event.

src/editor/contents.js
```diff
@@ -386,8 +386,6 @@ export default class Contents {
       elements.forEach((element) => {
         wrappingNode.append(element)
       })
-
-      $setSelection(null)
```
samuelpecher, line 390: Editor was losing selection on quote block transform

src/elements/toolbar_dialog.js
```diff
@@ -0,0 +1,102 @@
...
+    if (!this.#browserHandlesClose) { this.addEventListener("keydown", this.#handleKeyDown.bind(this)) }
+  }
+
+  get #browserHandlesClose() {
+    return this.dialog.closedBy === "any"
+  }
+
+  #handleKeyDown(event) {
+    if (event.key === "Escape") {
+      event.stopPropagation()
```
brunoprietog, line 97: We need to stop the propagation of the event. Unless I'm missing something, this wouldn't run if we have `closedby` available, right?

samuelpecher: This will not fire if closedBy is available, correct; because of the conditional handler on L87.

```javascript
#setupKeydownHandler() {
    if (!this.#browserHandlesClose) { this.addEventListener("keydown", this.#handleKeyDown.bind(this)) }
  }
```

brunoprietog: 👍 so you'd have to tweak this so that the propagation of the event stops in both cases.

src/elements/toolbar_dialog.js
```diff
@@ -0,0 +1,102 @@
+export class ToolbarDialog extends HTMLElement {
+  connectedCallback() {
+    this.dialog = this.querySelector("dialog")
+    if ("closedBy" in this.dialog.constructor.prototype) {
```
brunoprietog, line 4: Any particular reason for doing this in the constructor instead of maintaining the attribute? The getter below that checks this property would still work if you did it that way, and it's simpler IMO.

samuelpecher: I appreciate the question as I considered that!

I decided on enforcing it centrally here as we are _relying_ on the functionality rather than remembering to enable it on each separate child `<dialog>` and having to handle the separate "any" and "closerequest" cases.

jorgemanrubia (approved):
👏

jorgemanrubia:
Noticing that we should also place the focus on the first color button when opening the color dialog, so that you could select color with keys, right?

PR #468, opened 2025-11-27
More code languages

zoltanhosszu:
- Added PHP, Go, Bash, JSON and Diff. These are all languages in Trix, so we should support them in Lexxy, otherwise Trix content loses these values on edit.
- Some tweaking for code highlight color values


samuelpecher:
🥳

PR #469, opened 2025-11-27
Text highlight transform

samuelpecher:
- Uses Lexical's transform to set 'highlight' format, which sets the format on the `TextNode` on edit rather than having to toggle manually
- Fixes the a bug where `TextNode`s created on typing with a styled collapsed selection does not create a `<mark>`
- Corrects highlight button pressed state with improved selection highlight detection as the cursor can be styled with no highlighted nodes yet, `selection.hasFormat` only applied to the anchor


jorgemanrubia (approved):
👏👏

PR #472, opened 2025-11-28
Nested lists

zoltanhosszu:
- Tab to indent, Shift+Tab to outdent list items
- This only works on lists, the current behavior of Tab key is kept for regular text


src/elements/editor.js
```diff
@@ -340,6 +341,25 @@ export default class LexicalEditorElement extends HTMLElement {
     )
   }
 
+  #handleTab() {
```
jorgemanrubia: @zoltanhosszu I think I would move this to the command dispatcher class. We could add a `this.#registerKeyboardCommands` method that registers these. Since this is relative to list formatting, I feel that is a better place, to leave the central editor decoupled for those. I know we are doing the same with ENTER here but ENTER is kind of an special case (not tied to formatting, but we want to intercept and cancel it when pressing CTRL/CMD aloing with it)

src/elements/editor.js
```diff
@@ -340,6 +341,25 @@ export default class LexicalEditorElement extends HTMLElement {
...
+    this.editor.registerCommand(
+      KEY_TAB_COMMAND,
+      (event) => {
+        // Indend or outdent list items
```
packagethief: Typo:

```suggestion
        // Indent or outdent list items
```

src/elements/editor.js
zoltanhosszu: I updated the code @jorgemanrubia, hope this is how you meant it! 😉

src/editor/command_dispatcher.js
```diff
@@ -214,6 +219,25 @@ export class CommandDispatcher {
     this.editor.registerCommand(command, handler, priority)
   }
 
+  #registerKeyboardCommands() {
+    this.editor.registerCommand(
+      KEY_TAB_COMMAND,
+      (event) => {
+        // Indent or outdent list items
```
jorgemanrubia: One last request @zoltanhosszu : you can remove the comment and replace extract the logic to a method with a good name like `#handleListIndentation`. You can see the pattern in `#processSelectionChangeCommands` in the `selection.js` file.

src/editor/command_dispatcher.js
```diff
@@ -214,6 +219,25 @@ export class CommandDispatcher {
...
+            return this.editor.dispatchCommand(OUTDENT_CONTENT_COMMAND, undefined)
+          } else {
+            return this.editor.dispatchCommand(INDENT_CONTENT_COMMAND, undefined)
+          }
+        }
+        return false
+      },
+      COMMAND_PRIORITY_NORMAL
+    )
+  }
```
jorgemanrubia: Also @zoltanhosszu, would be great to have a test for this

zoltanhosszu:
@jorgemanrubia hope it feels right now 🫣

jorgemanrubia (approved):
YAY thanks @zoltanhosszu

PR #473, opened 2025-11-28
Dialog focus improvements

zoltanhosszu:
### Dialog to details
For the two existing toolbar dialogs, this PR changes their structure to `<details>`, similar to the overflow menu. This enables us to remove a lot of extra JS code that came with using `<dialog>`.

### Tab indexes
Changed how tabindexes are assigned. Instead of giving them all a unique value, they all get 0, as per [MDN specs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/tabindex). This simplifies the code but still follows the visual layout of the elements. Plus with the above change, the open dropdowns are now logically placed in the tab order.


src/elements/toolbar.js
```diff
@@ -118,10 +112,18 @@ export default class LexicalToolbarElement extends HTMLElement {
...
+      const baseTabIndex = parseInt(this.editorElement.editorContentElement.getAttribute("tabindex") ?? "0")
+      const buttons = [ ...this.#buttons, ...this.#dialogElements ]
+      buttons.forEach((button, index) => {
+        button.setAttribute("tabindex", `${baseTabIndex + index + 1}`)
+      })
+
+      // Dialogs get focus in Safari, so we need to set them to -1
+      this.#dialogs.forEach((dialog) => {
+        dialog.querySelector("dialog").setAttribute("tabindex", "-1")
+      })
```
samuelpecher: I think this should be handled in ToolbarDialog directly

src/elements/toolbar.js
```diff
@@ -273,6 +275,14 @@ export default class LexicalToolbarElement extends HTMLElement {
...
   }
 
+  get #dialogs() {
+    return Array.from(this.querySelectorAll(":scope > .lexxy-dialog"))
+  }
+
+  get #dialogElements() {
+    return Array.from(this.querySelectorAll(":scope > .lexxy-dialog button, :scope > .lexxy-dialog input"))
+  }
+
```
samuelpecher: I think that we'd be best selecting these in ToolbarDialog, which would let the implementations override it; not sure the toolbar should have so much knowledge about the dialog internals.

src/elements/toolbar.js
```diff
@@ -118,10 +112,18 @@ export default class LexicalToolbarElement extends HTMLElement {
     return [ ...modifiers, pressedKey ].join("+")
   }
 
-  #assignButtonTabindex() {
-    const baseTabIndex = parseInt(this.editorElement.editorContentElement.getAttribute("tabindex") ?? "0")
-    this.#buttons.forEach((button, index) => {
-      button.setAttribute("tabindex", `${baseTabIndex + index + 1}`)
+  #assignTabIndexes() {
+    requestAnimationFrame(() => {
+      const baseTabIndex = parseInt(this.editorElement.editorContentElement.getAttribute("tabindex") ?? "0")
```
samuelpecher: Might be nice to get `Content` to expose it's own tab index and route it through the editor: `this.editorElement.contentTabIndex`

zoltanhosszu:
@samuelpecher would love your input on these changes 🙏

src/elements/toolbar.js
```diff
@@ -280,17 +251,21 @@ export default class LexicalToolbarElement extends HTMLElement {
...
   }
 
   get #buttonsWithSeparator() {
     return Array.from(this.querySelectorAll(":scope > button, :scope > [role=separator]"))
   }
 
-  #reconnect() {
-    this.disconnectedCallback()
-    this.connectedCallback()
+  get #dialogs() {
```
samuelpecher: I'll remove this, I think the callbacks should be handled by the implementations with `beforetoggle`

src/elements/toolbar_dropdown.js
```diff
@@ -0,0 +1,92 @@
...
+      this.close()
+    }
+  }
+
+  #assignTabIndexes() {
+    requestAnimationFrame(() => {
+      this.#interactiveElements.forEach((element) => {
+        element.setAttribute("tabindex", 0)
+      })
+    })
```
samuelpecher: awaiting `nextFrame` would flatten this nicely
```suggestion
      await nextFrame()
      this.#interactiveElements.forEach((element) => {
        element.setAttribute("tabindex", 0)
      })
```

src/elements/toolbar_dropdown.js
```diff
@@ -0,0 +1,92 @@
...
+    this.container.removeAttribute("open")
+  }
+
+  #handleToggle(event) {
+    if (this.container.open) {
+      this.open(event.target)
+    } else {
+      this.close()
+    }
+  }
```
samuelpecher: We need `close`, so `open` makes sense to also have. I think that onopen/onclose handlers should be separate, and could be fired based on `newState`/`oldState` in the `toggle` event, that way it doesn't matter how it's closed.

src/elements/toolbar_dropdown.js
```diff
@@ -0,0 +1,92 @@
...
+  get toolbar() {
+    return this.closest("lexxy-toolbar")
+  }
+
+  get editor() {
+    return this.toolbar.editor
+  }
+
+  open(triggerButton) {
+    this.triggerButton = triggerButton
```
samuelpecher: small nit: it's not a button

samuelpecher (commented):
This is a great improvement to the toolbar DOM, and any interaction we can drop is always a win. 🎉

I think the toggle/open/close situation can have clearer responsibilities and I've got a couple more nits.

I'll drop the state update callback mechanism and ~~roll it up into a `beforetoggle` handler~~ _sigh_, `<details>` doesn't have that event~~

src/elements/toolbar_dropdown.js
zoltanhosszu: I've made some changes here Sam, would appreciate your 2 cents!

src/elements/dropdown/highlight.js
```diff
@@ -51,6 +48,14 @@ export class HighlightDialog extends ToolbarDialog {
     return button
   }
 
+  #handleToggle({ newState }) {
+    if (newState === "open") {
+      this.editor.getEditorState().read(() => {
+        this.#updateColorButtonStates($getSelection())
+      })
+    }
```
samuelpecher, line 56: Just a general thought: I removed the callbacks yet I don't love a toolbar summary reaching into the editor like this. It's the trade-off for not scaffolding a whole series of callbacks.

src/elements/highlight_dropdown.js
```diff
@@ -96,6 +101,4 @@ export class HighlightDialog extends ToolbarDialog {
   }
 }
 
-// We should extend the native dialog and avoid the intermediary <dialog> but not
-// supported by Safari yet: customElements.define("lexxy-hightlight-dialog", HighlightDialog, { extends: "dialog" })
-customElements.define("lexxy-highlight-dialog", HighlightDialog)
+customElements.define("lexxy-dropdown-highlight", HighlightDropdown)
```
samuelpecher: A nit: I'd align the name with the class as it sounds nicest (same for link)

```suggestion
customElements.define("lexxy-highlight-dropdown", HighlightDropdown)
```

The alternative is move them to `./dropdowns/highlight.js` maybe? I'd lean towards the straight forward rename.

src/elements/toolbar_dropdown.js
```diff
@@ -0,0 +1,93 @@
...
+
+  #handleClickOutside({ target }) {
+    if (!this.container.open) return
+
+    const isClickInsideDropdown = this.contains(target)
+    const isClickOnTrigger = this.trigger.contains(target)
+
+    if (!isClickInsideDropdown && !isClickOnTrigger) {
+      this.close()
+    }
```
samuelpecher: Is another advantage of `<details>` that we can use `container` to ignore clicks (and not pass around `trigger`)? 
I'd then consider just inlining everything
```suggestion
    if (this.container.open && !this.container.contains(target)) this.close()
```

src/elements/toolbar_dropdown.js
```diff
@@ -0,0 +1,93 @@
+import { nextFrame } from "../helpers/timing_helpers"
+
+export class ToolbarDropdown extends HTMLElement {
+  connectedCallback() {
+    this.container = this.closest("details")
+
+    this.container.addEventListener("toggle", this.#handleToggle.bind(this))
+    this.container.addEventListener("keydown", this.#handleKeyDown.bind(this))
+
+    this.#assignTabIndexes()
```
samuelpecher: Technically it's "indices"
```ruby
bc3(dev):002> "index".pluralize
=> "indices"
```

src/elements/toolbar.js
```diff
@@ -134,17 +112,15 @@ export default class LexicalToolbarElement extends HTMLElement {
   }
 
   #assignButtonTabindex() {
```
samuelpecher: Can you fix this cap while you're at it please?

src/elements/toolbar_dropdown.js
zoltanhosszu: oh I love this!

src/elements/highlight_dropdown.js
zoltanhosszu: I ended up moving them to the a `dropdown` folder, that seems like a nice way for future extensions.

zoltanhosszu:
@jorgemanrubia would you mind taking a look at these changes? 🙏

jorgemanrubia (approved):
Great work 👏

PR #479, opened 2025-12-03
Editor bugfixes

zoltanhosszu:
### Editor bugfixes

This PR fixes a few small issues:
- a small edge case for a toolbar state inconsistency
- a weird blockquote unwrapping behavior
- a broken editor state
- undo/redo were triggering twice if keyboard shortcuts were used

More details in each commit.

### DIFF
--- a/test/system/editor_to_string_test.rb
+++ b/test/system/editor_to_string_test.rb
@@ -28,13 +28,13 @@ class EditorValueMethodsTest < ApplicationSystemTestCase
     end
     sleep 0.1
 
-    assert_editor_plain_text "\n\n[example.png]\n\n"
+    assert_editor_plain_text "[example.png]\n\n"
 
     find("figcaption textarea").click.send_keys("Example Image")
     find_editor.click
     sleep 0.1
 
-    assert_editor_plain_text "\n\n[Example Image]\n\n"
+    assert_editor_plain_text "[Example Image]\n\n"
   end
 
   test "toString returns content for custom_action_text_attachment (mention)" do


src/elements/editor.js
```diff
@@ -131,7 +131,7 @@ export default class LexicalEditorElement extends HTMLElement {
       const root = $getRoot()
       root.clear()
       if (html !== "") root.append(...this.#parseHtmlIntoLexicalNodes(html))
-      root.select()
+      root.selectEnd()
```
samuelpecher, line 163: Nice

src/elements/editor.js
```diff
@@ -143,8 +143,14 @@ export default class LexicalEditorElement extends HTMLElement {
...
-    if (!html) html = "<p></p>"
-    const nodes = $generateNodesFromDOM(this.editor, parseHtml(`<div>${html}</div>`))
+    const defaultHtml = "<p></p>"
+    if (!html) html = defaultHtml
+    let nodes = $generateNodesFromDOM(this.editor, parseHtml(`<div>${html}</div>`))
+
+    if (nodes.length === 0) {
+      nodes = $generateNodesFromDOM(this.editor, parseHtml(defaultHtml))
+    }
+
```
samuelpecher, line 177: From the caller in `set value=`on L30 and its caller in `#loadInitialValue` on L267, there's varying levels where a default is applied. `#parseHtmlIntoLexicalNodes` isn't even called if `html === ""`.

I'd be up for avoiding a double call to $generateNodesFromDOM if we can.

src/editor/contents.js
```diff
@@ -352,9 +352,19 @@ export default class Contents {
...
+      children.forEach((child) => {
+        if ($isTextNode(child) && child.getTextContent().trim() !== "") {
+          const newParagraph = $createParagraphNode()
+          newParagraph.append(child)
+          node.insertBefore(newParagraph)
+        } else if (!$isLineBreakNode(child)) {
+          node.insertBefore(child)
+        }
+      })
+    }
```
samuelpecher, line 366: Can we get any of this behavior for free with [Lexical's `$unwrapNode`](https://github.com/facebook/lexical/blob/89a093ffd00c0d1a1b2b7bcc66960005c6752b74/packages/lexical-utils/src/index.ts#L841)?

src/elements/toolbar.js
```diff
@@ -348,11 +348,11 @@ export default class LexicalToolbarElement extends HTMLElement {
  
       <div class="lexxy-editor__toolbar-spacer" role="separator"></div>
  
-      <button class="lexxy-editor__toolbar-button" type="button" name="undo" data-command="undo" title="Undo" data-hotkey="cmd+z ctrl+z">
+      <button class="lexxy-editor__toolbar-button" type="button" name="undo" data-command="undo" title="Undo">
```
samuelpecher, line 407: Absolutely fantastic find

src/editor/contents.js
zoltanhosszu: `$unwrapNode` works great if the content of a blockquote is wrapped in a paragraph. But this is totally valid HTML, that we are currently not transforming on import:
```html
<blockquote>Hello World</blockquote>
```

With the above example, `$unwrapNode` fails with [Lexical error #99](https://lexical.dev/docs/error?code=99).

src/elements/editor.js
zoltanhosszu: Thanks @samuelpecher ! Would [this change](https://github.com/basecamp/lexxy/pull/479/changes/7dadbb003c03791a54c9fa5d2e2e3d2aed7fb58f) make more sense?

zoltanhosszu: Scratch that, [I found a better solution](https://github.com/basecamp/lexxy/pull/479/changes/92a06f52fa09201de814d322d3fbbd25cfb0d7b5). We needed to check every toolbar action if it was performed on the root element, and handle it just by inserting the required element. I think this is a much better solution.

zoltanhosszu: @samuelpecher went back to forcing a `<p>` on empty state. How's this? :)

PR #488, opened 2025-12-07
Fix image captions with colons being stripped by DOMPurify

lylo:
DOMPurify validates attribute values against a URI regex unless the attribute is marked as "URI safe". The `caption` attribute was in `ALLOWED_ATTR` but not in `URI_SAFE_ATTRIBUTES`, causing values like "photographer: name" to be stripped because the colon made it look like an unknown protocol scheme.

Adding `caption` and `filename` to `ADD_URI_SAFE_ATTR` prevents this validation, allowing any text content in these attributes.


samuelpecher:
No objections to the principle. We're not using those values except storing them in the rich text.

@packagethief can I get your :eyes: on this?

packagethief:
No objections here 👍

PR #495, opened 2025-12-09
Toolbar accessibility

zoltanhosszu:
### This PR adds a few changes for better accessibility
- If the user is using the keyboard to trigger toolbar items, the selection stays in the toolbar.
- Toolbar is only keyboard accessible, if the editor is focused. Shift+Tab from the editor takes the focus to the toolbar.
- Tabindex handling is now working from code, with [rolling/roving tabindex](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#keyboardnavigationbetweencomponents(thetabsequence)) assignment. Only the currently focused item in the toolbar has `tabindex="0"`, the rest get `-1`. Navigation in the toolbar is now enabled using the arrow keys.
- Better handling of tab key on lists and code blocks

More details in the individual commits


src/editor/selection.js
```diff
@@ -145,6 +147,45 @@ export default class Selection {
...
+
+  get isInsideCodeBlock() {
+    const selection = $getSelection()
+    if (!$isRangeSelection(selection)) return false
+
+    const anchorNode = selection.anchor.getNode()
+    if (!anchorNode.getParent()) return false
+
+    const topLevelElement = anchorNode.getTopLevelElementOrThrow()
+    return $isCodeNode(topLevelElement) || selection.hasFormat("code")
```
samuelpecher: I think `hasFormat('code')` will pick up inline code

```suggestion
    return $isCodeNode(topLevelElement)
```

src/editor/selection.js
```diff
@@ -145,6 +147,45 @@ export default class Selection {
...
+
+    return isIndented
+  }
+
+  get isInsideCodeBlock() {
+    const selection = $getSelection()
+    if (!$isRangeSelection(selection)) return false
+
+    const anchorNode = selection.anchor.getNode()
+    if (!anchorNode.getParent()) return false
```
samuelpecher: I'm not sure I see why a parent is required here? If it's to ensure it's not the root, `$getRoot` seems more direct.

src/elements/toolbar.js
```diff
@@ -63,20 +65,28 @@ export default class LexicalToolbarElement extends HTMLElement {
...
+    this.editor.update(() => {
+      // Keep the focus on the toolbar when using a keyboard to trigger the command
+      const isKeyboard = event.detail === 0 || !event.isTrusted
+      if (isKeyboard) {
+        $addUpdateTag(SKIP_DOM_SELECTION_TAG);
+      }
+
+      const { command, payload } = button.dataset
+      this.editor.dispatchCommand(command, payload)
+    })
```
samuelpecher: I'd hoist the non-Lexical code out of the update callback for clarity, use the tag option, and destructure `dataset`.
Can we not use `pointerId === -1` to check for [a non-pointer pointing device](https://w3c.github.io/pointerevents/#event-attributes:~:text=pointerId%20MUST%20be%20%2D1%20and%20pointerType%20MUST%20be%20an%20empty%20string)?
```javascript
 #dispatchButtonCommand(event, { dataset: { command, payload }) {
    const isKeyboard = event instanceof PointerEvent && event.pointerId === -1

    this.editor.update(() => {
      this.editor.dispatchCommand(command, payload)
    }, { tag: isKeyboard ? SKIP_DOM_SELECTION_TAG : undefined } )
  }
```

src/editor/selection.js
```diff
@@ -145,6 +147,45 @@ export default class Selection {
...
+            if (grandparent && $isListItemNode(grandparent)) {
+              isIndented = true
+              break
+            }
+          }
+        }
+        current = current.getParent()
+      }
+      if (isIndented) break
+    }
```
samuelpecher, line 161: We can actually cheat here with `$getNearestNodeOfType` and if it's a list, let Lexical do the work with `$getListDepth`. Then just break out of the function entirely with `return true` if we have a hit.
 
```javascript
    for (const node of selection.getNodes()) {
      const closestListNode = $getNearestNodeOfType(node, ListNode)
      if (closestListNode && $getListDepth(closestListNode) > 1) {
        return true
      }
    }
```

src/editor/selection.js
```diff
@@ -145,6 +147,45 @@ export default class Selection {
...
+
+    return isIndented
+  }
+
+  get isInsideCodeBlock() {
```
samuelpecher, line 166: I think this could also be compressed with `$getNearestNodeOfType(node, CodeNode)`.

src/editor/command_dispatcher.js
```diff
@@ -220,7 +220,7 @@ export class CommandDispatcher {
   }
 
   #registerKeyboardCommands() {
-    this.editor.registerCommand(KEY_TAB_COMMAND, this.#handleListIndentation.bind(this), COMMAND_PRIORITY_NORMAL)
+    this.editor.registerCommand(KEY_TAB_COMMAND, this.#handleTabKey.bind(this), COMMAND_PRIORITY_NORMAL)
```
samuelpecher, line 280: :heart:

src/editor/command_dispatcher.js
```diff
@@ -270,14 +270,23 @@ export class CommandDispatcher {
...
+        return false
       } else {
+        event.preventDefault()
         return this.editor.dispatchCommand(INDENT_CONTENT_COMMAND, undefined)
       }
+    } else if (this.selection.isInsideCodeBlock) {
+      const selection = $getSelection()
+      if ($isRangeSelection(selection) && selection.isCollapsed()) {
+        return true
+      }
```
samuelpecher: Can I double check this is the expected behavior in the code block?
- collapsed
  - tab: jumps to the language selector
  - shift-tab: jumps to the toolbar
- !collapsed
  - tab: inserts a tab character
  - shift-tab: unindents / nothing

src/editor/command_dispatcher.js
```diff
@@ -270,14 +270,23 @@ export class CommandDispatcher {
...
     }
     return false
   }
```
samuelpecher, line 337: How about extracting the sub-handlers by node type, since they all `return`? This removes indentation and lets me reason a bit more about what's happening...

```javascript
#handleTabKey(event) {
    if (this.selection.isInsideList) {
      return this.#handleTabForList(event)
    }
    if (this.selection.isInsideCodeBlock) {
      return this.#handleTabForCode(event)
    }
    return false
  }
```

For the list handler I'd bail on the false condition early, allowing a simple switch on indent/outdent. While it's not "pure boolean logic", it makes clear the refusal to handle ShiftTab when the list isn't idented.
```javascript
  #handleTabForList(event) {
    if (event.shiftKey && !this.selection.isIndentedList) return false

    event.preventDefault()
    const command = event.shiftKey? OUTDENT_CONTENT_COMMAND : INDENT_CONTENT_COMMAND
    return this.editor.dispatchCommand(command, undefined)
  }

  #handleTabForCode(event) {
    return $isRangeSelection(selection) && selection.isCollapsed()
  }
```

src/elements/toolbar.js
```diff
@@ -121,9 +125,36 @@ export default class LexicalToolbarElement extends HTMLElement {
...
+  #handleFocus() {
+    this.#resetTabIndexValues()
+    this.#focusableItems[0].tabIndex = 0
+  }
+
+  #handleFocusOut() {
+    if (this.contains(document.activeElement)) return
+
+    this.#resetTabIndexValues()
+  }
```
samuelpecher:
```suggestion
    if (!this.contains(document.activeElement)) {
          this.#resetTabIndexValues()
    }
```

src/elements/toolbar_dropdown.js
```diff
@@ -39,6 +37,8 @@ export class ToolbarDropdown extends HTMLElement {
     this.trigger = trigger
```
samuelpecher: I think `this.trigger` isn't used any more right?

samuelpecher:
Nice functionality for keyboard-only toolbar navigation 👏. A few areas can be tightened up by handing-off to Lexical functions.

src/editor/selection.js
zoltanhosszu: oh this is beautiful!

src/editor/command_dispatcher.js
zoltanhosszu: uuuh I love this! 😍

src/editor/command_dispatcher.js
zoltanhosszu: Mostly yes, except for one thing: I would love to have shift+tab also get out of the !collapsed state, but didn't find a nice way from Lexical do to it. Any suggestions on that?

Issue #458, opened 2025-11-22
Cannot use Lexxy with Turbo when using custom toolbar nested in the editor element


daniel-rikowski:
In my app I'm using a custom toolbar which is nested in the editor element, similar to how the default toolbar is created.

```html
<lexxy-editor toolbar="toolbar-custom" value="...">
  <lexxy-toolbar id="toolbar-custom">
    <!-- ... -->
  </lexxy-toolbar>
</lexxy-editor>
```

(various attributes stripped for legibility)

This works almost 100%. (Some problems with lexxy-link-dialog and nested form elements...)

But when I navigate to the page containing the editor with Turbo, Lexxy cannot find the toolbar.  ("this.toolbarElement.setEditor is not a function")

Call stack:

```
#attachToolbar lexxy.esm.js:3452
#initialize lexxy.esm.js:3263
connectedCallback lexxy.esm.js:3154
renderElement turbo.es2017-esm.js:5349
assignNewBody turbo.es2017-esm.js:5524
replaceBody turbo.es2017-esm.js:5430
preservingPermanentElements turbo.es2017-esm.js:1795
preservingPermanentElements turbo.es2017-esm.js:1896
replaceBody turbo.es2017-esm.js:5428
```

Lexxy assumes correctly that there is a toolbar, but then cannot find the element.

This seems to be the relevant code:

```js
  #attachToolbar() {
    if (this.#hasToolbar) {
      this.toolbarElement.setEditor(this);
    }
  }
```
```js
  get #hasToolbar() {
    return this.getAttribute("toolbar") !== "false"
  }
```
```js
 get toolbarElement() {
    if (!this.#hasToolbar) return null

    this.toolbar = this.toolbar || this.#findOrCreateDefaultToolbar();
    return this.toolbar
  }
```
```js
  #findOrCreateDefaultToolbar() {
    const toolbarId = this.getAttribute("toolbar");
    return toolbarId ? document.getElementById(toolbarId) : this.#createDefaultToolbar()
  }
```

When the toolbar is outside the editor it works (which then requires some CSS changes to not look broken...) 

I'm not 100% sure what is happening, but it appears that when `connectedCallback` is called, the toolbar cannot yet be found using `document.getElementById`. It looks like this happens, because it is later in the DOM tree, although that is no problem on the initial page load. Perhaps there is something different going on when Turbo calls `document.body.replaceWith`.

Perhaps a `requestAnimationFrame` is all which is needed to solve this problem.

[...]

PR #496, opened 2025-12-09
Fix attachToolbar race under Turbo navigation

finist:
## Fix race condition when using custom toolbar with Turbo navigation

This PR fixes a race condition that occurs when `<lexxy-editor>` is initialized before the custom `<lexxy-toolbar>` is upgraded as a Web Component during Turbo-driven navigation.

### Problem

When navigating with Turbo, DOM elements are inserted incrementally.  
`<lexxy-editor>` often connects **before** the custom toolbar is upgraded, so:

```js
this.toolbarElement.setEditor(this)
```

throws:

```
TypeError: this.toolbarElement.setEditor is not a function
```

This interrupts editor initialization and prevents content from loading.

Related issue:  
https://github.com/basecamp/lexxy/issues/458

### What this PR changes

- `attachToolbar()` no longer assumes that `toolbarElement` is already a fully upgraded custom element.
- The method now:
  - attempts to attach immediately (for full page loads),
  - but if `setEditor` is not yet available, it waits for:

    ```js
    customElements.whenDefined("lexxy-toolbar")
    ```

    and retries on the next animation frame.
- When a custom toolbar is specified via the `toolbar` attribute, Lexxy no longer creates a default toolbar.  
  This prevents the duplicate-toolbar issue under Turbo navigation.

### Result

- Custom toolbars now work reliably with Turbo navigation.
- No errors are thrown during editor initialization.
- Editor content loads correctly.
- Default behavior (when no custom toolbar is provided) remains unchanged.


zoltanhosszu:
@finist thank you for this! Could you move these changes to `src/elements/toolbar.js` and use the build commands to rebuild the source?

finist:
@zoltanhosszu done

justwiebe:
Is there any workaround until this is merged?

justwiebe:
It's pretty hacky but I was able to do the following:
```js
document.addEventListener("turbo:before-render", (event) => {
  const { newBody, resume } = event.detail
  if (!newBody?.querySelector("lexxy-toolbar")) return

  event.preventDefault()
  try {
    document.adoptNode(newBody)
    customElements.upgrade(newBody)
  } finally {
    resume()
  }
})
```

I also needed to add `data-turbo-permanent` on the toolbar so that morph wouldn't kill it

src/elements/editor.js
```diff
@@ -424,73 +344,53 @@ export class LexicalEditorElement extends HTMLElement {
...
+      const toolbar = this.toolbarElement;
+      if (toolbar && typeof toolbar.setEditor === "function") {
+        toolbar.setEditor(this);
+      }
+    };
+
+    attach();
+
+    if (!this.toolbar || typeof this.toolbar.setEditor !== "function") {
+      const toolbarId = this.getAttribute("toolbar");
```
justwiebe, line 378: I think for the rebase this will need to be updated

```suggestion
      const toolbarId = this.config.get("toolbar");
```

PR #501, opened 2025-12-16
Add isEmpty, isBlank, and toString()

samuelpecher:
### `isEmpty` and `isBlank`

Added these convenience functions to the public API. 
* `isEmpty === true` signifies a completely empty editor
* `isBlank === true` signifies an editor with only whitespace in its plain text representation

### `toString()`

Provide `toString()` in the public API to extract the plain text content without needing to processing the html returned by `value`. The value is cached for speed and cleared on content update.

This leverages's lexical's `getTextContent()` and values added to our custom nodes:
* Attachments: the attachment caption or filename in square brackets `[example.png]`
* Custom Attachments: the attachment innerText or the content-type in square brackets
* Horizontal Rule: `┄`

### Sandbox Plain Text view

A Plain Text view has been added to the sandbox as a details view which can be expanded to replace the HTML view if desired


src/elements/editor.js
```diff
@@ -99,6 +99,14 @@ export default class LexicalEditorElement extends HTMLElement {
     return this.dataset.blobUrlTemplate
   }
 
+  get isEmpty() {
+    return [ "<p><br></p>", "<p></p>", "" ].includes(this.value.trim())
+  }
+
+  get isBlank() {
+    return this.isEmpty || this.toString().match(/^\s*$/g) !== null
+  }
```
jorgemanrubia, line 108: Love these!

src/elements/editor.js
```diff
@@ -273,13 +291,18 @@ export default class LexicalEditorElement extends HTMLElement {
...
+      this.#clearCachedValues()
       this.#internalFormValue = this.value
       this.#toggleEmptyStatus()
       this.#setValidity()
     }))
   }
 
+  #clearCachedValues() {
+    this.cachedValue = null
+    this.cachedStringValue = null
```
jorgemanrubia, line 303: Good call extracting the method

test/test_helpers/editor_helper.rb
```diff
@@ -4,6 +4,22 @@ def find_editor(selector = "lexxy-editor")
     @handlers_by_selector[selector] ||= EditorHandler.new(page, selector)
   end
 
+  def editor_is_empty
```
jorgemanrubia: I would move these as methods of the `EditorHandler` we use in testing using Ruby convention: (`find_editor.empty?`) and would just use that in the tests

samuelpecher: I thought about going down that route, but hesitated as it was a facade for the js API.

You're right, it looks much better in Ruby's conventions.

PR #504, opened 2025-12-18
Configuration API

zachasme:
Based off suggestions from @jorgemanrubia, this PR allows configuring lexxy instances in three layers:

- Default config properties for all Lexxy instances.
- "Preset" configurations so that you can re-use special instances here and there.
- Instance-level configuration via element attributes for one-off overrides.

```js
// In app JS initializer
import { configure } from "lexxy"
configure({
  default: { ... }, // To override Lexxy defaults
  simple: {
    toolbar: false
  }
})

// Reference the preset and override specific values
<lexxy-editor preset="simple" attachments="false" ...>
```

I took into consideration the work on [custom-upload-events](https://github.com/basecamp/lexxy/tree/custom-upload-events), which will require:
- "global" options not attached to any instance, due to the static nature of Lexical's `importDOM` API. 

```js
lexxyConfig.global.get("attachmentTagName")
```

Note that I'm only exposing the `configure` function as public API, but we could include a getter if end-users need to read current config values.

As for next steps:

- We can introduce more fine-grained control over e.g. the toolbar. This will require additional logic around merging things like `{toolbar: true}` and `{toolbar:{bold:false}}`.
- Lexical Extensions! Should we consider refactoring parts of Lexxy into a set of extensions?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,19 @@ jobs:
           ruby-version: ruby-3.3.7
           bundler-cache: true
 
-      - name: Run tests
+      - name: Set up Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: '20'
+          cache: 'yarn'
+
+      - name: Install JavaScript dependencies
+        run: yarn install --frozen-lockfile
+
+      - name: Run JavaScript tests
+        run: yarn test
+
+      - name: Run Rails tests
         env:
           RAILS_ENV: test
           # REDIS_URL: redis://localhost:6379/0
--- a/config/ci.rb
+++ b/config/ci.rb
@@ -4,8 +4,9 @@ CI.run do
   step "Style: Ruby", "bin/rubocop -f github"
   step "Style: JavaScript", "yarn lint"
 
-  step "Tests: prepare", "env RAILS_ENV=test bin/rails db:test:prepare"
-  step "Tests: run", "env RAILS_ENV=test bin/rails test:all -d"
+  step "Tests: JavaScript", "yarn test"
+  step "Tests: Rails (prepare)", "env RAILS_ENV=test bin/rails db:test:prepare"
+  step "Tests: Rails (run)", "env RAILS_ENV=test bin/rails test:all -d"
 
   unless success?
     failure "Signoff: CI failed.", "Fix the issues and try again."
--- a/package.json
+++ b/package.json
@@ -18,15 +18,18 @@
     "@rollup/plugin-node-resolve": "^16.0.1",
     "@rollup/plugin-terser": "^0.4.4",
     "eslint": "^9.15.0",
+    "jsdom": "^27.3.0",
     "rollup": "^4.44.1",
     "rollup-plugin-copy": "^3.5.0",
-    "rollup-plugin-gzip": "^4.1.1"
+    "rollup-plugin-gzip": "^4.1.1",
+    "vitest": "^4.0.16"
   },
   "scripts": {
     "build": "rollup -c",
     "build:npm": "rollup -c rollup.config.npm.mjs",
     "watch": "rollup -wc --watch.onEnd=\"rails restart\"",
     "lint": "eslint",
+    "test": "vitest --environment=jsdom",
     "prerelease": "yarn build:npm",
     "release": "yarn build:npm && yarn publish"
   },
--- a/test/javascript/config/configuration.test.js
+++ b/test/javascript/config/configuration.test.js
@@ -1,0 +1,30 @@
+import { expect, test } from "vitest"
+import Configuration from "../../../src/config/configuration"
+
+const config = new Configuration({
+  default: { richText: true },
+})
+
+test("gets path", () => {
+  expect(config.get("default.richText")).toBe(true)
+})
+
+test("returns undefined for missing option", () => {
+  expect(config.get("default.missing")).toBe(undefined)
+})
+
+test("overwrites bool with hash", () => {
+  const config = new Configuration(
+    { toolbar: false },
+    { toolbar: { bold: true } }
+  )
+  expect(config.get("toolbar.bold")).toBe(true)
+})
+
+test("ovewrites hash with bool", () => {
+  const config = new Configuration(
+    { toolbar: { bold: true } },
+    { toolbar: false }
+  )
+  expect(config.get("toolbar")).toBe(false)
+})
--- a/test/javascript/editor/configuration.test.js
+++ b/test/javascript/editor/configuration.test.js
@@ -1,0 +1,38 @@
+import { expect, test } from "vitest"
+import { createElement } from "../helpers/dom_helper"
+import EditorConfiguration from "../../../src/editor/configuration"
+import { configure } from "../../../src/index"
+
+configure({
+  default: {
+    richText: true,
+    attachments: true
+  },
+  simple: {
+    richText: false
+  }
+})
+
+test("uses defaults", () => {
+  const element = createElement("<lexxy-editor></lexxy-editor>")
+  const config = new EditorConfiguration(element)
+  expect(config.get("attachments")).toBe(true)
+})
+
+test("uses preset", () => {
+  const element = createElement("<lexxy-editor preset='simple'></lexxy-editor>")
+  const config = new EditorConfiguration(element)
+  expect(config.get("richText")).toBe(false)
+})
+
+test("preset fallbacks to default", () => {
+  const element = createElement("<lexxy-editor preset='simple'></lexxy-editor>")
+  const config = new EditorConfiguration(element)
+  expect(config.get("attachments")).toBe(true)
+})
+
+test("overrides defaults", () => {
+  const element = createElement("<lexxy-editor rich-text='false'></lexxy-editor>")
+  const config = new EditorConfiguration(element)
+  expect(config.get("richText")).toBe(false)
+})
--- a/test/javascript/helpers/dom_helper.js
+++ b/test/javascript/helpers/dom_helper.js
@@ -1,0 +1,5 @@
+export function createElement(html) {
+  const element = document.createElement("div")
+  element.innerHTML = html
+  return element.firstChild;
+}
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,6 +2,201 @@
 # yarn lockfile v1
 
 
+"@acemir/cssom@^0.9.28":
+  version "0.9.29"
+  resolved "https://registry.yarnpkg.com/@acemir/cssom/-/cssom-0.9.29.tgz#3ae97903c49b388a5ff5d231a7a6d42a32e5b148"
+  integrity sha512-G90x0VW+9nW4dFajtjCoT+NM0scAfH9Mb08IcjgFHYbfiL/lU04dTF9JuVOi3/OH+DJCQdcIseSXkdCB9Ky6JA==
+
+"@asamuzakjp/css-color@^4.1.1":
+  version "4.1.1"
+  resolved "https://registry.yarnpkg.com/@asamuzakjp/css-color/-/css-color-4.1.1.tgz#e5c662f488367dc74194609179fe6621f2586114"
+  integrity sha512-B0Hv6G3gWGMn0xKJ0txEi/jM5iFpT3MfDxmhZFb4W047GvytCf1DHQ1D69W3zHI4yWe2aTZAA0JnbMZ7Xc8DuQ==
+  dependencies:
+    "@csstools/css-calc" "^2.1.4"
+    "@csstools/css-color-parser" "^3.1.0"
+    "@csstools/css-parser-algorithms" "^3.0.5"
+    "@csstools/css-tokenizer" "^3.0.4"
+    lru-cache "^11.2.4"
+
+"@asamuzakjp/dom-selector@^6.7.6":
+  version "6.7.6"
+  resolved "https://registry.yarnpkg.com/@asamuzakjp/dom-selector/-/dom-selector-6.7.6.tgz#9cd7671a61a9cb490852ed6a441b3b0950aab945"
+  integrity sha512-hBaJER6A9MpdG3WgdlOolHmbOYvSk46y7IQN/1+iqiCuUu6iWdQrs9DGKF8ocqsEqWujWf/V7b7vaDgiUmIvUg==
+  dependencies:
+    "@asamuzakjp/nwsapi" "^2.3.9"
+    bidi-js "^1.0.3"
+    css-tree "^3.1.0"
+    is-potential-custom-element-name "^1.0.1"
+    lru-cache "^11.2.4"
+
+"@asamuzakjp/nwsapi@^2.3.9":
+  version "2.3.9"
+  resolved "https://registry.yarnpkg.com/@asamuzakjp/nwsapi/-/nwsapi-2.3.9.tgz#ad5549322dfe9d153d4b4dd6f7ff2ae234b06e24"
+  integrity sha512-n8GuYSrI9bF7FFZ/SjhwevlHc8xaVlb/7HmHelnc/PZXBD2ZR49NnN9sMMuDdEGPeeRQ5d0hqlSlEpgCX3Wl0Q==
+
+"@csstools/color-helpers@^5.1.0":
+  version "5.1.0"
+  resolved "https://registry.yarnpkg.com/@csstools/color-helpers/-/color-helpers-5.1.0.tgz#106c54c808cabfd1ab4c602d8505ee584c2996ef"
+  integrity sha512-S11EXWJyy0Mz5SYvRmY8nJYTFFd1LCNV+7cXyAgQtOOuzb4EsgfqDufL+9esx72/eLhsRdGZwaldu/h+E4t4BA==
+
+"@csstools/css-calc@^2.1.4":
+  version "2.1.4"
+  resolved "https://registry.yarnpkg.com/@csstools/css-calc/-/css-calc-2.1.4.tgz#8473f63e2fcd6e459838dd412401d5948f224c65"
+  integrity sha512-3N8oaj+0juUw/1H3YwmDDJXCgTB1gKU6Hc/bB502u9zR0q2vd786XJH9QfrKIEgFlZmhZiq6epXl4rHqhzsIgQ==
+
+"@csstools/css-color-parser@^3.1.0":
+  version "3.1.0"
+  resolved "https://registry.yarnpkg.com/@csstools/css-color-parser/-/css-color-parser-3.1.0.tgz#4e386af3a99dd36c46fef013cfe4c1c341eed6f0"
+  integrity sha512-nbtKwh3a6xNVIp/VRuXV64yTKnb1IjTAEEh3irzS+HkKjAOYLTGNb9pmVNntZ8iVBHcWDA2Dof0QtPgFI1BaTA==
+  dependencies:
+    "@csstools/color-helpers" "^5.1.0"
+    "@csstools/css-calc" "^2.1.4"
+
+"@csstools/css-parser-algorithms@^3.0.5":
+  version "3.0.5"
+  resolved "https://registry.yarnpkg.com/@csstools/css-parser-algorithms/-/css-parser-algorithms-3.0.5.tgz#5755370a9a29abaec5515b43c8b3f2cf9c2e3076"
+  integrity sha512-DaDeUkXZKjdGhgYaHNJTV9pV7Y9B3b644jCLs9Upc3VeNGg6LWARAT6O+Q+/COo+2gg/bM5rhpMAtf70WqfBdQ==
+
+"@csstools/css-syntax-patches-for-csstree@^1.0.21":
+  version "1.0.21"
+  resolved "https://registry.yarnpkg.com/@csstools/css-syntax-patches-for-csstree/-/css-syntax-patches-for-csstree-1.0.21.tgz#1671e2d9d7c800c5c5f65f8c7ab6722c5233b0b9"
+  integrity sha512-plP8N8zKfEZ26figX4Nvajx8DuzfuRpLTqglQ5d0chfnt35Qt3X+m6ASZ+rG0D0kxe/upDVNwSIVJP5n4FuNfw==
+
+"@csstools/css-tokenizer@^3.0.4":
+  version "3.0.4"
+  resolved "https://registry.yarnpkg.com/@csstools/css-tokenizer/-/css-tokenizer-3.0.4.tgz#333fedabc3fd1a8e5d0100013731cf19e6a8c5d3"
+  integrity sha512-Vd/9EVDiu6PPJt9yAh6roZP6El1xHrdvIVGjyBsHR0RYwNHgL7FJPyIIW4fANJNG6FtyZfvlRPpFI4ZM/lubvw==
+
+"@esbuild/aix-ppc64@0.27.2":
+  version "0.27.2"
+  resolved "https://registry.yarnpkg.com/@esbuild/aix-ppc64/-/aix-ppc64-0.27.2.tgz#521cbd968dcf362094034947f76fa1b18d2d403c"
+  integrity sha512-GZMB+a0mOMZs4MpDbj8RJp4cw+w1WV5NYD6xzgvzUJ5Ek2jerwfO2eADyI6ExDSUED+1X8aMbegahsJi+8mgpw==
+
+"@esbuild/android-arm64@0.27.2":
+  version "0.27.2"
+  resolved "https://registry.yarnpkg.com/@esbuild/android-arm64/-/android-arm64-0.27.2.tgz#61ea550962d8aa12a9b33194394e007657a6df57"
+  integrity sha512-pvz8ZZ7ot/RBphf8fv60ljmaoydPU12VuXHImtAs0XhLLw+EXBi2BLe3OYSBslR4rryHvweW5gmkKFwTiFy6KA==
+
+"@esbuild/android-arm@0.27.2":
+  version "0.27.2"
+  resolved "https://registry.yarnpkg.com/@esbuild/android-arm/-/android-arm-0.27.2.tgz#554887821e009dd6d853f972fde6c5143f1de142"
+  integrity sha512-DVNI8jlPa7Ujbr1yjU2PfUSRtAUZPG9I1RwW4F4xFB1Imiu2on0ADiI/c3td+KmDtVKNbi+nffGDQMfcIMkwIA==
+
+"@esbuild/android-x64@0.27.2":
+  version "0.27.2"
+  resolved "https://registry.yarnpkg.com/@esbuild/android-x64/-/android-x64-0.27.2.tgz#a7ce9d0721825fc578f9292a76d9e53334480ba2"
+  integrity sha512-z8Ank4Byh4TJJOh4wpz8g2vDy75zFL0TlZlkUkEwYXuPSgX8yzep596n6mT7905kA9uHZsf/o2OJZubl2l3M7A==
+
+"@esbuild/darwin-arm64@0.27.2":
+  version "0.27.2"
+  resolved "https://registry.yarnpkg.com/@esbuild/darwin-arm64/-/darwin-arm64-0.27.2.tgz#2cb7659bd5d109803c593cfc414450d5430c8256"
+  integrity sha512-davCD2Zc80nzDVRwXTcQP/28fiJbcOwvdolL0sOiOsbwBa72kegmVU0Wrh1MYrbuCL98Omp5dVhQFWRKR2ZAlg==
+
+"@esbuild/darwin-x64@0.27.2":
+  version "0.27.2"
+  resolved "https://registry.yarnpkg.com/@esbuild/darwin-x64/-/darwin-x64-0.27.2.tgz#e741fa6b1abb0cd0364126ba34ca17fd5e7bf509"
+  integrity sha512-ZxtijOmlQCBWGwbVmwOF/UCzuGIbUkqB1faQRf5akQmxRJ1ujusWsb3CVfk/9iZKr2L5SMU5wPBi1UWbvL+VQA==
+
+"@esbuild/freebsd-arm64@0.27.2":
+  version "0.27.2"
+  resolved "https://registry.yarnpkg.com/@esbuild/freebsd-arm64/-/freebsd-arm64-0.27.2.tgz#2b64e7116865ca172d4ce034114c21f3c93e397c"
+  integrity sha512-lS/9CN+rgqQ9czogxlMcBMGd+l8Q3Nj1MFQwBZJyoEKI50XGxwuzznYdwcav6lpOGv5BqaZXqvBSiB/kJ5op+g==
+
+"@esbuild/freebsd-x64@0.27.2":
+  version "0.27.2"
+  resolved "https://registry.yarnpkg.com/@esbuild/freebsd-x64/-/freebsd-x64-0.27.2.tgz#e5252551e66f499e4934efb611812f3820e990bb"
+  integrity sha512-tAfqtNYb4YgPnJlEFu4c212HYjQWSO/w/h/lQaBK7RbwGIkBOuNKQI9tqWzx7Wtp7bTPaGC6MJvWI608P3wXYA==
+
+"@esbuild/linux-arm64@0.27.2":
+  version "0.27.2"
+  resolved "https://registry.yarnpkg.com/@esbuild/linux-arm64/-/linux-arm64-0.27.2.tgz#dc4acf235531cd6984f5d6c3b13dbfb7ddb303cb"
+  integrity sha512-hYxN8pr66NsCCiRFkHUAsxylNOcAQaxSSkHMMjcpx0si13t1LHFphxJZUiGwojB1a/Hd5OiPIqDdXONia6bhTw==
+
+"@esbuild/linux-arm@0.27.2":
+  version "0.27.2"
+  resolved "https://registry.yarnpkg.com/@esbuild/linux-arm/-/linux-arm-0.27.2.tgz#56a900e39240d7d5d1d273bc053daa295c92e322"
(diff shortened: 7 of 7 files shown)


src/nodes/custom_action_text_attachment_node.js
```diff
@@ -56,7 +57,7 @@ export class CustomActionTextAttachmentNode extends DecoratorNode {
   }
 
   createDOM() {
-    const figure = createElement("action-text-attachment", { "content-type": this.contentType, "data-lexxy-decorator": true })
+    const figure = createElement(lexxyConfig.get("global.attachmentTagName"), { "content-type": this.contentType, "data-lexxy-decorator": true })
```
samuelpecher: What do you think about passing the tag name to the Node constructor and keeping as internal node state rather than hooking into the config at every step?

Same would apply to `ActionTextAttachmentNode`

src/config/configuration.js
```diff
@@ -0,0 +1,57 @@
...
+
+    if (into === this.#tree) {
+      this.#notify()
+    }
+
+    return into
+  }
+
+  listen(path, listener) {
+    this.#listeners[listener] = path
```
samuelpecher: This could be a `Map` rather than `Object` as the keys [will be coersed](https://262.ecma-international.org/6.0/#sec-object-type) to `String`

src/nodes/custom_action_text_attachment_node.js
zachasme: I like it! I noticed you did the same for `custom-upload-events`, and considered including it here. But I also wanted to keep this minimal.

If you are expecting to merge `custom-upload-events` soonish I'd rather not introduce it here, otherwise I'm ok to include it.

src/config/dom_purify.js
```diff
@@ -37,8 +38,10 @@ DOMPurify.addHook("uponSanitizeElement", (node, data) => {
   }
 })
 
-DOMPurify.setConfig({
-  ALLOWED_TAGS: ALLOWED_HTML_TAGS,
-  ALLOWED_ATTR: ALLOWED_HTML_ATTRIBUTES,
-  SAFE_FOR_XML: false // So that it does not strip attributes that contains serialized HTML (like content)
+lexxyConfig.listen("global.attachmentTagName", (attachmentTagName) => {
```
jorgemanrubia: Why are we observing changes here instead of just grabbing the configuration directly? Is this a matter of changing the initialization order or something like that?

src/elements/editor.js
```diff
@@ -398,7 +403,7 @@ export default class LexicalEditorElement extends HTMLElement {
   }
 
   get #hasToolbar() {
-    return this.getAttribute("toolbar") !== "false"
+    return Boolean(this.config.get("toolbar"))
```
jorgemanrubia: Is this keeping the default `true` value?

src/config/configuration.js
zachasme: Good point! Ended up ditching it in favor of nested arrays, in case we want to use the same listener for multiple paths.

src/config/configuration.js
```diff
@@ -0,0 +1,57 @@
...
+      node = node[key]
+      if (node === true || node === false) {
+        return node
+      }
+    }
+
+    return node
+  }
+
+  merge(config, into = this.#tree) {
```
jorgemanrubia: I think the `into` param is a bit confusing. Why would you like to change that to other thing? This is an instance method, it is expected it will act on its private data. I see you use it recursively but it kind of makes the interface less clear. I would consider extracting a `deepMerge` helper method to `helpers/hash_helper.js` and reuse here. You could move `arePlainObjects` as a private (not exported) function there too.

src/config/configuration.js
```diff
@@ -0,0 +1,57 @@
+export default class Configuration {
+  #tree
+  #listeners = []
+
+  constructor(initial = {}) {
+    this.#tree = initial
+  }
+
+  // returns bools early, null if key is missing
+  get(path = "") {
```
jorgemanrubia: why would path by the empty string? I would remove the default value.

src/nodes/custom_action_text_attachment_node.js
samuelpecher: I'm splitting that branch into separate PRs back to `main`, so I'll handle that then 👌

src/config/configuration.js
```diff
@@ -0,0 +1,57 @@
...
+    let node = this.#tree
+    const keys = path.split(".")
+    for (const key of keys) {
+      if (!Object.hasOwn(node, key)) {
+        return null
+      }
+
+      node = node[key]
+      if (node === true || node === false) {
+        return node
```
jorgemanrubia: I think this should be more agnostic about the boolean value. It should loop through the path and return whatever it is at the end. Why that return?

src/config/configuration.js
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    if (into === this.#tree) {
+      this.#notify()
+    }
+
+    return into
+  }
+
+  listen(path, listener) {
```
jorgemanrubia: As mentioned, the listening mechanism seems a bit excessive to me. This is configuration stuff. It should be initialized and used, but why would you like to subscribe to dynamic changes?

src/editor/configuration.js
```diff
@@ -0,0 +1,45 @@
...
+  #editorElement
+
+  constructor(editorElement) {
+    super()
+    this.#editorElement = editorElement
+    this.merge(this.#overrides)
+  }
+
+  get(path) {
+    if (this.#preset) {
```
jorgemanrubia: I would define a getter `get config` at the editor level, that returns the value of the attribute or `"default"` if the attribute is not set. Then, this can be deterministic. You just get the value of `editorElement.config`. You can rename the current `config` to `configuration` to remove the collision.

src/editor/configuration.js
```diff
@@ -0,0 +1,45 @@
...
+      return super.get(path) ?? lexxyConfig.get(`${this.#preset}.${path}`) ?? lexxyConfig.get(`default.${path}`)
```
jorgemanrubia: Why checking the path before prepending the prefix path segment? I think this should be more deterministic. It should invoke the parent with the path prepending the `editorElement.config` segment. 

I am not sure about the inheritance here. `Configuration` feels like a simple data structure you use. I think `EditorConfiguration` could have an attribute configuration that it uses to delegate the `get` and `set`, but inheriting from it feels slightly off to me.

src/editor/configuration.js
```diff
@@ -0,0 +1,45 @@
...
+    } else {
+      return super.get(path) ?? lexxyConfig.get(`default.${path}`)
+    }
+  }
+
+  get #overrides() {
+    const overrides = {}
+    for (const attribute of EditorConfiguration.attributes) {
```
jorgemanrubia: Instead of having to maintain the list of attributes in this place too, I would grab the list from the defaults `Object.keys(configuration.get("defaullt"))`. This also adds some basic protection against typos in configuration keys when overriding them.

src/config/configuration.js
```diff
@@ -0,0 +1,57 @@
...
+        return null
+      }
+
+      node = node[key]
+      if (node === true || node === false) {
+        return node
+      }
+    }
+
+    return node
```
jorgemanrubia: If the configuration key is missing, I wonder if this should throw an error. I think it should, to prevent entering typos when trying to override settings, which could be hard to debug.

src/nodes/custom_action_text_attachment_node.js
jorgemanrubia: This would be a quite edge case but if we kept it in the internal node state, it would be copied/pasted carrying the tag, which assumes that the editor where you are pasting the node has the same override. I would keep it simple with a global access I think.

src/editor/configuration.js
```diff
@@ -0,0 +1,45 @@
...
+    }
+    return overrides
+  }
+
+  get #preset() {
+    return this.#editorElement.getAttribute("config")
+  }
+
+  #parseAttribute(attribute) {
+    const value = this.#editorElement.getAttribute(attribute)
```
jorgemanrubia, line 38: You could use `JSON.parse` instead of the ad-hoc booelan parsing. That way, it will work for boolean value types supported by JSON. This could be handy when providing Rails attributes during rendering.

src/config/configuration.js
zachasme: I'll remove it 👍 was left in so I could easily get the entire configuration tree.

src/config/configuration.js
zachasme: The intention was to eventually support different levels of granularity such as
```js
{ toolbar: true }
// or
{
  toolbar: {
    bold: true
  }
}
```
Such that querying for "toolbar.bold" would return `true` in both cases.

src/config/configuration.js
zachasme: Hm, that's a good idea. Currently I'm returning `null` so I can detect and fallback to `preset`/`default` in `EditorConfigutaion`. But I could validate against the structure in `default` if we keep all possible configs in there.

src/config/dom_purify.js
zachasme: This is necessary if we want modify the `attachmentTagName` using `Lexxy.configure`. Otherwise `DOMPurify` will remain configured with the default value.

Alternatively, `DOMPurify.sanitize` accepts a second configuration argument. We could avoid listening for changes if we build a fresh configuration for each call.

src/editor/configuration.js
zachasme: > Why checking the path before prepending the prefix path segment? I think this should be more deterministic. It should invoke the parent with the path prepending the editorElement.config segment.

I wanted to support three possible levels of configuration: element attributes `toolbar=false`, editor presets `config=simple`, and defaults. Unless there is no preset selected in which case we only check element attributes and defaults.

> I am not sure about the inheritance here. Configuration feels like a simple data structure you use. I think EditorConfiguration could have an attribute configuration that it uses to delegate the get and set, but inheriting from it feels slightly off to me.

Hm, yes, I'll try that!

src/editor/configuration.js
zachasme: Ooh nice! 👌

src/elements/editor.js
zachasme: Yes. The `Boolean` call technically isn't required, but it felt right since you can pass an element ID to `toolbar`.

zachasme:
Thank you for the feedback @samuelpecher and @jorgemanrubia! 👌

> We need to add tests here. So far, Capybara has served us well. But here, to test the configuration interfaces and such, I think we could add jest to Lexxy. I would still rely on Capybara as our main testing mechanism, but it is nice to have more traditional unit testing available too. We could configure in the local CI script and in GH actions too. That would let us test the Configuration objects properly here.

Agreed. I wanted to hear your thoughts on the general approach before writing new tests.

> I would still add some high level system tests to make sure the configurations work as expected. I would be fine with just testing the overrides in capybara, assuming the system is tested through jest. WDYT?

Yes, I think adding jest is justified at this point. Though I'd rather tackle that in a separate PR. Until we introduce more granular options for toolbar/etc, I believe I can cover most of the configuration system with some high-level system tests.

jorgemanrubia:
To me, the motivation of jest would be to test the system we are bootstrapping here, more than the granular options themselves. I think those tests belong to this PR.

src/config/dom_purify.js
samuelpecher: I was thinking of mirroring Trix's implementation: call `setConfig` before each call. This allows a known good configuration and keeps the `sanitize` call clear.

Then we can 'pull' config and not need the listeners

samuelpecher:
> Good spike @zachasme!
> 
> We need to add tests here. So far, Capybara has served us well. But here, to test the configuration interfaces and such, I think we could add jest to Lexxy.

If you're open to it I recommend [vitest](https://vitest.dev/) which is powered by the rapid `vite` which uses esbuild

zachasme:
> To me, the motivation of jest would be to test the system we are bootstrapping here, more than the granular options themselves. I think those tests belong to this PR.

Makes sense. Alright, I'll set up jest here. @samuelpecher I'll give `vitest` a spin, it's "jest-compatbile" so why not? :+1:

src/nodes/custom_action_text_attachment_node.js
samuelpecher: @jorgemanrubia: the exchange format will be `text/html` unless the editors share the same namespace. The rendered HTML will include the rendered tag name and  I think we should match that behavior.

Though I do propose that differently configured Lexxy editors shouldn't share a namespace and we should [set it](https://github.com/basecamp/lexxy/blob/HEAD/src/elements/editor.js#L199-L199) depending on the loaded config:
```js
   // ...
   namespace: `Lexxy.${configName}`,
```

app/assets/javascript/lexxy.js
```diff
@@ -9802,12 +9915,16 @@ class LexicalEditorElement extends HTMLElement {
     return this.isEmpty || this.toString().match(/^\s*$/g) !== null
   }
 
+  get preset() {
```
jorgemanrubia, line 10175: Since you configure the element with `config=` it makes sense to me that this is `get config` instead of using a new name. We do that with other attributes lke name. Feels more consistent. Any reason to use a different name here?

src/editor/configuration.js
```diff
@@ -0,0 +1,50 @@
...
+import lexxyConfig from "../config/lexxy"
+import { dasherize } from "../helpers/string_helper"
+
+export default class EditorConfiguration {
+  #editorElement
+  #config
+
+  constructor(editorElement) {
+    this.#editorElement = editorElement
+    this.#config = new Configuration(this.#overrides)
```
jorgemanrubia: I think like this should be something like

```js
this.#config = new Configuration(lexxyConfig.get("default"))
this.#config.merge(this.#overrides)
```

So that `this.#overrides` only returns the set of properties overridden by the editor. And then, you can use `this.#config` deterministically. I think the `get` having to check whether it is an override or not is a smell . Our config data structure allow overrides via merge, we should leverage that here, I believe.

app/assets/javascript/lexxy.js
zachasme: No reason, they should definitely be consistent. I've been back and forth on config/preset. I prefer `preset`, it feels more descriptive than the generic `config`.

src/editor/configuration.js
zachasme: In my mind the editor configuration would reflect any changes made to defaults/presets after editor instantiation, hence the reference directly in `get`.

But I realize now this is probably *worse*, not being able to assume a consistent config during an editor lifecycle.

Removing the runtime look-ups should make things much simpler!

app/assets/javascript/lexxy.js
jorgemanrubia: But `config` is what we expose as the public API to be used by the editor users. What is confusing to me is using different terms here:

```html
<lexxy-editor config="whatever"
```

I would expect that if you do `editorElement.config` returns `whatever`. Now, we want to make that, if config is not set, then its default value is `default`, so that you can query `editorElement.config` consistently in any scenario. See what I mean?

We could also change the public api method to `preset`, but I think it is important that things remain consistent here. I like `config` slightly better because we are talking about configurations when defining these.

src/config/dom_purify.js
jorgemanrubia: Sorry I still don't understand.

We configure dompurify when running:

```js
import "./config/dom_purify"
```

Wouldn't this be a matter of requiring the global configuration before that code runs, and then making dom purify use the global configuration?

app/assets/javascript/lexxy.js
jorgemanrubia: Maybe we could go with *presets* indeed at the configuration level. So the lexxy configuration could expose two different structures:

```js
{
  global: new Configuration(...)
  presets: new Configuration({
    default: {
    }
   ...
  })
}
```

So that we use `preset` at the public API (setter) and as a getter in the element.

And we make global a explicit configuration object to query like `config.global.get("tagName")` or whatever. And same for the presets: `config.presets.get("default.whatever")` or just `config.presets.default` to grab the properties we want for a preset.

jorgemanrubia: So, for creating a editor-specific config we could do:

```js
this.#config = new Configuration(lexxyConfig.presets.default)
this.#config.merge(this.#overrides)
```
Thinking we could even extend the configuration constructor to admit a list of configurations (`...configurations`) that it will merge to avoid the explicit merge:

```js
this.#config = new Configuration(lexxyConfig.presets.default, this.#overrides)
```

I think that is pretty clean. The code expresses what we are trying to do in a single line.

src/config/configuration.js
jorgemanrubia: Hmm not sure about this, it does not look super intuitive. I think it is important that we keep the core config abstraction lean and generic. I would rather interpret this at toolbar-rendering time: assume that everything is on unless there is a FALSE in the settings.

src/editor/configuration.js
jorgemanrubia: Yeah, I think it is totally fine to assume that the properties are grabbed when the editor is instantiated, and subsequent changes won't be grabbed. That seems like a reasonable expectation to establish.

src/elements/editor.js
```diff
@@ -126,7 +133,7 @@ export default class LexicalEditorElement extends HTMLElement {
   get value() {
     if (!this.cachedValue) {
       this.editor?.getEditorState().read(() => {
-        this.cachedValue = sanitize($generateHtmlFromNodes(this.editor, null))
+        this.cachedValue = sanitize($generateHtmlFromNodes(this.editor, null), { additionalAllowedTags: [ lexxyConfig.get("global.attachmentTagName") ] })
```
jorgemanrubia: If this is set globally for dompurify, we don't need to do this right?

src/nodes/action_text_attachment_node.js
```diff
@@ -17,7 +18,7 @@ export class ActionTextAttachmentNode extends DecoratorNode {
 
   static importDOM() {
     return {
-      "action-text-attachment": (attachment) => {
+      [lexxyConfig.get("global.attachmentTagName")]: (attachment) => {
```
jorgemanrubia: Checking these usages, I think making `global` a dedicated conttainer would read slightly better. And I think it is right to make `global` its own *thing*.

```js
lexxyConfig.global.get("attachmentTagName")
```

src/test/config/configuration.test.js
```diff
@@ -0,0 +1,13 @@
+import { expect, test } from "vitest"
+import Configuration from "../../config/configuration"
+
+test("gets path", () => {
+  const c = new Configuration({
```
jorgemanrubia: We normally avoid single-letter identifiers. Better to go with `config` or `configuration`.

src/test/editor/configuration.test.js
```diff
@@ -0,0 +1,34 @@
+import { expect, test, vi } from "vitest"
```
jorgemanrubia: Great to have JS tests available now 👍. Please configure an entry in Github actions and our local ci setup to run them as part of our pipeline.

src/test/editor/configuration.test.js
```diff
@@ -0,0 +1,34 @@
...
+import EditorConfiguration from "../../editor/configuration"
+import lexxyConfig from "../../config/lexxy"
+
+test("uses defaults", () => {
+  lexxyConfig.merge({
+    default: { singleLine: true }
+  })
```
jorgemanrubia: Here we are modifying a global data structure as part of a single test run, which doesn't look great. Also, these tests are stubbing some internals. Ideally we would test the real thing here.

I think we could modify `EditorConfiguration` to receive as a second argument the `presetConfiguration` to use (with a default) so that, each test can use a fresh data structure for the defaults. I think these should be more like a black box: I have *these* defaults and I pass *this* element. I think we could also use jsdom to create a proper dom element from some HTML string for testing this? That way we could get rid of stubbing `hasAttribute`, which is an internal implementation detail.

zachasme:
Thanks for the thorough feedback @jorgemanrubia! The PR might have been a bit too WIP, but the responses here are really clearing things up for me. It's slowly coming together nicely. 

I think this is an important discussion:
```js
{ toolbar: true }
// or
{
  toolbar: {
    bold: true
  }
}
```

>  Hmm not sure about this, it does not look super intuitive. I think it is important that we keep the core config abstraction lean and generic. I would rather interpret this at toolbar-rendering time: assume that everything is on unless there is a FALSE in the settings.

I agree we should assume everything is on by default, fully defined by the default preset. But I also think we should allow users to 

```js
configure({ default: { toolbar: false }})
```

In cases where a bool is overriding a hash, I think it would be valuable to recursively write the bool into nested hashes. It will complicate the merge logic, but it will make `get` super simple, we can more easily validate reads/writes and consumers of the config can rely on a known structure.

Alternatively, we can keep the config class completely generic and rely on the rest of the codebase to handle these differences.

> We could also change the public api method to `preset`, but I think it is important that things remain consistent here. I like `config` slightly better because we are talking about configurations when defining these.

I think they both have merit. In isolation `config=` makes sense for the public API, it matches `configure`. But internally `preset` is more descriptive, and we avoid having both `config` and `configuration` referring to separate things. I also think it makes sense to introduce this as `presets` in the docs rather than configs.

> Great to have JS tests available now 👍. Please configure an entry in Github actions and our local ci setup to run them as part of our pipeline.

I've also changed our workflow .yml to run on all pushes. Was there a specific reason we previously had:
```yml
on:
  pull_request:
  push:
    branches: [ main ]
```

src/test/dom_helper.js
```diff
@@ -0,0 +1,5 @@
+import { JSDOM } from "jsdom"
```
jorgemanrubia: I think I would prefer to keep `test/` as a first level folder, parallel to `src/`. I think it is cleaner, so that it doesn't look like another system inside the source. I see that in trix it was inside, but in trix we also had a top level `src/trix` folder that we don't have here. 

Also, I would move the `dom_helper.js` to a folder `test/helpers`.

src/test/editor/configuration.test.js
```diff
@@ -0,0 +1,38 @@
...
+    attachments: true
+  },
+  simple: {
+    toolbar: false
+  }
+})
+
+test("uses defaults", () => {
+  const element = createElement("<lexxy-editor></lexxy-editor>")
+  const config = new EditorConfiguration(element, "default")
```
jorgemanrubia: With the new implementation, we don't have the problem I mentioned of having tests altering the same global structure over and over. I would remove the second param (`preset`) in the editor configuration constructor, since we don't need to override the config for testing purposes anymore. In the tests, I would insert that preset at the element level `createElement("<lexxy-editor preset="simple"></lexxy-editor>")` or just omit it there if you want to pass the default.

```js
  const element = createElement("<lexxy-editor></lexxy-editor>")
  const config = new EditorConfiguration(element)

// or, for a specific one:

  const element = createElement("<lexxy-editor preset="simple"></lexxy-editor>")
  const config = new EditorConfiguration(element)
```

That way we would be testing the real thing.

src/editor/configuration.js
```diff
@@ -0,0 +1,40 @@
...
+    )
+  }
+
+  get(path) {
+    return this.#config.get(path)
+  }
+
+  get #overrides() {
+    const overrides = {}
+    for (const option of Object.keys(DEFAULT_PRESET)) {
```
jorgemanrubia: I would rather keep the `DEFAULT_PRESET` as an internal constant and keep the configuration interface as the only way of interacting with the configurations. So I would use `lexxyConfig.presets.get("default")` instead of depending on the constant.

src/config/lexxy.js
```diff
@@ -0,0 +1,21 @@
...
+}
+
+const configuration = {
+  global: new Configuration({
+    attachmentTagName: "action-text-attachment"
+  }),
+  presets: new Configuration({ default: DEFAULT_PRESET })
+}
+
+export function configure(changes) {
```
jorgemanrubia: Since this is a function to import, maybe `configureLexxy` instead of just `configure` would be more convenient.

The thing is that I would prefer an OO interface, so that you would do:

```js
Lexxy.configure(...)
Lexxy.config.global
Lexxy.config.presets.get(...)
```

I think that feels more consistent and offers a proper namespace for other stuff we may add down the road. This way also you don't have to figure out how to rename lexxy functions and identifiers: `lexxyConfig`, `lexxyConfigure`... they are properly namespaced.

src/elements/editor.js
```diff
@@ -40,6 +41,7 @@ export default class LexicalEditorElement extends HTMLElement {
 
   connectedCallback() {
     this.id ??= generateDomId("lexxy-editor")
+    this.config = new Configuration(this, this.preset)
```
jorgemanrubia: Mentioned in other place, but passing an object and an attribute of that object as params feel off. It is an editor configuration: receiving the editor element should be enough. Notice the original mechanism I had suggested for tests was intended for tests only, as a mechanism to override the default param.

src/elements/editor.js
jorgemanrubia: I still think it is redundant. Isn't JS semantics enough for evaluating true/false enough here? I would only add additional code here if it is really justified. It confuses me a bit whenever I see it.

src/helpers/string_helper.js
```diff
@@ -1,3 +1,11 @@
+export function camelize(value) {
```
jorgemanrubia: this function is not used right? I think we could nix

src/editor/configuration.js
```diff
@@ -0,0 +1,40 @@
...
+import { dasherize } from "../helpers/string_helper"
+
+export default class EditorConfiguration {
+  #editorElement
+  #config
+
+  constructor(editorElement, preset = "default") {
+    this.#editorElement = editorElement
+    this.#config = new Configuration(
+      lexxyConfig.presets.get("default"),
```
jorgemanrubia: Mentioned above: I would remove the `preset` second argument here. It is an editor config that receives an editor element, totally fine to query the editor element for its preset now:

```js
    this.#config = new Configuration(
      lexxyConfig.presets.get("default"),
      lexxyConfig.presets.get(editorElement.preset),
      this.#overrides
    )
```

jorgemanrubia:
> Alternatively, we can keep the config class completely generic and rely on the rest of the codebase to handle these differences.

I would wait to see if we need the extra complication. If someone sets that the toolbar is `false`, the editor won't configure any toolbar, so who is going to check nested toolbar properties? I would lean towards simple and generic unless the are specific reasons. Maybe there are specific scenarios where we need this right away, but what are those?

jorgemanrubia:
On @zachasme and we also need to document the new configuration API in the README 🙏

src/test/dom_helper.js
zachasme: I've moved them into `test/javascript/`. 👍 

My thinking was we wanted them separate from the rails tests, but I agree this is cleaner.

src/test/editor/configuration.test.js
zachasme: Indeed. I kept the param because I had trouble getting jsdom to play nice with custom elements. I found a solution, though, and the param is gone.

src/config/lexxy.js
zachasme: If we're talking public interface, we could keep the existing named exports and also provide a convenience default export, so the following would work the same:
```js
// same as
import Lexxy from "lexxy"
// usage
Lexxy.configure(...)
Lexxy.highlightAll()
```
We could also assign the same object to a global `window.Lexxy`. What do you think?

In terms of internal usage, how about exporting the following from `config/lexxy.js`?
```js
export default {
  global,
  presets,
  configure(changes) {
    return presets.merge(changes)
  }
}
// usage
import Lexxy from "./config/lexxy"
Lexxy.global.get(..)
```
And then re-export `Lexxy.configure` for the public interface.

zachasme:
> I would lean towards simple and generic unless the are specific reasons. Maybe there are specific scenarios where we need this right away, but what are those?

Good point, let's keep it simple.

> we also need to document the new configuration API in the README 🙏

Almost there!

zachasme:
Further refinements:
- Documentation.
- I've stripped down Configuration objects to only `merge` and naive `get`, no early returns for bools.
- Support `toolbar` config:
	- Previously the `toolbar` element attribute accepted dom id for external toolbars. I've renamed it to `toolbartarget` (inspired by `popovertarget`), such that `toolbar` config is always boolean.
	- I've created a test for external toolbars. I had to change the implementation to construct a full `<lexxy-toolbar>` inside the provided target. Is this how you intended it to work?
- Support `markdown` config:
	- I'm falling back to stock Lexical rich text paste when markdown is disabled. 
	- I attempted to change our `PASTE_COMMAND` handler to skip `preventDefault` if no special paste branch is taken. But that required replacing the asynchronous `DataTransferItem.getAsString` with synchronous `DataTransfer.getData`. Turns out we rely on asynchronous handling (or at least the testsuite does), and comments indicate some special care needs to taken for safari, so instead I fell back to directly calling `$insertDataTransferForRichText` and not touch existing paste logic when markdown is enabled.
- Support `multiline` config
	- The previous option was called `singleLine`. Since we prefer to have all default options enabled by default, it made more sense have a `multiline` option.

src/editor/clipboard.js
```diff
@@ -70,8 +71,10 @@ export default class Clipboard {
       } else if (isUrl(text)) {
         const nodeKey = this.contents.createLink(text)
         this.#dispatchLinkInsertEvent(nodeKey, { url: text })
-      } else {
+      } else if (this.editorElement.supportsMarkdown) {
         this.#pasteMarkdown(text)
+      } else {
+        this.#pasteRichText(clipboardData)
```
jorgemanrubia, line 77: nice 👏

src/elements/editor.js
```diff
@@ -415,19 +427,24 @@ export default class LexicalEditorElement extends HTMLElement {
...
+    }
   }
 
   get #hasToolbar() {
-    return this.getAttribute("toolbar") !== "false"
+    return this.config.get("toolbar")
   }
 
-  #createDefaultToolbar() {
+  #createDefaultToolbar(container = this) {
```
jorgemanrubia: Why this new parameter? `#findOrCreateDefaultToolbar` was tighter before. Why expanding the container and pass it as a param now?

README.md
```diff
@@ -142,15 +142,45 @@ Under the hood, this will insert a `<lexxy-editor>` tag, that will be a first-cl
...
+  simple: {
+    attachments: false
+  }
+})
+<lexxy-editor preset="simple"></lexxy-editor>
+
+// you can override specific options with attributes on editor elements
+<lexxy-editor preset="simple" attachments="true"></lexxy-editor>
+```
+
```
jorgemanrubia, line 176: This should include a reference to `Lexxy.global` and the option to configure tag names, right?

jorgemanrubia (commented):
Looking tight @zachasme 👏! Couple of minor comments.

src/elements/editor.js
zachasme: I'm using it to optionally construct the default toolbar inside an external container, if provided:

> Previously the `toolbar` element attribute accepted dom id for external toolbars. I've renamed it to `toolbartarget` (inspired by `popovertarget`), such that `toolbar` config is always boolean.
>
> I've created a test for external toolbars. I had to change the implementation to construct a full `<lexxy-toolbar>` inside the provided target. Is this how you intended it to work?

I wanted to add a test for external toolbars, and couldn't make it work without building the full default toolbar. But I'm possibly misunderstanding how external toolbars should work.

README.md
zachasme: Right! ✅

What do you think of the public interface? Right now I only expose `Lexxy.configure` which re-routes it's input to `global`/`presets`.

```js
configure({ global: newGlobal, ...newPresets }) {
  if (newGlobal) {
    global.merge(newGlobal)
  }
  presets.merge(newPresets)
}
```

jorgemanrubia: I think it is good 👍

src/elements/editor.js
jorgemanrubia: @zachasme if a toolbar ID is provided, then Lexical should just grab that toolbar. So we just need to keep the previous behavior/code there.

I see that you changed `toolbar` to `toolbarTarget`. That is a breaking change here, I would just stick to `toolbar`. 

I think the semantic should be:

- If `false`: no toolbar
- If `true`: default toolbar
- If other string: reuse that toolbar directly (existing behavior)

Does that make sense to you?

zachasme: Makes total sense. Splitting `toolbar` into `toolbar`/`toolbartarget` was just personal preference, I'll revert that part.

That is unrelated to the new parameter mentioned above, though. I added that to make the following pattern work, which currently fails on `main`:

```js
<div id="external"></div>
<lexxy-editor toolbar="external">
// Uncaught TypeError: this.toolbarElement.setEditor is not a function
```

I assumed it a bug that we weren't constructing a full toolbar inside the external element (README says *"or pass an element ID to render the toolbar in an external element"*).

jorgemanrubia: Oh but the toolbar should be a lexxy-toolbar element, not a regular div

jorgemanrubia: BTW good point that this is confusing. We can mention in the README that the expected element is a `lexxy-toolbar`

zachasme: I tried that as well but got another error.

```js
<lexxy-toolbar id="external"></lexxy-toolbar>
<lexxy-editor toolbar="external">
// Uncaught TypeError: Cannot read properties of null (reading 'children')
```

I could only make it work by providing a full toolbar directly in html:
```html
<lexxy-toolbar id="external">
  <button class="lexxy-editor__toolbar-button" type="button" name="bold" data-command="bold" title="Bold">
    <svg ...
  ...
</lexxy-toolbar>
<lexxy-editor toolbar="external">
```

Is this how it's meant to be used?

jorgemanrubia: Yes: it is meant to let you define a fully customizable toolbar as an external element and reference it.

zachasme: I see! Thanks for clearing that up 👍 I'll revert, update the README and test the intended usage instead.

PR #505, opened 2025-12-18
Tables

zoltanhosszu:
## Tables in Lexxy
This PR introduces advanced support for adding and managing tables.

image

Most of the underlying architecture relies on the [Lexical's Table package](https://lexical.dev/docs/packages/lexical-table). A new `table_handler` class is introduced to show the floating controls above the table.

### Features
A new Insert Table button is added to the Toolbar, that adds a simple 3x3 table in the editor.

When a table is selected, table controls are displayed above to manage the table's structure:

- Quickly add or remove rows and columns with [+] [-] buttons
- Add row above or below
- Add column before or after
- Delete a row/column
- Toggle header style on a row/column
- Delete the whole table


zoltanhosszu:
**Worthy additions to consider to this PR before merging:**
- [x] Keyboard support for the table control element. Currently it's only controllable with a pointing device, need to figure out a way to move the focus to the control element
- [ ] Support for _CSV file upload → table_
- [ ] Support for _CSV paste from clipboard → table_

src/elements/editor.js
```diff
@@ -320,10 +325,16 @@ export default class LexicalEditorElement extends HTMLElement {
...
     registerHistory(this.editor, this.historyState, 20)
     registerList(this.editor)
+    this.#registerTableComponents()
     this.#registerCodeHiglightingComponents()
     registerMarkdownShortcuts(this.editor, TRANSFORMERS)
   }
 
+  #registerTableComponents() {
+    registerTablePlugin(this.editor)
+    this.append(createElement("lexxy-table-handler"))
```
jorgemanrubia: Because editors can be connected and disconnected, you should make sure we remove the handler when the editor gets disconnected, or it will be duplicated:

1. grab the element in a variable like `this.tableHandler = ...`
2. In `#reset()`, do `this.tableHandler?.remove()`

src/elements/table_handler.js
```diff
@@ -0,0 +1,480 @@
...
+  $getTableRowIndexFromTableCellNode,
+  $insertTableColumnAtSelection,
+  $insertTableRowAtSelection,
+  $isTableCellNode,
+  TableCellHeaderStates
+} from "@lexical/table"
+import { $getSelection, $isRangeSelection } from "lexical"
+import { createElement } from "../helpers/html_helper"
+
+export class TableHandler extends HTMLElement {
```
jorgemanrubia, line 19: Love how this contains all our table handling 👏

test/test_helpers/editor_helper.rb
```diff
@@ -51,4 +51,30 @@ def wait_for_editor
...
+
+  def click_table_handler_button(aria_label)
+    find(".lexxy-table-handle-buttons button[aria-label='#{aria_label}']").click
+    sleep 0.1
+  end
+
+  def open_table_more_menu
+    more_menu = find(".lexxy-table-handle-buttons details.lexxy-table-control__more-menu")
+    more_menu.click
+    sleep 0.1
```
jorgemanrubia: Sometimes we can't avoid them, but in general it is better to avoid sleep. You could do `assert_css "some selector"`. And Capybara will wait a few seconds for that selector to exist before failing. This is a better alternative to active waits. I think you could handle both sleeps that way.

src/elements/table_handler.js
```diff
@@ -0,0 +1,480 @@
...
+  #setUpButtons() {
+    this.buttonsContainer = createElement("div", {
+      className: "lexxy-table-handle-buttons"
+    })
+
+    const rowButtonsContainer = this.#createRowButtonsContainer()
+    this.buttonsContainer.appendChild(rowButtonsContainer)
+
+    const columnButtonsContainer = this.#createColumnButtonsContainer()
+    this.buttonsContainer.appendChild(columnButtonsContainer)
```
jorgemanrubia: You could inline both constants, and reduce these 5 lines to 2 with just the `appendChild` sentences.

src/elements/table_handler.js
```diff
@@ -0,0 +1,480 @@
...
+
+  #deleteTableRow(direction) {
+    this.#manageTable("delete", "row", direction)
+  }
+
+  #deleteTableColumn(direction) {
+    this.#manageTable("delete", "column", direction)
+  }
+
+  #manageTable(action = "insert", childType = "row", direction) {
```
jorgemanrubia: minor but instead of `manageTable` I would look for something more assertive like `#executeTableCommand`

src/elements/table_handler.js
```diff
@@ -0,0 +1,480 @@
...
+      const firstCell = $getTableCellNodeFromLexicalNode(cells[0])
+      if (!firstCell) return
+
+      const currentStyle = firstCell.getHeaderStyles()
+      const newStyle = currentStyle ^ TableCellHeaderStates.ROW
+
+      cells.forEach(cell => {
+        const tableCellNode = $getTableCellNodeFromLexicalNode(cell)
+
+        if (tableCellNode) {
```
jorgemanrubia: could extract a `#setCellHeaderStyle(cell, newStyle)` method here

src/elements/table_handler.js
```diff
@@ -0,0 +1,480 @@
...
+
+      const cells = row.getChildren()
+      const selectedCell = $getTableCellNodeFromLexicalNode(cells[this.#currentColumn])
+      if (!selectedCell) return
+
+      const currentStyle = selectedCell.getHeaderStyles()
+      const newStyle = currentStyle ^ TableCellHeaderStates.COLUMN
+
+      rows.forEach(row => {
+        const cell = row.getChildren()[this.#currentColumn]
```
jorgemanrubia, line 437: Same, could extract a private method for the action executed inside the loop

src/elements/table_handler.js
```diff
@@ -0,0 +1,480 @@
...
+    const relativeCenter = (tableRect.left + tableRect.right) / 2 - editorRect.left
+    this.buttonsContainer.style.top = `${relativeTop}px`
+    this.buttonsContainer.style.left = `${relativeCenter}px`
+  }
+
+  #setTableFocusState(focused) {
+    this.#editorElement.querySelector("div.node--selected:has(table)")?.classList.remove("node--selected")
+
+    if (focused) {
+      if (!this.currentTableNode) return
```
jorgemanrubia: could do `if (focused && this.currentTablenode)` and remove the guard return completely.

src/elements/table_handler.js
```diff
@@ -0,0 +1,480 @@
...
+    this.#editor.update(() => {
+      const currentCell = this.#currentCell
+      if (!currentCell) return
+
+      if (direction === "end") {
+        this.#selectLastTableCell()
+      }
+
+      switch (action) {
+        case "insert":
```
jorgemanrubia: I was not sure about this method, and explored some ways of compacting it, but the implementation makes sense and is quite clear 👍.

src/elements/table_handler.js
```diff
@@ -0,0 +1,480 @@
...
+        case "delete":
+          switch (childType) {
+            case "row":
+              $deleteTableRowAtSelection()
+              break
+            case "column":
+              $deleteTableColumnAtSelection()
+              break
+          }
+      }
```
jorgemanrubia: could consider extracting a method that contains the `case` statement. Something like `#dispatchTableCommand` that receives the params it needs. That way, the parent method will be composed of methods all working at a similar level of abstraction.

test/system/table_test.rb
```diff
@@ -0,0 +1,132 @@
...
+    more_menu = open_table_more_menu
+
+    click_table_handler_button("Toggle column style")
+
+    html = find_editor.value
+    th_in_first_col_after = html.scan(/<tr>.*?<th\b/m).count
+    assert_not_equal initial_th_in_first_col, th_in_first_col_after, "Column header style should have changed"
+
+    more_menu.click
+    sleep 0.1
```
jorgemanrubia: We should try to get rid of these active sleeps @zoltanhosszu, they are not a good practice. If we really need to wait we should try to assert some selector. Let me know if you need a hand here, I can take care of that.

test/system/table_test.rb
```diff
@@ -0,0 +1,132 @@
...
+
+    assert_match(/<table>/, find_editor.value)
+
+    open_table_more_menu
+
+    click_table_handler_button("Delete table")
+
+    html = find_editor.value
+    assert_no_match(/<table>/, html, "Table should be removed")
+  end
```
jorgemanrubia, line 116: We need to add tests that save a table and that assert that the table can be rendered with action text. Check this one for reference:

https://github.com/basecamp/lexxy/blob/2d8d9ea227ca5e753f3a4cc4b253baa79957259f/test/system/color_highlighter_test.rb#L29-L36

So save the post and, in the screen you land, which shows the post, assert that all the table elements are there. This will detect problems with sanitizing things in action text and such.

test/system/table_test.rb
```diff
@@ -0,0 +1,132 @@
...
+    wait_for_editor
+  end
+
+  test "adding a table" do
+    find_editor.toggle_command("insertTable")
+    sleep 0.1
+
+    html = find_editor.value
+    assert_match(/<table>/, html)
+    assert_table_structure(3, 3)
```
seanpdoyle: Is this an opportunity to utilize [Capybara.string](https://rubydoc.info/github/teamcapybara/capybara/master/Capybara.string) and [assert_selector](https://rubydoc.info/github/teamcapybara/capybara/master/Capybara/Node/Matchers#assert_selector-instance_method)?

```suggestion
    assert_selector Capybara.string(find_editor.value), :table do |table|
      assert_selector table, :table_row, count: 3
    end
```

test/system/table_test.rb
```diff
@@ -0,0 +1,132 @@
...
+  test "writing in table fields" do
+    find_editor.toggle_command("insertTable")
+    sleep 0.1
+
+    find_editor.send "Test Cell"
+    sleep 0.1
+
+    html = find_editor.value
+    assert_match(/Test Cell/, html)
+    assert_match(/<table>.*?Test Cell.*?<\/table>/m, html)
```
seanpdoyle, line 21: Rather than manually sleeping, had you tried relying on Capybara's synchronization mechanisms that are built-in to its assertions?

```suggestion
    find_editor.toggle_command("insertTable")
    
    within find_editor do
      assert_selector :table

      find_editor.send "Test Cell"
    
      assert_selector :table, text: "Test Cell"
    end
```

zoltanhosszu:
@jorgemanrubia thank you for all the feedback! I have refactored the code based on your comments, and also did some changes on how the table commands are sent through the commands_dispatcher. 

Also added keyboard shortcuts to focus the table controller (_Cmd/Ctrl + Shift + F10_). Navigation within the control should be handled by https://github.com/basecamp/lexxy/pull/495 - will add it there, once this gets merged to `main`.

test/system/table_test.rb
jorgemanrubia: Thanks @seanpdoyle

Issue #494, opened 2025-12-09
Removing attachments with backspace is broken: '"_cachedNodes" is read-only'


OtherCroissant:
I have setup the prompts for selecting things with the hashtag (#) in a lexxy editor. Implementing this was super simple! Very nice feature!

Only stumbled upon an error when I wanted to remove the attachment by using the backspace. The steps:

<img width="699" height="254" alt="Image" src="https://github.com/user-attachments/assets/3098b78d-2a68-407c-ae11-7cf35245f80c" />

`<backspace>`

<img width="679" height="234" alt="Image" src="https://github.com/user-attachments/assets/fe5448fc-f07c-4f59-953a-768c39b5f502" />

`<backspace>`

JS Error: `"_cachedNodes" is read-only`

<img width="910" height="604" alt="Image" src="https://github.com/user-attachments/assets/2de4ee33-951f-493b-baff-2ca8234723d6" />

Or in text:

```
Uncaught TypeError: "_cachedNodes" is read-only
    getNodes http://localhost:3002/packs/js/vendors-node_modules_37signals_lexxy_dist_lexxy_esm_js-node_modules_hotwired_stimulus-webpack-f44598.js:48210
    node_modules 37signals/lexxy/dist/lexxy.esm.js/deleteSelectedNodes/<@http://localhost:3002/packs/js/vendors-node_modules_37signals_lexxy_dist_lexxy_esm_js-node_modules_hotwired_stimulus-webpack-f44598.js:2733
    $processNestedUpdates http://localhost:3002/packs/js/vendors-node_modules_37signals_lexxy_dist_lexxy_esm_js-node_modules_hotwired_stimulus-webpack-f44598.js:51022
    $beginUpdate http://localhost:3002/packs/js/vendors-node_modules_37signals_lexxy_dist_lexxy_esm_js-node_modules_hotwired_stimulus-webpack-f44598.js:51072
    updateEditorSync http://localhost:3002/packs/js/vendors-node_modules_37signals_lexxy_dist_lexxy_esm_js-node_modules_hotwired_stimulus-webpack-f44598.js:51153
    triggerCommandListeners http://localhost:3002/packs/js/vendors-node_modules_37signals_lexxy_dist_lexxy_esm_js-node_modules_hotwired_stimulus-webpack-f44598.js:50949
[...]

PR #507, opened 2025-12-19
Fix deleting selected attachments

bluesnotred:
Selecting an image and pressing Backspace/Delete throws:

TypeError: Cannot assign to read only property '_cachedNodes' of object '#<_NodeSelection>'

The cached `this.#selection.current` may be frozen after certain Lexical operations. Using fresh `$getSelection()` instead fixes this.

Related: https://github.com/facebook/lexical/issues/6290

### DIFF
--- a/src/editor/selection.js
+++ b/src/editor/selection.js
@@ -409,7 +409,12 @@ export default class Selection {
     if (this.current) {
       this.editor.update(() => {
         this.clear()
-        fn(this.current.getNodes()[0])
+        // Use fresh selection - cached this.current may be frozen
+        // See: https://github.com/facebook/lexical/issues/6290
+        const selection = $getSelection()
+        if ($isNodeSelection(selection)) {
+          fn(selection.getNodes()[0])
+        }
         this.editor.focus()
       })
     }


bluesnotred:
Added a second fix in 6fd45b8 - the `#withCurrentNode()` method in `selection.js` had the same frozen selection problem, causing errors when using arrow keys to navigate between attachments.

OtherCroissant:
#494

ksweetie:
I'm experiencing the same issue and can confirm that this PR fixes it for me. Thanks for fixing!

jorgemanrubia:
Hey @bluesnotred thanks for looking into this one. I think the problem here was related to having some sort of stale assets. I could reproduce with the version from main when building the assets with `yarn watch`, but then I ran `yarn build` and the error went away. I need to look into the reasons. Could you confirm if the last version fixes the problem for you? https://github.com/basecamp/lexxy/releases/tag/v0.1.25.beta

jorgemanrubia:
Will prevent this here https://github.com/basecamp/lexxy/pull/540

jorgemanrubia:
Tentatively closing @bluesnotred @ksweetie @OtherCroissant. Please let me know if you are still experiencing the issue with latest version.

OtherCroissant:
@jorgemanrubia I still have the same error deleting the attachments. Although I have the new table feature now, so the new version seems to be loaded?
Scherm­afbeelding 2026-01-05 om 14 40 00

Scherm­afbeelding 2026-01-05 om 14 40 47


I also don't see the html content of the lexxy editor in my console anymore so the development mode seems to be disabled correctly

jorgemanrubia:
Thanks @OtherCroissant. How are you installing Fizzy in your app? Using the gem or via the NPM package? I would like to reproduce this. I have tried to [update the gem in Fizzy](https://github.com/basecamp/fizzy/pull/2292) and not having issues there with deleting attachments.

jorgemanrubia:
Oh I see it is included via the NPM package. Maybe that is the key. I will look into it.

jorgemanrubia:
@OtherCroissant I just released a new version that should take care of this one

https://github.com/basecamp/lexxy/releases/tag/v0.1.27.beta

I would appreciate it if you can confirm. Thanks!

seanhogge:
Hi everyone. I was having this same issue in a project that has a LOT of JavaScript going on, so I don't know if my experience is pure enough to be useful.

However, 0.1.27-beta fixes this issue and I can now delete inserted attachments (images and mentions).

The work here is appreciated, we're hoping to use Lexxy as soon as it's viable.

jorgemanrubia:
Great thanks for confirming @seanhogge

OtherCroissant:
Works now for me as well!! Thanks!

bluesnotred:
Hey Jorge, thanks for looking into this. I've confirmed the fix works:
- Reproduced the _cachedNodes error on v0.1.23.beta again
- Verified v0.1.27.beta resolves the issue

Appreciate the thorough fix.

PR #510, opened 2025-12-23
Fix toolbar focus behavior to skip on tab navigation

italomatos:
Remove toolbar elements from tab order by setting tabindex=-1 on all focusable elements (buttons and summary elements). This ensures that when users press Tab to navigate to the editor, focus goes directly to the textarea instead of stopping at toolbar buttons first.

The toolbar remains fully functional via mouse clicks and keyboard shortcuts.


brunoprietog:
Hi @italomatos, the toolbar must also be accessible with the keyboard, not just with shortcuts or the mouse. This is already being addressed in #495, so I'm closing in favor of that one.

PR #539, opened 2026-01-05
Plain-text mode

zachasme:
The PR allows us to use Lexxy for "plain-text" editing. This is useful for search and other fields where we want to support prompts without formatting.

Passing `rich-text="false"` will register `@lexical/plain-text`, and disable `@lexical/rich-text`, lists, tables, code highlighting and markdown.

I went with `rich-text="false"` over `plain-text="true"` based on discussions in #504 (have all default options enabled by default).


jorgemanrubia (approved):
Nice one @zachasme. I guess we want to merge the config system first and handle this new option as part of the new system, right?

zachasme:
I'd actually prefer to merge this one right away, and get it into BC (for Search). The `config` branch I can just rebase.

Issue #508, opened 2025-12-19
Table header background colour is hard-coded using inline style tag


lylo:
In the Lexxy sandbox, the table headers have the following style:

```html
<th class="lexxy-content__table-cell--header"
    style="background-color: rgb(242, 243, 245);">
```

Is this inline style necessary? I'm not entirely sure where it's coming from. The `lexxy-content.css` file [contains this style](https://github.com/basecamp/lexxy/commit/ebadf6b549ee7533ad6c92de0dd3a3a88fc5b9bd#diff-5505d335b069b9a78da0eca1e0b49d6827715a44c971afc8df1e6d86a96c9926R223) which uses the `--lexxy-color-ink-lightest` variable:

```css
&.lexxy-content__table-cell--header {
  background-color: var(--lexxy-color-ink-lightest);
  font-weight: bold;
}
```

It would be better (for me and my app, at least!) if these inline styles were not present so we can rely on the Lexxy colour variables and easily override.

zoltanhosszu:
This style seems to come from [Lexical itself](https://github.com/facebook/lexical/blob/5bbbe849bd229e1db0e7b536e6a919520ada7bb2/packages/lexical-table/src/LexicalTableCellNode.ts#L187). Will check for a way to remove it, it bothers me too.

lylo:
Thank you for looking at this @zoltanhosszu 🙏

thooams:
A temporary solution:

```javascript
//  app/javascript/controllers/remove_th_style_controller.js
import { Controller } from "@hotwired/stimulus";

export default class extends Controller {
  connect() {
    this.removeThStyles();
  }

  removeThStyles() {
    const thElements = this.element.querySelectorAll("th");
    thElements.forEach((th) => {
      th.removeAttribute("style");
    });
  }
}
```

and in your html :
```html
<div data-controller="syntax-highlight remove-th-style" class="lexxy-content">...</div>
```

zoltanhosszu:
@thooams this should be fixed on `main` already. Are you still seeing the issue?

PR #541, opened 2026-01-05
Unstyled table header cells

zoltanhosszu:
Fix for https://github.com/basecamp/lexxy/issues/508

### DIFF
--- a/app/assets/stylesheets/lexxy-editor.css
+++ b/app/assets/stylesheets/lexxy-editor.css
@@ -64,7 +64,7 @@
         border-color: var(--lexxy-color-table-cell-selected-border);
       }
     }
-    
+ 
     &.lexxy-content__table--selection {
       ::selection {
         background: transparent;


src/helpers/table_helper.js
```diff
@@ -0,0 +1,12 @@
+import { TableCellNode } from "@lexical/table"
+
+// Prevent the hardcoded background color
+// A background color value is set by Lexical if background is null:
+// https://github.com/facebook/lexical/blob/5bbbe849bd229e1db0e7b536e6a919520ada7bb2/packages/lexical-table/src/LexicalTableCellNode.ts#L187
+export function registerTableCellTransform(editor) {
```
samuelpecher: Nice surgical use of a transform, I don't think there's a cleaner way to achieve this. It's certainly not worth sub-classing the entire Node.

What do you think about being more explicit about the transform's purpose ?Something like `registerHeaderBackgroundTransform`

zoltanhosszu:
@samuelpecher I renamed the function as you suggested, and did some CSS refactoring too. Would you be okay with merging this then?

app/assets/stylesheets/lexxy-editor.css
```diff
@@ -54,15 +54,17 @@
...
-    .lexxy-content__table-cell--selected {
-      background-color: var(--lexxy-color-table-cell-selected-bg) !important;
-      border-color: var(--lexxy-color-table-cell-selected-border) !important;
+      &.lexxy-content__table-cell--selected {
+        background-color: var(--lexxy-color-table-cell-selected-bg);
+        border-color: var(--lexxy-color-table-cell-selected-border);
+      }
     }
-  
+
```
samuelpecher:
```suggestion
 
```

PR #543, opened 2026-01-06
Add raw markdown mode toggle to toolbar

aviflombaum:
# Add Raw Markdown Support

  - Toggle between rich text and raw markdown editing via toolbar button
  - Uses Lexical's built-in $convertToMarkdownString/$convertFromMarkdownString
  - Button shows pressed state when in markdown mode
  - Form submit syncs markdown to HTML before submission
  - Handles Turbo cache by resetting to rich text mode
  
  
CleanShot 2026-01-06 at 13 24 57


aviflombaum:
There is a slight issue tbh. In theory, when you are in Raw Markdown mode, the Lexxy toolbar should still work by inserting proper markdown syntax. Once in raw markdown, the Lexxy toolbar won't add markdown syntax to the plain text. Maybe that's okay, but it is certainly something we could add.

richjdsmith:
Stoked to see this @aviflombaum. Hope it gets merged, or some version of it.

jorgemanrubia:
Thanks for the contribution! A raw markdown editing mode is interesting but goes beyond Lexxy's core scope as a rich text editor. This kind of feature would work well as an extension. Closing for now, but appreciate the work.

PR #546, opened 2026-01-08
Remove unnecessary markup

zoltanhosszu:
@samuelpecher noticed that the table handler and code language selector had empty DOM elements as we were manually adding control elements into the editor. This PR merges the controls and the controller elements:

Before:
```html
<lexxy-editor>
  ...
  <lexxy-table-handler></lexxy-table-handler>
  <lexxy-code-language-picker></lexxy-code-language-picker>
  <div class="lexxy-table-handle-buttons">[table controls]</div>
  <select class="lexxy-code-language-picker">[language select]</select>
</lexxy-editor>
```

After:
```html
<lexxy-editor>
  ...
  <lexxy-table-handler>[table controls]</lexxy-table-handler>
  <lexxy-code-language-picker>[language select]</lexxy-code-language-picker>
</lexxy-editor>
```


zoltanhosszu:
@jorgemanrubia I've got a flaky test locally in `test/system/prompts_test.rb` for line 35 `"hasOpenPrompt reports correct status"`. It seems like the issue is the async/await nature of prompt building, that causes this test to fail. Adding a sleep function could solve this, but it feels hacky. Any suggestions on that? Claude was no help...

PR #549, opened 2026-01-08
Highlight config and sanitization

samuelpecher:
This PR implements configuration for the highlighter for both the provided buttons as well as pasted style sanitization and canonicalization. 

`Highlighter`'s lexical functions have been packaged into a [Lexical extension](https://lexical.dev/docs/extensions/intro), registering the components according the the lifecycle phases according to the passed configuration. `Highlighter` remains to  apply the configuration from the editor to the extensions and expose the  light interface to the command dispatcher. Commands are routed to the editor via the new `TOGGLE_HIGHLIGHT_COMMAND`.

This suggests a three part structure to layer functionality:
- A high-level controller: `Highlighter`
- Low-level Lexical operations: `highlight_extension`
- UI in the toolbar and drop-downs.

 The extension structure allows collapsing the fake `HighlightNode` into just a `html.import` configuration. Particularly, the [change `TrixContentExtension` highlights this transition](https://github.com/basecamp/lexxy/compare/highlight-config-sanitization?expand=1#diff-95e54bb30f8b3c600734936528dae4b667dd8ad6ee52d7dfaee57c5e41435bd5) from fake Lexical node to basic extension. Low-level Lexical operations are then moved into `highlight_extension`.

Since `html` configuration is defined before non-global config is available, the sanitization step is deferred to a Node Transform through the use of Node State marking the node with a `hasPastedStyles` state. This has the double advantage that any node subsequently marked by other code leveraging `$applyHighlight` will benefit from the sanitization (such as `TrixContentExtension`).

Styles are sanitizes and canonicalized through an extension-scoped persistant `StyleCanonicalizer`. This optimization avoids repeating DOM element creation for calculating base RGB(A) values of the allowed styles, which is necessary as pastes have computed RGB(A) values. Lexical's CSS object cache is also leveraged for speed.

The config shape (by button/preset) was chosen as it is most likely users will simply want to change the buttons or permit additional colors rather than change both settings for `color` or `background-color`.

### DIFF
--- a/test/system/paste_test.rb
+++ b/test/system/paste_test.rb
@@ -24,7 +24,7 @@ class PasteTest < ApplicationSystemTestCase
     visit edit_post_path(posts(:hello_world))
     find_editor.paste "https://37signals.com"
 
-    assert_editor_html %(<p><a href=\"https://37signals.com\">https://37signals.com</a>Hello everyone</p>)
+    assert_editor_html %(<p>Hello everyone<a href=\"https://37signals.com\">https://37signals.com</a></p>)
   end
 
   test "create links when pasting URLs keeps formatting" do


test/system/paste_with_styles_test.rb
```diff
@@ -0,0 +1,76 @@
...
+
+    visit edit_post_path(post)
+
+    assert_equal_html existing_styled_html, find_editor.value
+  end
+
+  private
+    def paste_with_style(style)
+      find_editor.paste "some styled text", html: %(some <mark style="#{style}">styled text</mark>)
+      sleep 0.1
```
jorgemanrubia: I would love to avoid sleeps if possible. Could we assert_css that the expected markup is in the editor or something like that?

samuelpecher:
> suite is failing

I suspect this is a difference with GH action's environment. I'll see if I can make it work before mocking it out.

test/system/paste_with_styles_test.rb
```diff
@@ -0,0 +1,83 @@
...
+      assert_equal_html %(<p>some <mark style="#{style};">styled text</mark></p>), find_editor.value
+    end
+
+    def highlight_1_rgb
+      dark_mode? ? "rgb(240, 200, 22)" : "rgb(136, 118, 38)"
+    end
+
+    def dark_mode?
+      page.evaluate_script "window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches"
+    end
```
samuelpecher, line 82: @jorgemanrubia This was fun: the headless Chrome instance was picking-up the machine's dark mode setting. Local Omarchy is in dark mode, GH Actions is in light mode. Rather than depart from the pre-configured `:selenium_chrome_headless` I judo'd and made the test value dynamic.

cc @zoltanhosszu

jorgemanrubia: OH LOL debugging that must have been fun!

PR #551, opened 2026-01-09
Embed table wrapper

zoltanhosszu:
Tables are weird creatures of HTML. No matter how you force them to have a max-width of 100%, if their content is larger, they flow out of their container. So adding a wrapper element around them with `overflow-x: auto` allows us to limit this to a scrollable container.

We can easily wrap them in a container inside the editor with `setScrollableTablesActive(this.editor, true)`, but that doesn't change the output.

Previously we didn't include a wrapper around the tables to leave it for the rendering to handle it, but decided with @samuelpecher that we probably should, to ease the use of pure Lexxy outputs.

With this PR tables are now getting wrapped within a `<figure class="lexxy-content__table-wrapper">` element.


src/nodes/wrapped_table_node.js
```diff
@@ -0,0 +1,31 @@
...
+    const originalAfter = output.after
+
+    return {
+      ...output,
+      after: (tableElement) => {
+        if (originalAfter) {
+          tableElement = originalAfter(tableElement)
+        }
+
+        if (tableElement) {
```
samuelpecher: This could be collapsed into 

```suggestion
        if (tableElement = orignalAfter?.(tableElement) {
```

src/nodes/wrapped_table_node.js
```diff
@@ -0,0 +1,31 @@
...
+          const clonedTable = tableElement.cloneNode(true)
+          const figure = createElement("figure", { className: "lexxy-content__table-wrapper" }, clonedTable.outerHTML)
+          return figure
```
samuelpecher: I'd consider a tad more description
```suggestion
          const wrappedTable = createElement("figure", { className: "lexxy-content__table-wrapper" }, clonedTable.outerHTML)
          return wrappedTable
```

src/nodes/wrapped_table_node.js
samuelpecher: How about a more explicit name to not confuse with this func's ouput and then collapse the conditions using conditional chaining:

```suggestion
    const superExport = super.exportDOM(editor)

    return {
      ...superExport,
      after: (tableElement) => {
        if (tableElement = superExport.after?.(tableElement) {
```

test/system/table_test.rb
```diff
@@ -122,4 +122,11 @@ class TableTest < ApplicationSystemTestCase
     assert_no_selector "lexxy-editor"
     assert_table_structure(3, 3)
   end
+
+  test "table is wrapped in figure.table-wrapper" do
+    find_editor.toggle_command("insertTable")
+
+    html = find_editor.value
+    assert_match(/<figure[^>]*class=["'][^"']*lexxy-content__table-wrapper[^"']*["'][^>]*>\s*<table>.*?<\/table>\s*<\/figure>/m, html, "Table should be wrapped in figure.lexxy-content__table-wrapper")
```
samuelpecher: <s>This can use Rails' selectors 
```ruby
    assert_select "figure.lexxy-content__table-wrapper > table, "Table should..."
```
</s>

**edit for posterity**: No you can't since it's not a Capybara::Node. _Yet_.

PR #552, opened 2026-01-09
Remove active sleeps in tests

jorgemanrubia:
In general, active delays in tests are a bad practice. They slow the suite down, and they are unreliable.

Capybara comes with automatic waiting for the built-in helpers, we should always try to use those when possible. E.g: if something changes in the screen, assert with a selector the expected change, so that Capybara waits until it happens. 

I added a custom `wait_until` helper method for custom conditions that we can't cover with capybara.

I left a couple of sleeps after pressing keys, since I can't think of a way to avoid those.

### DIFF
--- a/test/system/editor_to_string_test.rb
+++ b/test/system/editor_to_string_test.rb
@@ -26,13 +26,11 @@ class EditorValueMethodsTest < ApplicationSystemTestCase
     attach_file file_fixture("example.png") do
       click_on "Upload file"
     end
-    sleep 0.1
 
     assert_editor_plain_text "[example.png]\n\n"
 
     find("figcaption textarea").click.send_keys("Example Image")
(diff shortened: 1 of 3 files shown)


jorgemanrubia:
cc @samuelpecher @zoltanhosszu

zoltanhosszu:
Ah this is awesome!

samuelpecher:
Not sure if it's this or #479. If I rebuild the assets on `main` they are different and I get this consistent failure:

```
Failure:
ActionTextLoadTest#test_Clearing_value_of_editor_doesn't_add_trailing_markup [test/system/form_test.rb:70]:
Expected: ""
  Actual: "<p><br></p>"


bin/rails test test/system/form_test.rb:65
```

PR #553, opened 2026-01-10
Attachment configs

samuelpecher:
Handle and test configurable authenticated uploads

- Adds`authenticatedUploads` to global config to set `withCredentials: true` on ActiveStorage upload XHR for authenticated storage controllers
- Adds optional authenticated upload subdomain to dummy app for testing proper CORS cookie handling in the XHR
- Some attachment clean-up
  - Lexical style expects a conversion function to have an element argument
  - Move global config evaluation to static class-level getters

I've skipped a test which is [failing on `main`](https://github.com/basecamp/lexxy/pull/552#issuecomment-3731150800)


src/nodes/action_text_attachment_node.js
```diff
@@ -33,46 +34,51 @@ export class ActionTextAttachmentNode extends DecoratorNode {
...
+            }
+          }, priority: 1
         }
       }
     }
   }
 
-  constructor({ sgid, src, previewable, altText, caption, contentType, fileName, fileSize, width, height }, key) {
+  static get TAG_NAME() {
+    return Lexxy.global.get("attachmentTagName")
```
jorgemanrubia, line 74: nice!

src/nodes/action_text_attachment_upload_node.js
```diff
@@ -23,8 +24,13 @@ export class ActionTextAttachmentUploadNode extends ActionTextAttachmentNode {
     return null
   }
 
-  constructor({ file, uploadUrl, blobUrlTemplate, editor, progress }, key) {
-    super({ contentType: file.type }, key)
+  static get AUTHENTICATED_UPLOADS() {
+    return Lexxy.global.get("authenticatedUploads")
```
jorgemanrubia: I like the `TAG_NAME` one, as you end up doing `this.TAG_NAME` which reads nicely, but do you think this one is an improvement? At the end, you are replacing `Lexxy.global.get("authenticatedUploads")` with ` ActionTextAttachmentUploadNode.AUTHENTICATED_UPLOADS`. Feels like unnecessary ceremony to me.

samuelpecher: I like keeping it simple over the symmetry of the static globals. 👍

test/dummy/app/controllers/authenticated_direct_uploads_controller.rb
```diff
@@ -0,0 +1,12 @@
+class AuthenticatedDirectUploadsController < ActiveStorage::DirectUploadsController
+  skip_forgery_protection
```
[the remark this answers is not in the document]
samuelpecher: This a demo app, so no concern.

PR #554, opened 2026-01-10
Refactor Tests

samuelpecher:
- Introduce `assert_editor_html` which compares html to editor.value
- Allow passing a block which allows working with Capybara selectors
- Add `within_table` helper which takes a block in the same way to clarify table tests

Test time is stable before/after at ~45s

```ruby
# Comparing editor value
test "lexxy value" do
  assert_editor_html "<p>Hello there</p>"
end

# Passing a block
test "table is wrapped in figure.table-wrapper" do
  find_editor.toggle_command("insertTable")

  assert_editor_html do
    assert_selector "figure.lexxy-content__table-wrapper" do |figure|
      figure.has_table?
    end
  end
end

# Using `within_table`
test "writing in table fields" do
  find_editor.toggle_command("insertTable")

  find_editor.send "Test Cell"

  within_table do
    assert_selector "td", text: "Test Cell"
  end
end
```

### DIFF
--- a/test/system/table_test.rb
+++ b/test/system/table_test.rb
@@ -16,9 +16,9 @@ class TableTest < ApplicationSystemTestCase
 
     find_editor.send "Test Cell"
 
-    html = find_editor.value
-    assert_match(/Test Cell/, html)
-    assert_match(/<table>.*?Test Cell.*?<\/table>/m, html)
+    within_table do
+      assert_selector "td", text: "Test Cell"
+    end
   end
 
   test "adding a new row" do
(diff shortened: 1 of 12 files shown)


test/system/table_test.rb
```diff
@@ -126,9 +131,18 @@ class TableTest < ApplicationSystemTestCase
...
-    assert_match(/<figure class="lexxy-content__table-wrapper">.*?<table>.*?<\/table>.*?<\/figure>/m, html, "Table with content should be nested inside the figure wrapper")
+    assert_editor_html do
+      assert_selector "figure.lexxy-content__table-wrapper" do |figure|
+        figure.has_table?
+      end
+    end
   end
+
+  private
+
```
jorgemanrubia: ✂ newline

test/test_helpers/editor_helper.rb
```diff
@@ -12,6 +12,14 @@ def assert_editor_plain_text(value)
     wait_until { find_editor.plain_text_value == value }
   end
 
+  def assert_editor_html(value = nil, &block)
+    if block
+      Capybara.string(find_editor.value).instance_exec(&block)
+    else
+      wait_until { assert_equal_html find_editor.value, value }
```
jorgemanrubia: I don't think this is going to wait as you expect: if the assert fails it will raise. Coincidentally I added a similar helper here when looking into that other failing test you found: 

https://github.com/basecamp/lexxy/blob/d80dd8623d8292de66ee62b352b25da2f2af8914/test/test_helpers/editor_helper.rb#L14-L19

The block idea is great but I would incorporate the same approach I used here. If the assertion timeouts, perform a real assertion so that you get a meaningful comparison error.

jorgemanrubia: BTW feel free to keep your  `assert_editor_html` naming and remove my `assert_editor_html_value` 👍. We should use the same method everywhere.

samuelpecher: 🙏

PR #555, opened 2026-01-10
Multiple prompt attachables and configurable content-type

samuelpecher:
Support multiple attachments from a single prompt item. This is useful for scenarios like mentioning groups of users.  Now, each `<template type="editor">` elements inside the prompt item is inserted. Attributes `sgid` and `content-type` can be specified at this level.

```erb
<lexxy-prompt-item search="<%= group.name %>" >
  <template type="menu"><%= group.name %></template>

  <template type="editor" sgid="<% first.attachable_sgid %>">
    <%# first content %>
  </template>
  <template type="editor" sgid="<% second.attachable_sgid %>">
    <%# second content %>
  </template>
  <%# etc... %>
</lexxy-prompt-item>
```

In addition, the global default `content-type` namespace can be changed from the default `actiontext` using a config.

```js
Lexxy.configure({ global: { attachmentContentTypeNamespace: "myapp" } })

// ... default content-type is now `application/vnd.myapp.${type}`
```

------
Requires #553 Attachment Configs


src/nodes/custom_action_text_attachment_node.js
```diff
@@ -53,12 +53,16 @@ export class CustomActionTextAttachmentNode extends DecoratorNode {
     return Lexxy.global.get("attachmentTagName")
   }
 
+  static get CONTENT_NAMESPACE() {
+    return Lexxy.global.get("attachmentContentTypeNamespace")
```
jorgemanrubia: I feel it would be fine to access the global property directly, but tbh this is fine too if you like it more 👍. I just don't love ceremony/boilerplate, so I don't think the codebase should send the signal that whenever you need to access a global property you need to set a constant for it.

PR #557, opened 2026-01-10
Fix failing test

jorgemanrubia:


jorgemanrubia:
cc @samuelpecher I think this should make the test pass more reliably. It was passing for me locally. Also, the assets were stale in main, so rebuilding them here.

UPDATE: it is still failing, looking

jorgemanrubia:
Oh this behavior changed in https://github.com/basecamp/lexxy/commit/5ee903b7c3d650faf932236b14b6e166e8b0a9a7

Not sure why that PR didn't fail though.

samuelpecher:
Thanks @jorgemanrubia ! I think the PR failed as the assets will have been stale somewhere.

PR #563, opened 2026-01-13
Fix Loading HTML with Custom Tags

### LINKED ISSUE #564: Loading content with custom attachment tags issue
I think there's an issue with loading existing that that uses a custom attachment tag. The attachment gets purged when setting the value of the editor. I _think_ I was able to reproduce it here: https://github.com/basecamp/lexxy/pull/563

Anyone else running into this?

tonysm:
### Fixed

- There was a race condition between registering the element and overriding global configs that prevented things from overriding the attachment tag name from working correctly. This PR fixes it by deferring the initialization of the elements so it runs after everything runs on the main call stack, which allows the configuration to take place before the elements init. 

---

closes #564


tonysm:
I left a note on the issue with step to reproduce it manually too (besides the automated test included here): https://github.com/basecamp/lexxy/issues/564#issuecomment-3779714619

src/elements/index.js
```diff
@@ -0,0 +1,26 @@
...
+    "lexxy-prompt": Prompt,
+    "lexxy-code-language-picker": CodeLanguagePicker,
+    "lexxy-table-tools": TableTools,
+  }
+
+  Object.entries(elements).forEach(([ name, element ]) => {
+    if (!customElements.get(name)) {
+      customElements.define(name, element)
+    }
+  })
```
samuelpecher, line 23: I think it's best not to guard re-registration, which in most cases would be an error?

```suggestion
  Object.entries(elements).forEach(([ name, element ]) => {
    customElements.define(name, element)
  })
```

src/elements/index.js
```diff
@@ -0,0 +1,26 @@
+import Toolbar from "./toolbar"
+
+import Editor from "./editor"
+import DropdownLink from "./dropdown/link"
+import DropdownHighlight from "./dropdown/highlight"
+import Prompt from "./prompt"
+import CodeLanguagePicker from "./code_language_picker"
+import TableTools from "./table/table_tools"
+
+export function start() {
```
samuelpecher: I'd lean to a more descriptive name like `defineElements`.

src/index.js
```diff
@@ -21,3 +13,5 @@ export { default as Extension } from "./extensions/lexxy_extension"
 
 // legacy export for <=v0.7
 export { highlightCode as highlightAll } from "./helpers/code_highlighting_helper"
+
+setTimeout(start, 0)
```
samuelpecher: This would be a great place to document the reasoning for the timeout!

samuelpecher:
While I back this approach over an explicit `Lexxy.init()`, it's worth being aware we're not totally removing the race condition. There's an implied requirement to configure Lexxy synchronously after import. 

@tonysm would you be up to making this requirement explicit in the docs?

src/index.js
tonysm: I've added a comment on why, let me know if that's clear enough

tonysm:
@samuelpecher I've handled the suggestions and added a notice to the documentation on the importance of calling configure right after importing the lexxy module and why. Let me know if there's anything else you'd want to see changed.

src/elements/table/table_controller.js
```diff
@@ -367,3 +367,4 @@ export class TableController {
     return true
   }
 }
+
```
samuelpecher: :scissors: 
```suggestion
```

PR #581, opened 2026-01-14
Add unique id attributes to headings

lylo:
This PR generates slug-based IDs from heading text and adds it as an `id` attribute on the heading element. Duplicate headings get numbered suffixes (-1, -2, etc).


brunoprietog:
I guess this is so you can have anchor links, right? Even so, this has big potential to generate duplicate IDs between content from different instances, or even collide with already defined IDs, given that we have no control over user input.

lylo:
> I guess this is so you can have anchor links, right

Correct. 

> this has big potential to generate duplicate IDs between content from different instances, or even collide with already defined IDs, given that we have no control over user input.

Fair point about multiple instances, I hadn't considered that. Does the benefit of heading ids outweigh this (arguably v rare) risk? I appreciate it might be seen as an opinionated convention too far.

If there's a better way to accomplish getting ids so I can link to headings, I'd be interested in ideas!

It could use the `name` attribute rather than `id` which would (I think) still allow anchor links to work.

brunoprietog:
The name attribute is not global, which means it is not valid for headings (see https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes).

I don't think this is necessarily an uncommon case. Consider an example such as templates for anything. Then imagine that you list many documents that follow those templates, on the same page, all inline. They would all have the same heading IDs.

How about transforming that on the server side for that specific use case?

lylo:
Yeah I think I'll do that. Thanks for the feedback 🙏

PR #587, opened 2026-01-14
Lexxy extensions

samuelpecher:
I wanted to get feedback on adding a hook for extending Lexxy and the underlying Lexical instance.

> [!NOTE]
>  The aim is to allow downstream users to easily hook into Lexxy to add Lexical behavior and provide toolbar markup. I'm not proposing heading towards a fully built-out extension system like Lexical with dependencies etc.

A Lexxy extension is a thin class provided to `Lexxy.configure({ global: { extensions: [ ] } })`. The class itself is a means to:
- provide the `editorElement` to the extension, mainly to get the config `this.editorElement.config.get(...)`
- query whether the extension should be enabled via an overridable `get enabled()`, so it can respond to (for example) the editor configuration 
``` js
get enabled() { return this.editorElement.supportsRichText } 
```
- load toolbar markup in it's `intializeToolbar` which is called if Lexxy has a toolbar, passing the toolbar element

I've included [proposed docs](https://github.com/basecamp/lexxy/blob/e993cb76263b4312c81343fe6e352228339f4266/docs/extensions.md) for more details.

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -18,6 +18,14 @@ Lexxy Extensions are a wrapper around [Lexical Extensions](https://lexical.dev/d
 
 An instance of the extension will be initialized per editor: `new MyLexxyExtension(lexxyElement)`. There is no need to supply a constructor as the base class provides `this.editorElement` to access Lexxy. If you specify a custom constructor you must pass the `lexxyElement` to `super`. If the return of `lexicalExtension` is truthy, it will be loaded at the time of Lexical editor creation. You can optionally provide an `enabled` getter which will be checked to see if your extension should be loaded by Lexxy.
 
+Should you need to extend any of Lexxy's nodes, they are exported at the top-level module:
+
+```javascript
+import { ActionTextAttachmentNode } from "@37signals/lexxy"
+```
+
+## Example
+
 ```js
 import { defineExtension } from "lexical"
 import * as Lexxy from "@37signals/lexxy"
--- a/docs/css-setup.md
+++ b/docs/css-setup.md
@@ -29,17 +29,17 @@ For applying the same styles to rendered Action Text content, you need to overri
 </div>
 ```
 
-To apply syntax highlighting to rendered Action Text content, you need to call the `highlightAll` function from Lexxy. For example, create a Stimulus controller in `app/javascript/controllers/syntax_highlight_controller.js`:
+To apply syntax highlighting to rendered Action Text content, you need to call the `highlightCode` function from Lexxy. For example, create a Stimulus controller in `app/javascript/controllers/syntax_highlight_controller.js`:
 
 ```javascript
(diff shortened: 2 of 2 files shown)


jorgemanrubia (approved):
Love this @samuelpecher 

Another interesting extension point could be letting you extend the nodes we configure in the editor. But we can support that down the road. This is a great start for the feature 👏

docs/extensions.md
```diff
@@ -0,0 +1,68 @@
+---
+title: Extensions
+layout: default
+parent: Configuration
+nav_order: 3
+---
+
+# Extending Lexxy
+
+While Lexxy can be [configured] to provide the functionality needed, for advanced use cases you can define Lexxy extension. Please contribute your feature if others would find it useful.
```
jorgemanrubia: This `[configured]` link is missing the URL, right? You need to use `(configfure.html)` I think.

samuelpecher:
> Another interesting extension point could be letting you extend the nodes we configure in the editor. But we can support that down the road. This is a great start for the feature 👏

I'm working toward exposing `import {...} from "@37signals/lexxy/nodes"` and `import {...} from "@37signals/lexxy/helpers"`

Then [providing additional nodes](https://github.com/facebook/lexical/blob/21458a9b80054e870eb509cfac3735c46b370586/packages/lexical/src/extension-core/types.ts#L374C2-L374C73) and node replacements happens in the hooked `lexicalExtension`.

zoltanhosszu:
This is a great addition @samuelpecher! 👏👏👏

Just one idea for the future: as discussed, maybe we could introduce a dedicated default placement into the toolbar, where extensions control are placed by default. This might introduce some unexpected behaviors with the overflow menu though, so I don't have strong feelings about it.

samuelpecher:
> Just one idea for the future: as discussed, maybe we could introduce a dedicated default placement into the toolbar, where extensions control are placed by default. This might introduce some unexpected behaviors with the overflow menu though, so I don't have strong feelings about it.

Let's see if it crops up. My take was anyone wanting to extend Lexical is probably up to injecting some HTML.

src/helpers/code_highlighting_helper.js
```diff
@@ -1,7 +1,7 @@
 import { createElement } from "./html_helper"
 import Prism from "prismjs"
 
-export function highlightAll() {
+export function highlightCode() {
```
samuelpecher, line 4: cc @packagethief 

Note: we're going to keep `highlightAll` around to not break anything immediately

samuelpecher:
> I'm working toward exposing `import {...} from "@37signals/lexxy/nodes"` and `import {...} from "@37signals/lexxy/helpers"`

To close the loop on this:

I've exported the nodes from the top level module. The complication and refactoring didn't make sense to bundle "lexxy/nodes" separately for five nodes and as the entirely Lexical package is a dependency.

PR #596, opened 2026-01-15
Disable prefix partial path when rendering attachments

jorgemanrubia:
So that when used in nested controllers it does not try to search missing attachable templates
in namespaced controllers (it will raise!)

Temporary, as we will soon get rid of this, with the adaptarized action text approach.

### DIFF
--- a/lib/lexxy/rich_text_area_tag.rb
+++ b/lib/lexxy/rich_text_area_tag.rb
@@ -25,6 +25,7 @@ module Lexxy
       def render_custom_attachments_in(value)
         if value.respond_to?(:body)
           if html = value.body_before_type_cast.presence
+            self.prefix_partial_path_with_controller_namespace = false if respond_to?(:prefix_partial_path_with_controller_namespace=)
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@37signals/lexxy",
-  "version": "0.1.27-beta",
+  "version": "0.7.0-beta",
```
jorgemanrubia, line 3: It seems I didn't commit this one on the last release, will do now.

lib/lexxy/rich_text_area_tag.rb
```diff
@@ -25,6 +25,7 @@ def lexxy_rich_textarea_tag(name, value = nil, options = {}, &block)
       def render_custom_attachments_in(value)
         if value.respond_to?(:body)
           if html = value.body_before_type_cast.presence
+            self.prefix_partial_path_with_controller_namespace = false if respond_to?(:prefix_partial_path_with_controller_namespace=)
```
jorgemanrubia, line 28: there are some helper tests where thers is no controller in place. That is why we use a `respond_to?` condition here.

PR #599, opened 2026-01-16
Tables UX tweaks

zoltanhosszu:
This pull request significantly improves table editing in the editor, both in terms of usability and internal structure.

before, after comparison

#### Dedicated dropdowns for rows and columns
On the UI side, the table controls have been reworked to be clearer and more space-efficient. The “…” menu has been split into dedicated row and column dropdowns, moving these controls closer to their respective +/− buttons. This makes table actions easier to discover and use, especially in constrained layouts. The table tools themselves have been visually refined, with improved button styling, hover and focus states, and more polished dropdown menus.

#### Visual feedback in dropdown actions
To make interactions feel more intuitive, table controls now provide playful visual feedback that previews their effect while hovering (for example, showing where rows or columns will be inserted or removed). This helps users understand what an action will do before committing to it.

#### Table resizing
Table resizing behavior has also been improved as much as current CSS limitations allow. Since browser-native table resizing can’t easily be tuned without a custom JavaScript solution, minimum and maximum widths were added to table cells. This reduces overly eager resizing and results in more predictable layouts.

#### Table selection
Instead of the whole table getting an outline, only the selected cell is highlighted now.
table selection visual change

#### Delete table
Delete table icon has a bit more visual feedback now to reinforce the action.
delete button confirmation

## Code changes
Under the hood, a large refactor was necessary to support these changes. The `TableHandler` class has been split into clearer responsibilities. A new `TableController` class now encapsulates table behavior such as selection, header toggling, row and column insertion/deletion, cell navigation, and keyboard handling, while visual floating controls are handled separately. This makes the codebase much easier to reason about, maintain, and extend going forward.

As part of this work, the editor was updated to use `lexxy-table-tools` instead of `lexxy-table-handler`. Several table-related styles were also improved, including selection and focus states, header backgrounds, and visual indicators for insert, delete, and toggle actions.

A few unrelated Lexxy bugs were fixed along the way as well.


src/elements/table/table_tools.js
```diff
@@ -10,13 +10,16 @@ import { TableAction, TableChildType, TableController, TableDirection } from "./
 import { handleRollingTabIndex } from "../../helpers/accessibility_helper"
 import { createElement } from "../../helpers/html_helper"
 
+const HIGHLIGHT_CLASS = "lexxy-content__table-cell--highlight"
+const FOCUS_CLASS = "lexxy-content__table-cell--focus"
+
```
samuelpecher: Could these be moved to the lexical theme?

src/elements/table/table_controller.js
```diff
@@ -0,0 +1,337 @@
...
+export const TableChildType = Object.freeze({
+  ROW: "row",
+  COLUMN: "column",
+  TABLE: "table"
+})
+
+export const TableDirection = Object.freeze({
+  BEFORE: "before",
+  AFTER: "after"
+})
```
samuelpecher: This feels like a lot of ceremony rather than in-lining the constants, unless I've missed somewhere they are being enumerated?

Naturally, readability is key, but I don't think `action == "insert"` or `childType == "column"` are hard to parse. Same for L339: `["insert" + "row" + "before"]` vs `[TableAction.INSERT + TableChildType.ROW + TableDirection.BEFORE]:`, or better `"insert_row_before":`. Happy to be wrong though...

src/editor/command_dispatcher.js
```diff
@@ -162,9 +162,7 @@ export class CommandDispatcher {
   }
 
   dispatchInsertHorizontalDivider() {
-    this.editor.update(() => {
-      this.contents.insertAtCursorEnsuringLineBelow(new HorizontalDividerNode())
-    })
+    this.contents.insertAtCursorEnsuringLineBelow(new HorizontalDividerNode())
```
samuelpecher, line 151: I think we should move towards Lexical-esque `$createHorizontalDividerNode()`

src/editor/command_dispatcher.js
```diff
@@ -251,14 +249,14 @@ export class CommandDispatcher {
...
 
-      const anchorNode = selection.anchor.getNode()
-      const tableNode = $findTableNode(anchorNode)
-      tableNode.remove()
-    })
+    const anchorNode = selection.anchor.getNode()
+    const tableNode = $findTableNode(anchorNode)
+    if (!tableNode) return
+
+    tableNode.remove()
```
samuelpecher: Nice... all these are dispatched as commands so no need for the `editor.update`

src/elements/table/table_controller.js
```diff
@@ -0,0 +1,385 @@
...
+    if (!cells) return false
+
+    return cells.indexOf(this.currentCell) === 0
+  }
+
+  #handleBackspaceKey() {
+    // We can't prevent these externally using regular keydown because Lexical handles it first.
+    this.unregisterBackspaceKeyHandler = this.editor.registerCommand(
+      KEY_BACKSPACE_COMMAND,
+      (event) => {
```
samuelpecher: This feels like a `registerBackspaceHandler` which should register the handler.
Unless another `backspace` command make sense, I'd use `COMMAND_PRIORITY_HIGH`

```javascript
#registerBackspaceHandler() {
  this.unregisterBackspaceKeyHandler = this.editor.registerCommand(
    KEY_BACKSPACE_COMMAND,
    event => this.#bacspaceHandler(event)
  )
}

#backspaceHandler() { /* ... */ }
```

src/elements/table/table_controller.js
```diff
@@ -0,0 +1,385 @@
...
+        const tableCell = $getTableCellNodeFromLexicalNode(node)
+        if (!tableCell) return
+
+        tableCell.selectEnd()
+      })
+    })
+  }
+
+  #selectCellAtIndex(rowIndex, columnIndex) {
+    requestAnimationFrame(() => {
```
samuelpecher: Might be nice to have a comment explaining why we need `requestAnimationFrame`. The `nextFrame` [helper](https://github.com/basecamp/lexxy/blob/main/src/helpers/timing_helpers.js#L20-L20) can help reduce indentation

src/elements/table/table_controller.js
```diff
@@ -0,0 +1,385 @@
...
+  get currentTableNode() {
+    if (!this.currentTableNodeKey) return null
+
+    const cell = this.currentCell
+    if (!cell) return null
+
+    return this.editor.getEditorState().read(() => {
+      return $findTableNode(cell)
+    })
+  }
```
samuelpecher, line 59: I'm not sure it's clear why store the `TableNode` key and then get it from the cell.

src/elements/table/table_controller.js
```diff
@@ -0,0 +1,385 @@
...
+
+    const rowLength = this.currentRowCells?.length ?? 0
+    this.#selectCellAtIndex(this.currentRowIndex - 1, rowLength - 1)
+  }
+
+  #deleteLastRowAndSelectNext() {
+    this.executeTableCommand({ action: TableAction.DELETE, childType: TableChildType.ROW })
+
+    this.editor.update(() => {
+      const next = this.currentTableNode?.getNextSibling()
```
samuelpecher: This pattern is forcing an editor DOM render. Since `currentTableNode` reads state, it will flush the update from the previous command. Is it possible to move the `currentTableNode` call above the command?

```
executeTableCommand // editor.update

this.editor.update(() => {
  this.currentTableNode // editor.readState // ** Flushes Changes **

  // ... more update
}
```

```suggestion
    const tableNode = this.currentTableNode
    this.executeTableCommand({ action: "delete", childType: "row" })

    this.editor.update(() => {
      const next = tableNode?.getNextSibling()
```

src/elements/table/table_controller.js
```diff
@@ -0,0 +1,385 @@
...
+          return true
+        }
+
+        return false
+      },
+      COMMAND_PRIORITY_NORMAL
+    )
+  }
+
+  #handleEnterKey() {
```
samuelpecher: Same as above 🙂

src/elements/table/table_tools.js
```diff
@@ -0,0 +1,374 @@
...
+      cell.removeAttribute("data-direction")
+    })
+
+    this.#setTableCellFocus()
+  }
+
+  #icon(command) {
+    const { action, childType, direction } = command
+    return ICONS[action + childType + (action == TableAction.INSERT ? direction : "")]
+  }
```
samuelpecher, line 315: consider extracting the `ICON` SVG const to another file, and simplifying the definitions in line with my comment above

```suggestion
  #icon(command) {
    const { action, childType } = command
    const direction = (action == "insert" ? command.direction : "")
    const iconId = [action, childType, direction].join()
    return ICONS[iconId]
  }
```

src/elements/table/table_tools.js
zoltanhosszu: Not sure... These are custom classes I use to highlight the table interactions, and to draw a border around the currently selected cell. Neither of these (surprisingly) are supported by themes as far as I know.

samuelpecher: Ah it's "just" `import theme from "../config/theme"`, not need to reach into the editor config

zoltanhosszu: Oh nice! Will do that :)

src/elements/table/table_controller.js
zoltanhosszu: Good catch!

src/editor/command_dispatcher.js
zoltanhosszu: I agree, but I don't think it belongs to this PR :)

src/elements/table/table_controller.js
```diff
@@ -0,0 +1,371 @@
...
+      event.preventDefault()
+      this.#selectPreviousCell()
+      return true
+    }
+
+    return false
+  }
+
+  #handleEnterKey(event) {
+    if (event.shiftKey ||!this.currentTableNode) return false
```
samuelpecher: Do we want to also not handle `ctrlKey` for example, Basecamp uses that for submit

```suggestion
    if (event.shiftKey || !this.currentTableNode) return false
```

src/elements/table/table_controller.js
```diff
@@ -0,0 +1,373 @@
...
+  #setHeaderStyle(cell, newStyle, headerState) {
+    const tableCellNode = $getTableCellNodeFromLexicalNode(cell)
+
+    if (tableCellNode) {
+      tableCellNode.setHeaderStyles(newStyle, headerState)
+    }
+  }
+
+  async #selectCellAtIndex(rowIndex, columnIndex) {
+    // We wait for next frame, otherwise table operations have might've not completed yet.
```
packagethief: Typo:

```suggestion
    // We wait for next frame, otherwise table operations might not have completed yet.
```

src/elements/table/table_controller.js
```diff
@@ -0,0 +1,373 @@
...
+    if (!rows) return
+
+    const row = rows[rowIndex]
+    if (!row) return
+
+    this.editor.update(() => {
+      const cell = $getTableCellNodeFromLexicalNode(row.getChildAtIndex(columnIndex))
+      if (!cell) return
+
+      cell.selectEnd()
```
packagethief: Minor style nit: you can use the optional chaining operator here.


```suggestion
      cell?.selectEnd()
```

src/elements/table/table_tools.js
```diff
@@ -0,0 +1,322 @@
...
+
+  disconnectedCallback() {
+    this.#unregisterKeyboardShortcuts()
+
+    this.unregisterUpdateListener?.()
+    this.unregisterUpdateListener = null
+
+    this.removeEventListener("keydown", this.#handleToolsKeydown)
+
+    this.tableController?.destroy?.()
```
packagethief: We know the `TableController` has a `destroy()` method, so the second `?` isn't necessary:


```suggestion
    this.tableController?.destroy()
```

src/elements/table/table_tools.js
```diff
@@ -0,0 +1,322 @@
...
+      cell.select()
+      this.#editor.focus()
+    })
+
+    this.#update()
+  }
+
+  #handleCommandButtonHover() {
+    requestAnimationFrame(() => {
+      this.#clearCellStyles()
```
packagethief: We can await the `nextFrame` here instead of rAF to remove a level of nesting:

```js
async #handleCommandButtonHover() {
  await nextFrame()

  // ...
}
```

src/editor/command_dispatcher.js
packagethief: Consider using optional chaining to remove the need for a local variable:

```suggestion
    $findTableNode(anchorNode)?.remove()
```

src/elements/table/table_controller.js
```diff
@@ -0,0 +1,373 @@
...
+    this.editor.update(() => {
+      const selection = $getSelection()
+      if (!selection) return
+
+      const node = selection.getNodes()[0]
+
+      const cellNode = $findCellNode(node)
+      if (!cellNode) return
+
+      cellNode.selectEnd()
```
packagethief: Consider using optional chaining here too, to eliminate the need for a local variable. I think the result is just as clear.


```suggestion
      $findCellNode(node)?.selectEnd()
```

src/helpers/string_helper.js
```diff
@@ -24,3 +24,7 @@ export function normalizeFilteredText(string) {
 export function filterMatches(text, potentialMatch) {
   return normalizeFilteredText(text).includes(normalizeFilteredText(potentialMatch))
 }
+
+export function capitalizeFirstLetter(string) {
```
samuelpecher: considering matching Rails: `upcase_first` (so `upcaseFirst` in js)

src/elements/table/table_tools.js
```diff
@@ -0,0 +1,322 @@
...
+      } else {
+        this.#hide()
+      }
+    })
+  }
+
+  #executeTableCommand(command) {
+    if (command.action === "delete" && command.childType === "table") {
+      this.tableController.deleteTable()
+    } else {
```
samuelpecher: I wonder if branch feels better in the controller since all the other commands are passed-through.

zoltanhosszu:
OMG this PR keeps growing... :D Thanks @packagethief and @samuelpecher for the reviews!

I've tweaked all comments, and added a few more fixes. Mind taking a last look on those, or should we merge this?

src/editor/command_dispatcher.js
```diff
@@ -203,6 +206,10 @@ export class CommandDispatcher {
     setTimeout(() => input.remove(), 1000)
   }
 
+  dispatchInsertTable() {
+    this.editor.dispatchCommand(INSERT_TABLE_COMMAND, { "rows": 3, "columns": 3, "includeHeaders": true })
+  }
+
```
samuelpecher: cc @zoltanhosszu I felt commands in the default toolbar should stay in the dispatcher at the moment.

Issue #441, opened 2025-11-12
Lexxy editor inserts <p><br></p> into params even when editor is empty


tabosque:
I'm using Lexxy in a Ruby on Rails project, and I found that the editor sends unwanted HTML even when no content is entered.

#### Environment
- Ruby: 3.4.7  
- Rails: 8.1.1  
- Lexxy: 0.1.20.beta

#### Setup
`Post` model has `body` in ActionText format.
```ruby
has_rich_text :body
```

### Problem
When I submit the form without typing anything in the Lexxy editor, the params already contain `<p><br></p>` instead of an empty string.
Here’s the captured params:
```ruby
#<ActionController::Parameters {
  "authenticity_token" => "TOKEN",
  "post" => {"title" => "", "body" => "<p><br></p>"},
  "lexxy-code-language" => "plain",
  "commit" => "Create Post",
  "controller" => "posts",
  "action" => "create"
} permitted: false>
```

### Expected behavior
If the editor is empty, I expect `params[:post][:body]` to be an empty string (`""`), not `<p><br></p>`.

tabosque:
Related commit
https://github.com/basecamp/lexxy/pull/440/commits/fa7dd3853c2ccf4ba2952fad4797f589f719c815

ronaldlangeveld:
I believe  `<p><br></p>` is default Lexical behaviour, as I recall from discussions a few years ago while working on Ghost's Lexical editor. Essentially, the editor defaults to an empty paragraph with a line break (containing no text content) to ensure an editable contentEditable text area remains in the editor, preventing issues like cursor placement failures, focus loss, unreliable typing, and layout collapse that can occur in completely empty editable regions. This minimal placeholder keeps the editor functional and user-friendly without visible clutter.

https://github.com/TryGhost/Koenig/blob/8b8aba578bad77b29f2107ea1c18c535e07c00d0/packages/kg-lexical-html-renderer/lib/convert-to-html-string.ts#L22-L30 - Seem's a bit hacky but this fixed it for Koenig Editor's purposes.

The ideal fix would be to normalise empty editor content on the server side - in the editor that passes the html to ActionText, ensuring that `<p><br></p>` gets treated as an empty string before parsing, rendering, or persisting to the database. This way, we avoid storing or displaying meaningless HTML in Action Text as @tabosque explained in the issue.

tabosque:
That makes complete sense.
Thanks @ronaldlangeveld!!

PR #637, opened 2026-01-21
Fixed HTML rendered including empty trailing paragraphs

ronaldlangeveld:
ref https://github.com/basecamp/lexxy/issues/441

- Added `normalizeEmptyContent` helper function to convert Lexical's default `<p><br></p>` (used for empty editor content to maintain `contentEditable` functionality) into an empty string, preventing unwanted HTML from being submitted to the server.

- Updated the value `getter` in `src/elements/editor.js` to apply this normalization after sanitizing the generated HTML, ensuring clean data persistence without breaking editor usability.

- This fix addresses the issue where empty editors would submit `<p><br></p>` instead of an empty string, improving data integrity and user experience by avoiding meaningless HTML storage in Action Text fields.

- Follows knowledge from my previous work on [Ghost's Lexical Editor](https://github.com/TryGhost/Koenig) :)


src/elements/editor.js
```diff
@@ -171,7 +172,9 @@ export default class LexicalEditorElement extends HTMLElement {
   get value() {
     if (!this.cachedValue) {
       this.editor?.getEditorState().read(() => {
-        this.cachedValue = sanitize($generateHtmlFromNodes(this.editor, null))
+        this.cachedValue = normalizeEmptyContent(
+          sanitize($generateHtmlFromNodes(this.editor, null))
+        )
```
samuelpecher: I think the unnested route is clearer here

```suggestion
        const html = sanitize($generateHtmlFromNodes(this.editor, null))
        this.cachedValue = normalizeEmptyContent(html)
```

src/helpers/html_normalization_helper.js
samuelpecher, line 1: This be rolled into HTML helpers file, or sanitization. We can keep it terse without the documentation, the intent is nice and clear.

test/javascript/editor/value.test.js
samuelpecher, line 1: I'd lean towards dropping the unit test and relying on one ruby system test.

samuelpecher (commented):
Thanks for the PR! I like the direct approach vs checking the node state.

I've left a few comments to just touch up the code in line with our conventions.

ronaldlangeveld:
Thank you so much for the thorough feedback @samuelpecher . Let me know if there's anything else I can do, happy to help.

PR #670, opened 2026-01-22
Dialog close simplification

zoltanhosszu:
This PR simplifies the code for toolbar dialog close actions.

- Stripped out the close handler from `toolbar_dropdown.js` and attached it to editor selection change, which the toolbar already monitors
- The `lexxy:focus` event fires if any element within `<lexxy-editor>` is focused (e.g. toolbar). Previously it fired only if the writing area got focus.
- The `lexxy:blur` event only fires if the focus moves out of the `<lexxy-editor>` element, not just the writing area.


src/elements/toolbar.js
```diff
@@ -166,10 +166,17 @@ export default class LexicalToolbarElement extends HTMLElement {
     this.editor.registerUpdateListener(() => {
       this.editor.getEditorState().read(() => {
         this.#updateButtonStates()
+        this.#closeDropdowns()
       })
     })
```
samuelpecher: If the point is monitoring just selection change, there's actually a command to hook into it. This would prevent the concern that non-selection non-user initiated editor state changes (such as a download completing) would close the dialogs. 

```javascript
return this.editor.registerCommand(
  SELECTION_CHANGE_COMMAND,
  () => {
    this.#closeDropdowns()
    this.#updateButtonStates()
    return false
  }, COMMAND_PRIORITY_HIGH)
```

If not going down that route, we can clean it up a bit since the handler is passed `editorState`, and then use `tags` to ignore non-selection updates:
```javascript
this.editor.registerUpdateListener(({ editorState, tags }) => {
  editorState.read(() => {
    if (tags.has(SKIP_DOM_SELECTION_TAG)) return 
    this.#updateButtonStates()
    this.#closeDropdowns()
  })
})
```

src/elements/toolbar_dropdown.js
```diff
@@ -26,47 +25,21 @@ export class ToolbarDropdown extends HTMLElement {
   }
 
   close() {
+    this.editor.focus()
     this.container.removeAttribute("open")
```
samuelpecher: How about...?
```suggestion
    this.container.open = false
```

src/elements/toolbar_dropdown.js
```diff
@@ -9,7 +9,6 @@ export class ToolbarDropdown extends HTMLElement {
   }
 
   disconnectedCallback() {
-    this.#removeClickOutsideHandler()
```
samuelpecher, line 12: There's an explicit difference between clicking outside the dropdown, and interacting with the editor here... just confirming that's the desired experience?

If you want to close the drop-downs on `blur` there's [`BLUR_COMMAND`](https://lexical.dev/docs/api/modules/lexical#blur_command), though if that fires after open it can just close it again.

src/elements/toolbar.js
```diff
@@ -166,10 +166,17 @@ export default class LexicalToolbarElement extends HTMLElement {
...
   }
 
+   #closeDropdowns() {
+    this.querySelectorAll("details").forEach((details) => {
+      details.removeAttribute("open")
```
samuelpecher: If you change the other one...
```suggestion
      details.open = false
```

src/elements/toolbar.js
zoltanhosszu: This works great, thanks Sam! Interestingly registering the same `SELECTION_CHANGE_COMMAND` for `table_tools` and `code_language_picker` results in some delayed handling, so I'll keep those as they are.

src/elements/toolbar_dropdown.js
zoltanhosszu: Updated the handling, hope it makes sense like this. Needed to rework `lexxy:focus` and `lexxy:blur` events to fire for all elements within `lexxy-editor`, instead of just the editorElement (ripped out Lexical's `FOCUS_COMMAND` and `BLUR_COMMAND`).

src/elements/editor.js
samuelpecher, line 1: Nice and tight 💫

src/elements/toolbar_dropdown.js
```diff
@@ -29,13 +29,15 @@ export class ToolbarDropdown extends HTMLElement {
...
     if (this.container.open) {
-      this.#handleOpen(event.target)
+      this.#handleOpen()
     }
   }
 
-  #handleOpen() {
+  async #handleOpen() {
+    // We wait for next frame, otherwise editor blur event fires
+    await nextFrame()
```
samuelpecher: This was a race condition with the highlight child class's `toggle` handler. Resolved by [creating an editor promise on the toolbar](https://github.com/basecamp/lexxy/pull/670/commits/97e66a94e54047fc4758ef6d08d64735dbc49678) for dropdowns to await.

PR #676, opened 2026-01-23
Multi-character `prompt:` triggers

zachasme:
Adds support for multiple characters in prompt triggers. We're hoping to use those for BC5 Search.


src/elements/prompt.js
```diff
@@ -81,20 +81,25 @@ export default class LexicalPromptElement extends HTMLElement {
         const { node, offset } = this.#selection.selectedNodeWithOffset()
```
samuelpecher, line 81: While you're here I'd check if we can improve the registered listener, which might quash some bugs since I suspect `this.#editor.read()` can fetch stale state.

Listeners are passed the current state, so the form is:
```javascript
registerUpdateListener(({ editorState })=> {
  editorState.read(() => { /* ... */ })
}
```

Ultimately, I think this would leverage a [node mutation listener](https://lexical.dev/docs/concepts/listeners#registermutationlistener) since we're only addressing text nodes .

test/system/prompts_test.rb
samuelpecher, line 1: I think it's worth testing the expected interaction between `supportsSpaceInSearches` and multi-char.

zachasme: Good point! 👍

PR #690, opened 2026-01-26
Attachment nodes refactor to prevent race condition

samuelpecher:
Refactor of the Attachment nodes to be more 'lexical' and prevent a race condition where sibling nodes refreshed each other, cascading race conditions and multiplying uploads. Updates are now routed via the node, which is more OO and more "lexical", removing some custom event dispatch/handle logic.

The changes have been done commit-by-commit, so the history is easier to follow.

I removed which loaded the image from the remote before node replacement. This may well have been to avoid vertical-content-flashing, but since the entire DOM is replaced when the node becomes another type, this doesn't actually prevent it.

I've kept the OO-feel of keeping the `this.editor` reference, which isn't true "Lexical", but prevents passing the editor reference around when attaching handlers etc. The node has a ref to its editor via `this.editor = $getEditor()` in the constructor.

cc @packagethief


src/nodes/action_text_attachment_upload_node.js
```diff
@@ -100,92 +126,125 @@ export class ActionTextAttachmentUploadNode extends ActionTextAttachmentNode {
...
-      return Promise.resolve()
-    } else {
-      return loadFileIntoImage(this.file, image)
-    }
+  #createProgressBar() {
+    return createElement("progress", { value: this.progress ?? 0, max: 100 })
   }
 
-  async #startUpload(progressBar, figure) {
+  #setDimensionsFrom({ width, height }) {
```
packagethief, line 133: Consider naming `setDimensionsFromImage`. Can still destructure `width` and `height` but including "image" in the name will clarify their origin.

src/nodes/action_text_attachment_upload_node.js
```diff
@@ -100,92 +126,125 @@ export class ActionTextAttachmentUploadNode extends ActionTextAttachmentNode {
...
+    this.editor.update(() => {
+      this.replace(this.#toActionTextAttachmentNodeWith(blob))
+    }, { tag: SILENT_UPDATE_TAGS })
+  }
+
+  #toActionTextAttachmentNodeWith(blob) {
+    const conversion = new AttachmentNodeConversion(this, blob)
+    return conversion.toAttachmentNode()
+  }
+
```
packagethief, line 212: ✂️

PR #691, opened 2026-01-26
Add Image Upload/Attachment Optimization

Matarim:
I brought this idea here: [Idea Discussion](https://github.com/basecamp/lexxy/discussions/687), but the goal of this work is to provide the ability to optimize images that are attached to Lexxy form.

**Can this work be done elsewhere:** 
This work can be done in an initializer in your project that patches the ActiveStorage::Attachment to pull down the attached images loop through and modify the storage object and the blob.  This requires modifying the HTML to point to the new key after conversion and honestly would not be optimal for heavy usage, but is possible.

**Why I think it belongs here:**
Doing the work at the time of the preview front-loads the optimization process and gives an easier way to handle modifying the optimization based on where it is being used by simply changing the configuration.

I'm attaching an example of how to use below, but that is also available in the discussion brought up before.

```

Lexxy.configure({
    global: {
        imageOptimization: {
            enabled: true,
            format: "webp",
            maxWidth: 1200,
            maxHeight: 900,
            quality: 0.8
        },
    }
})
```

Additional consideration and thoughts on restricting optimization on certain file types like svg may be a good plan here, and I'm open to discussion and ideas on how to better handle image file types that shouldn't be converted to a different format/quality.


samuelpecher:
Hey @Matarim, I've been turning this over for a bit and the needs of this feature will be quite application-specific. I think the better approach here is to handle the [`lexxy:file-accept` event](https://basecamp.github.io/lexxy/events.html#lexxyfile-accept) and then do any custom handling client side before uploading the file manually via `editorElement.contents.uploadFiles()`

Matarim:
> Hey @Matarim, I've been turning this over for a bit and the needs of this feature will be quite application-specific. I think the better approach here is to handle the [`lexxy:file-accept` event](https://basecamp.github.io/lexxy/events.html#lexxyfile-accept) and then do any custom handling client side before uploading the file manually via `editorElement.contents.uploadFiles()`

Understood,  I will close this. Thank you!

PR #713, opened 2026-01-30
Preserve ordered list starting number

mdiebolt:
At Poll Everywhere we have a use-case where we have multiple Lexxy editors on the same page representing choices to a multiple choice question. Our customers want to be able to mark choices with 1., 2., 3. and have them styled as a list. 

Prior to this change the editor rendered the lists as authored but on save the list element `start` attribute would get sanitized away, resulting in each choice showing `1.`. 

This PR adds `start` to both the DOM purify and ActionText attribute allowlists.

**Before**
before

**After**
after

### DIFF
--- a/lib/lexxy/engine.rb
+++ b/lib/lexxy/engine.rb
@@ -39,7 +39,7 @@ module Lexxy
         ActionText::ContentHelper.allowed_tags = default_allowed_tags + %w[ video audio source embed table tbody tr th td ]
 
         default_allowed_attributes = Class.new.include(ActionText::ContentHelper).new.sanitizer_allowed_attributes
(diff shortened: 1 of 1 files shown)


mdiebolt:
@samuelpecher is this type of thing something y’all would be open to incorporating into Lexxy?

samuelpecher:
@mdiebolt I think the right direction here would be to make a configuration path to allow more attributes so each application can customize for their own needs rather than upstream all the required attributes into core Lexxy

jorgemanrubia:
Yes I agree with @samuelpecher. It would be great to make this configurable instead of hard-coded

adam-h:
I took a stab at making this configurable in https://github.com/basecamp/lexxy/pull/925

PR #717, opened 2026-02-03
Performance updates

zoltanhosszu:
## Performance and Memory Management Improvements for Lexxy

This PR addresses performance bottlenecks and memory leaks identified in Lexxy, particularly noticeable when editing large documents.

#### DOM Query Caching in Toolbar
- Cache button and menu element references on initialization instead of querying the DOM on every update
- Removes ~15 `querySelector` calls per keystroke

#### Table State Consolidation
- Consolidate scattered table state reads into a single cached `tableState` getter
- Previously, each table operation triggered separate editor state reads; now they share one cached result per update cycle

#### Memory Leak Fixes
- Add `destroy()` method to `CommandDispatcher` with proper cleanup of drag/drop handlers and command registrations
- Store and call unregister functions for all `registerUpdateListener` and `registerCommand` calls
- Fix event listener cleanup in toolbar, code language picker, and table tools

#### Update Listener Optimization
- Skip update listener callbacks when `dirtyElements` and `dirtyLeaves` are both empty (no actual content changes)
- Reduces unnecessary work during selection-only changes


src/elements/toolbar.js
```diff
@@ -327,30 +355,6 @@ export class LexicalToolbarElement extends HTMLElement {
    })
  }
 
-  get #dropdowns() {
-    return this.querySelectorAll("details")
-  }
```
samuelpecher, line 300: I'm a fan of this form for caching, which can happen on-demand and reads clearly. 

```javascript
  get dropdowns() {
    return this.#dropdowns ??= this.querySelectorAll("details")
  }
```

I'd promote the getters to public over having an internal var named `#_dropdowns`

src/elements/toolbar.js
```diff
@@ -285,7 +313,7 @@ export class LexicalToolbarElement extends HTMLElement {
   }
 
   #compactMenu() {
-    const buttons = this.#buttons.reverse()
+    const buttons = [ ...this.#buttons.values() ].reverse()
```
samuelpecher:
```javascript
Array.from(this.#buttons.values())
```

Though this makes me wonder about the API here...
```javascript
getButton(name) {
  return buttonMap.get(name)
}

get buttons() {
  return Array.from(buttonMap.values())
}
```

src/elements/toolbar.js
```diff
@@ -205,16 +204,11 @@ export class LexicalToolbarElement extends HTMLElement {
       this.editor.getEditorState().read(() => {
```
samuelpecher: If this is being changed, it's worth using the passed editor state
```javascript
this.editor.registerUpdateListener(({ editorState }) => {
  editorState.read(/*...*/)
}
```

test/dummy/app/javascript/controllers/lexxy_output_controller.js
samuelpecher, line 1: It's clearer to have the debounce as a generic helper that wraps the function

See https://github.com/basecamp/haystack/blob/master/app/javascript/helpers/debounce_helpers.js#L1-L9

src/elements/editor.js
```diff
@@ -186,12 +186,7 @@ export class LexicalEditorElement extends HTMLElement {
       root.selectEnd()
 
       this.#toggleEmptyStatus()
-
-      // The first time you set the value, when the editor is empty, it seems to leave Lexical
-      // in an inconsistent state until, at least, you focus. You can type but adding attachments
-      // fails because no root node detected. This is a workaround to deal with the issue.
-      requestAnimationFrame(() => this.editor?.update(() => { }))
-    })
+    }, { discrete: true })
```
samuelpecher, line 193: I'm onboard with the idea here, especially removing the frame await, but this can't be a large part of the performance win since it's only on programmatic `set value` ?

Do we have a test that fails without the previous code, and then passes with `{ discrete: true }` to be sure this handles the problem?

I'd also port the comment to explain the use of `discrete`

src/elements/table/table_controller.js
```diff
@@ -40,68 +40,26 @@ export class TableController {
...
-      const cell = $getNodeByKey(this.currentCellKey)
-      return (cell instanceof TableCellNode) ? cell : null
-    })
-  }
+      if (!this.currentCellKey || !this.currentTableNodeKey) return null
 
-  get currentTableNode() {
-    if (!this.currentTableNodeKey) return null
+      const cell = $getNodeByKey(this.currentCellKey)
+      if (cell instanceof TableCellNode === false) return null
```
samuelpecher: This could use `$isTableCellNode`

src/elements/table/table_controller.js
```diff
@@ -40,68 +40,26 @@ export class TableController {
...
-    if (!currentCell) return 0
-
-    return this.editor.getEditorState().read(() => {
-      return $getTableRowIndexFromTableCellNode(currentCell)
-    }) ?? 0
-  }
-
-  get currentColumnCells() {
-    const columnIndex = this.currentColumnIndex
+      if (tableNode instanceof TableNode === false) return null
```
samuelpecher: see above

src/elements/table/table_controller.js
```diff
@@ -40,68 +40,26 @@ export class TableController {
...
 
-    const rows = this.tableRows
-    if (!rows) return null
-
-    return this.editor.getEditorState().read(() => {
-      return rows.map(row => row.getChildAtIndex(columnIndex))
-    }) ?? null
-  }
+      const rows = tableNode?.getChildren() ?? []
```
samuelpecher: Given `tableNode` is guaranteed to not be nullish thanks to the guard above, this could be a straight 
```javascript
const rows = tableNode?.getChildren() ?? []
```

src/elements/table/table_controller.js
```diff
@@ -40,68 +40,26 @@ export class TableController {
...
-    const currentCell = this.currentCell
-    if (!currentCell) return 0
+      const currentRowIndex = $getTableRowIndexFromTableCellNode(cell)
+      const currentColumnIndex = $getTableColumnIndexFromTableCellNode(cell)
 
-    return this.editor.getEditorState().read(() => {
-      return $getTableColumnIndexFromTableCellNode(currentCell)
-    }) ?? 0
-  }
+      const rowCells = rows[currentRowIndex]?.getChildren() ?? null
```
samuelpecher: This `?` and `?? []` also feels unnecessary unless Lexical is returning duff indexes.

src/elements/table/table_controller.js
```diff
@@ -204,12 +163,10 @@ export class TableController {
     // We wait for next frame, otherwise table operations might not have completed yet.
     await nextFrame()
 
-    if (!this.currentTableNode) return
-
-    const rows = this.tableRows
-    if (!rows) return
+    const tableState = this.tableState
+    if (!tableState) return
```
samuelpecher: This construction suggests that we're being cautious not to call `this.tableState` repeatedly for performance reasons, which suggests either a more appropriate name is `getTableState()` or  `get tableState()` should have memoization (and invalidation) built-in.

src/elements/table/table_controller.js
```diff
@@ -275,46 +233,38 @@ export class TableController {
...
 
   #isCurrentCellEmpty() {
-    if (!this.currentTableNode) return false
-
-    const cell = this.currentCell
-    if (!cell) return false
+    const { tableNode, cell } = this.tableState
+    if (!tableNode || !cell) return false
 
     return cell.getTextContent().trim() === ""
```
samuelpecher: This is more "blank" than "empty". For empty I'd expect `getTextContentSize() === 0`

src/elements/table/table_tools.js
```diff
@@ -214,15 +217,16 @@ export class TableTools extends HTMLElement {
...
-
-      const tableNode = this.tableController.currentTableNode
-      if (tableNode) {
-        this.#show()
-      } else {
-        this.#hide()
-      }
+        this.tableController.updateSelectedTable()
+
+        const tableState = this.tableController.tableState
```
samuelpecher: This can be destructured

src/elements/table/table_tools.js
```diff
@@ -283,10 +287,10 @@ export class TableTools extends HTMLElement {
   }
 
   #setTableCellFocus() {
-    const cell = this.tableController.currentCell
-    if (!cell) return
+    const tableState = this.tableController.tableState
```
samuelpecher: This can be destructured

src/elements/table/table_controller.js
samuelpecher, line 1: I like the simplification that `tableState` gives, and nice work on the memoization

src/editor/command_dispatcher.js
```diff
@@ -228,23 +233,41 @@ export class CommandDispatcher {
...
 
   #registerCommandHandler(command, priority, handler) {
-    this.editor.registerCommand(command, handler, priority)
+    this.unregisterCommands = this.unregisterCommands || []
+    this.unregisterCommands.push(this.editor.registerCommand(command, handler, priority))
   }
 
   #registerKeyboardCommands() {
-    this.editor.registerCommand(KEY_TAB_COMMAND, this.#handleTabKey.bind(this), COMMAND_PRIORITY_NORMAL)
+    this.unregisterKeyboardCommands = this.unregisterKeyboardCommands || []
```
samuelpecher: I can't seem to find where this is called.

```suggestion
    this.unregisterKeyboardCommands ||= []
```

src/elements/code_language_picker.js
```diff
@@ -68,7 +73,7 @@ export class CodeLanguagePicker extends HTMLElement {
   }
 
   #monitorForCodeBlockSelection() {
-    this.editor.registerUpdateListener(() => {
+    this.unregisterUpdateListener = this.editor.registerUpdateListener(() => {
       this.editor.getEditorState().read(() => {
```
samuelpecher:
```suggestion
    this.unregisterUpdateListener = this.editor.registerUpdateListener(({ editorState }) => {
      editorState.read(() => {
```

src/editor/command_dispatcher.js
```diff
@@ -228,23 +233,41 @@ export class CommandDispatcher {
...
+      root.addEventListener("dragleave", this.#handleDragLeave.bind(this))
     }
   }
 
+  #unregisterDragAndDropHandlers() {
+    const root = this.editor.getRootElement()
+    root.removeEventListener("dragover", this.#handleDragOver.bind(this))
+    root.removeEventListener("drop", this.#handleDrop.bind(this))
+    root.removeEventListener("dragenter", this.#handleDragEnter.bind(this))
+    root.removeEventListener("dragleave", this.#handleDragLeave.bind(this))
```
samuelpecher: `bind()` creates a new function reference, so these won't be unregistered.

src/elements/table/table_controller.js
```diff
@@ -31,77 +33,51 @@ export class TableController {
...
-  get tableRows() {
-    return this.editor.getEditorState().read(() => {
-      return this.currentTableNode?.getChildren()
-    }) ?? null
+      this.#cachedTableState = { cell, tableNode, rows, currentRowIndex, currentColumnIndex, rowCells, columnCells }
+      return this.#cachedTableState
+    })
   }
 
   updateSelectedTable() {
```
samuelpecher, line 83: It's worth invalidating the cached state here, in case this is called before the update listener

src/elements/toolbar.js
```diff
@@ -86,10 +96,32 @@ export class LexicalToolbarElement extends HTMLElement {
     }
   }
 
+  #cacheButtonsDOM() {
+    this.#buttons = new Map()
+    this.querySelectorAll("[name]").forEach(btn => {
```
samuelpecher: This might capture more than just buttons?

samuelpecher (commented):
A great step in the right direction. I particularly like the `tableState` extraction if we avoid any issues around batched updates potentially calling stale state.

src/elements/toolbar.js
zoltanhosszu: Good catch! 🙏

src/elements/editor.js
zoltanhosszu: I'll be honest, this was a suggestion from Claude, after analyzing Lexical+Lexxy code. Tests pass in all 3 variants, so we might not have a proper test for this?

src/editor/command_dispatcher.js
zoltanhosszu: I moved it to `this.unregisterCommands` instead of tracking these separately.

jorgemanrubia:
Great to improve performance but this works needs some validation test or benchmark with a before/after, or at least some test case where you can see the improvement clearly.

At first glance, not removing listeners shouldn't cause any memory leak in general, as they will be garbage collected by the browser when the elements get removed from the dom. It would be good to narrow down what this is fixing and keept the change focused on that.

samuelpecher:
@zoltanhosszu I seem to remember the initial issue was with large tables?

It's a good point generally that we don't have a harness to performance test Lexxy.

zoltanhosszu:
Correct @samuelpecher, the main trigger was tables with 2,000+ rows. But also any long document results in a performance hit (long text docs, or many code blocks). I can separate out some parts of this PR (e.g. the querySelector changes) if you think that would make more sense!

PR #724, opened 2026-02-10
Remove wrapping <div>, clean up & refactor

samuelpecher:
- Remove two layers of `<div>` wrapping, both in the rails helper and in js API's `set value()`
- Clean-up value-set unwrapping code
- Refactor functions used solely within an `editor.update()` to not create a new update loop
- Refactor functions using a listener to use the passed state
- Remove console warnings for WrappedTableNode
- Turns off stimulus debug

cc @packagethief


src/elements/editor.js
```diff
@@ -196,23 +196,37 @@ export class LexicalEditorElement extends HTMLElement {
...
+  // We wrap those in <p>
+  #wrapTextNode(node) {
+    if (!$isTextNode(node)) return node
+
+    const paragraph = $createParagraphNode()
+    paragraph.append(node)
+    return paragraph
+  }
+
+  // Custom decorator block elements such action-text-attachments get wrapped into <p> automatically by Lexical.
```
packagethief: Missing word: "such *as* action-text-attachments"

test/dummy/app/javascript/controllers/application.js
```diff
@@ -3,7 +3,7 @@ import { Application } from "@hotwired/stimulus"
 const application = Application.start()
 
 // Configure Stimulus development experience
-application.debug = true
+application.debug = false
```
packagethief, line 6: /play live

PR #726, opened 2026-02-10
Provisional paragraph

samuelpecher:
Introduce a ProvisionalParagraphNode and accompanying extension to solve the issue of Lexical selection between DecoratorNodes and the editor boundaries. These nodes serve as selection targets for the cursor without rendering output. Previously Lexical would introduce a fake cursor which appeared foreign and behaved sub-optimally, particularly in image galleries (to be introduced in #716).

- `ProvisionalParagraph`s are created when a `DecoratorNode` or non-text `ElementNode` are neighbors or at the top/bottom of the editor.
- A `ProvisionalParagraph` is added to an empty editor
- They are converted to a regular `ParagraphNode` on edit
- They are removed when they are no longer required

I've also defined the transforms directly on the node, with Lexical's new `static transform` API, which keeps the code neater.

This cleans-up some edge cases:
- inserting `<p>` around nodes to ensure selection is fluid
- handling empty editor state


src/editor/contents.js
```diff
@@ -551,20 +548,25 @@ export default class Contents {
     const firstNode = nodes[0]
     const lastNode = nodes[nodes.length - 1]
 
-    return firstNode?.getPreviousSibling() || lastNode?.getNextSibling()
+    const previousSibling = firstNode?.getPreviousSibling()
+    const isPreviousSibling = Boolean(previousSibling)
+
+    const previousOrNextSibling = previousSibling || lastNode && $getNextSiblingOrParentSibling(lastNode)?.[0]
+
+    return [ previousOrNextSibling, isPreviousSibling ]
```
packagethief: I think we should probably rename this function now that it returns more detail, namely the direction. Renaming to something like `findAdjacentNodeWithDirection` would clarify. If you wanted to be more explicit, returning an Object would let you name the values:

```js
#findAdjacentNodeWithDirection(nodes) {
  const firstNode = nodes[0]
  const lastNode = nodes[nodes.length - 1]

  const previousSibling = firstNode?.getPreviousSibling()
  const nextSibling = lastNode && $getNextSiblingOrParentSibling(lastNode)?.[0]

  if (previousSibling) {
    return { node: previousSibling, direction: "previous" }
  } else if (nextSibling) {
    return { node: nextSibling, direction: "next" }
  } else {
    return { node: null, direction: null }
  }
}
```

But actually, maybe there's an opportunity to be more targeted with naming and purpose, since we only use these functions for one thing. Purpose-named functions are better than loosely connected variables. With something like this, we describe the intent, not the mechanism:

```js
deleteSelectedNodes() {
  this.editor.update(() => {
    if (!this.#selection.hasNodeSelection) return

    const nodesToRemove = $getSelection().getNodes()
    if (nodesToRemove.length === 0) return

    const focusTarget = this.#findFocusTargetAfterDeletion(nodesToRemove)

    this.#deleteNodes(nodesToRemove)
    this.#selectAfterDeletion(focusTarget)

    this.editor.focus()
  })
}

#findFocusTargetAfterDeletion(nodes) {
  const firstNode = nodes[0]
  const lastNode = nodes[nodes.length - 1]

  const previousSibling = firstNode?.getPreviousSibling()
  const nextSibling = lastNode && $getNextSiblingOrParentSibling(lastNode)?.[0]

  if (previousSibling) {
    return { node: previousSibling, direction: "previous" }
  } else if (nextSibling) {
    return { node: nextSibling, direction: "next" }
  } else {
    return { node: null, direction: null }
  }
}

#selectAfterDeletion({ node, direction }) {
  const root = $getRoot()

  if (root.getChildrenSize() === 0) {
    const newParagraph = $createParagraphNode()
    root.append(newParagraph)
    newParagraph.selectStart()
  } else if (node) {
    if ($isTextNode(node) || $isParagraphNode(node)) {
      direction === "previous" ? node.selectEnd() : node.selectStart()
    } else {
      direction === "previous" ? node.selectNext(0, 0) : node.select()
    }
  }
}
```

What do you think? If you think it's a bit too much, renaming the function to include the directionality bit would go a long way 😉

src/extensions/provisional_paragraph_extension.js
```diff
@@ -0,0 +1,55 @@
...
+
+  for (const node of $firstToLastIterator(rootNode)) {
+    const nextNode = node.getNextSibling()
+    if (ProvisionalParagraphNode.neededBetween(node, nextNode)) {
+      node.insertAfter(new ProvisionalParagraphNode)
+    }
+  }
+}
+
+function $removeUneededProvisionalParagraphs(rootNode) {
```
packagethief: Typo: "uneeded" -> "unneeded"

src/editor/contents.js
samuelpecher: I would have gone for the latter, clearer changes but since `deleteSelectedNodes` is actually no longer called I've removed these functions ✂✂✂

PR #731, opened 2026-02-13
Update JavaScript assets path in initializer

### LINKED ISSUE #730: Engine registers wrong asset path for Javascript files
The `lexxy.assets` initializer in [`lib/lexxy/engine.rb`](https://github.com/basecamp/lexxy/blob/main/lib/lexxy/engine.rb)  adds `app/javascript` to the asset paths:                           

```
  initializer "lexxy.assets" do |app|                                                                                   
    if Rails.application.config.respond_to?(:assets)                                                                    
      app.config.assets.paths << root.join("app/assets/stylesheets")
      app.config.assets.paths << root.join("app/javascript")
    end
  end
```

However, the JavaScript files actually live in `app/assets/javascript/`, not `app/javascript/` (which doesn't exist in the gem).

This means the asset pipeline never discovers the JS files. The stylesheet path is correct (`app/assets/stylesheets`), so CSS loads fine.

jfi:
The `lexxy.assets` initializer in `lib/lexxy/engine.rb` added `app/javascript` to the asset paths, but the correct path is `app/assets/javascript`.

Fixes #730.

### DIFF
--- a/lib/lexxy/engine.rb
+++ b/lib/lexxy/engine.rb
@@ -29,7 +29,7 @@ module Lexxy
     initializer "lexxy.assets" do |app|
       if Rails.application.config.respond_to?(:assets)
         app.config.assets.paths << root.join("app/assets/stylesheets")
-        app.config.assets.paths << root.join("app/javascript")
+        app.config.assets.paths << root.join("app/assets/javascript")
       end
     end
 


samuelpecher:
The [rails guides suggests the line isn't actually needed](https://guides.rubyonrails.org/asset_pipeline.html#asset-organization) at all. @jfi were you getting an error using the engine?

PR #736, opened 2026-02-17
Add upload lifecycle events

### LINKED ISSUE #735: Add events for attachment upload start/end
Rails with Trix has `direct-upload:start` and `direct-upload:end` which are useful for blocking form submissions while uploads are in progress. Currently I'm query the DOM for the `progress` element as a proxy for an in-progress upload and using the `change` event to check whether an upload started/stopped:

```js
// form_controller.js
export default class extends Controller {
  // Guard programmatic submissions
  submit() {
    if (this.#uploadInProgress) return
    // ...
  }

  // Disable submit buttons reactively
  toggleSubmission() {
    this.#submitters.forEach(el => el.disabled = this.#uploadInProgress)
  }

  get #uploadInProgress() {                                                              
    return this.element.querySelector("lexxy-editor progress") !== null
  }

  get #submitters() {
    return Array.from(this.element.elements).filter(el =>
      el.type === "submit" || (el.tagName === "BUTTON" && el.type !== "button")
    )
  }
}
```

```html+erb
<%= form.rich_text_area :description, data: {
  action: "lexxy:change->form#toggleSubmission"
} %>

### PR DESCRIPTION
**Pull request by MatheusRich**
Dispatches lexxy:upload-start, lexxy:upload-progress, and lexxy:upload-end events on the editor element during file uploads. This lets consumers react to the upload lifecycle — for example, disabling submit buttons or showing custom progress UI.

Events bubble up from the <lexxy-editor> element with relevant detail (file, progress, error).

Closes #735

### CODE REVIEW

File: app/assets/javascript/lexxy.js [hunk 1]
```diff
@@ -7630,6 +7637,14 @@ class ActionTextAttachmentUploadNode extends ActionTextAttachmentNode {
     const conversion = new AttachmentNodeConversion(this, blob);
     return conversion.toAttachmentNode()
   }
+
+  #dispatchEvent(name, detail) {
+    dispatch(this.#editorElement, name, detail);
+  }
+
+  get #editorElement() {
+    return this.editor.getRootElement().parentElement
```
MatheusRich: I don't love this. We could instead invert the dependency here and pass `editorElement` instead of `editor`?

src/nodes/action_text_attachment_upload_node.js
```diff
@@ -209,6 +216,14 @@ export class ActionTextAttachmentUploadNode extends ActionTextAttachmentNode {
...
   }
+
+  #dispatchEvent(name, detail) {
+    dispatch(this.#editorElement, name, detail)
+  }
+
+  get #editorElement() {
+    return this.editor.getRootElement().parentElement
+  }
 }
```
samuelpecher: > I don't love this. We could instead invert the dependency here and pass editorElement instead of editor?

@MatheusRich I agree this doesn't feel great. `this.editor` isn't passed in: the reference is [created at node init](https://github.com/basecamp/lexxy/blob/4bcf8d0213971560e90734493f8bd835addc3642/src/nodes/action_text_attachment_node.js#L92). 

Nodes having knowledge of top-level DOM, even if only to fire an event, feels against the grain of Lexical's design. How would you feel about firing a bubbling event against the Node's DOM? This could be achieved by passing the DOM ref we already have in `createDOM`, or using [`getElementByKey`](https://github.com/facebook/lexical/blob/8d64fa83eb6b83134ced3eb8c129289838364b65/packages/lexical/src/LexicalEditor.ts#L1247) if the delegate had the node `key`.

Alternatively, the messages might be passed via registered Lexical commands, but that facade might be too much weight for this?

MatheusRich: @samuelpecher I've changed the implementation. Seems much simpler now. Thank you!

test/system/events_test.rb
```diff
@@ -62,6 +62,29 @@ class EventsTest < ApplicationSystemTestCase
...
+  test "dispatch lexxy:upload-progress during upload" do
+    visit edit_post_path(posts(:empty))
+
+    attach_file file_fixture("example.png") do
+      click_on "Upload file"
+    end
+
+    assert_image_figure_attachment content_type: "image/png", caption: "example.png"
+    assert_dispatched_event "lexxy:upload-progress"
+  end
```
samuelpecher: Could this be combined with the test above?

MatheusRich: Done. Docs were also added.

PR #738, opened 2026-02-17
Add SoftBreakExtension for opt-in soft line break behavior

flavorjones:
When loaded, the `SoftBreakExtension` changes some keystrokes to preserve true markdown-style vertical whitespace:

- Single Enter inserts a `<br>` instead of a paragraph break,
- double Enter converts to a paragraph break,
- and Backspace at a paragraph start merges with a `<br>`.

This allows apps like BC5 and Fizzy to format pasted Markdown properly while preserving the application-specific typing experience.

Consumers opt in via:

```
  import { configure, SoftBreakExtension } from "lexxy"
  configure({ global: { extensions: [SoftBreakExtension] } })
```

ref: https://app.3.basecamp.com/2914079/buckets/44335813/messages/9589745694


src/extensions/soft_break_extension.js
```diff
@@ -0,0 +1,186 @@
...
+  name: "lexxy/soft-break",
+  register(editor) {
+    editor.registerCommand(
+      KEY_ENTER_COMMAND,
+      (event) => handleEnterKey(event),
+      COMMAND_PRIORITY_NORMAL
+    )
+
+    editor.registerCommand(
+      KEY_BACKSPACE_COMMAND,
```
samuelpecher, line 26: I think the better move here is to move an abstraction level up and handle  `INSERT_PARAGRAPH_COMMAND` and `DELETE_CHARACTER_COMMAND` (with backwards=true) unless you really need to handle the event.

src/extensions/soft_break_extension.js
```diff
@@ -0,0 +1,186 @@
...
+
+function $findParagraphNode(node) {
+  let currentNode = node
+  while (currentNode) {
+    if ($isParagraphNode(currentNode)) {
+      return currentNode
+    }
+    currentNode = currentNode.getParent()
+  }
+  return null
```
samuelpecher, line 92: this pattern can usually be collapsed into `$getNearestNodeOfType(node, ParagraphNode)` or `$findMatchingParent(node, $isParagraphNode)` ( I prefer n ° 1 :wink:)

There's a few of these in the code due to be tightened

src/extensions/soft_break_extension.js
```diff
@@ -0,0 +1,186 @@
...
+}
+
+function $mergeParagraphsWithLineBreak(firstParagraph, secondParagraph) {
+  const lineBreak = $createLineBreakNode()
+  firstParagraph.append(lineBreak)
+
+  const secondChildren = secondParagraph.getChildren()
+  secondChildren.forEach(child => {
+    firstParagraph.append(child)
+  })
```
samuelpecher, line 172: Very nittish: `append` can take a destructed array


```suggestion
    firstParagraph.append(...secondChildren)
```

OR a step further
```suggestion
    firstParagraph.append(lineBreak, ...secondChildren)
```

src/extensions/soft_break_extension.js
```diff
@@ -0,0 +1,186 @@
...
+  if ($isRangeSelection(selection)) {
+    lineBreak.selectNext()
+  }
+}
+
+export default class SoftBreakExtension extends LexxyExtension {
+  get lexicalExtension() {
+    return SoftBreakExtensionDefinition
+  }
+}
```
samuelpecher, line 186: As a user of Lexxy Extensions, does the API feel right?

flavorjones:
Please see https://github.com/basecamp/lexxy/pull/741 for an alternative approach.

flavorjones:
Closing in favor of #741 which is simpler, more extensible, and doesn't modify the editing experience.

PR #740, opened 2026-02-20
Fix form attribute being discarded in rich_text_area_tag

ophilbert:
lexxy_rich_textarea_tag was removing 'form' from html options.
This prevented using lexxy from outside a form like we can do with regular inputs:

`<input form="myform">
`

This change was made in actionText [when they extracted ActionText::TrixEditor](https://github.com/rails/rails/commit/cb0fd2254acb8b7c1c9354ef34538ba51bf28d1f)

This attribute should pass through so we can associate lexxy editor with external form.


samuelpecher:
Thanks @ophilbert

PR #741, opened 2026-02-20
Introduce markdownPasteTransforms

flavorjones:
... to allow the application to modify the DOM generated by a markdown paste event. One or more functions can be composed over the DOM by setting an array of mutating transform functions via `configure()`:

```
<script type="module">
  import { configure, addBlockSpacers } from "lexxy"
  configure({ default: { markdownPasteTransforms: [ addBlockSpacers ] } })
</script>
```

In practice, this is probably most useful for adjusting whitespace, and so this also introduces `addBlockSpacers` which inserts `<p><br></p>` spacer elements between adjacent block-level siblings, except after headings (h1-h6).

Multiple transforms are applied in left-to-right order, so `[ a, b ]` will result in `b(a(dom))`

This is an alternative to https://github.com/basecamp/lexxy/pull/738

- [x] tested with basecamp

Example of what the `addBlockSpacers` looks like on a site that overrides Lexxy's CSS variable `--lexxy-content-margin` to avoid margin between block elements (before and after):

screenshot-2026-02-20_11-43-14


flavorjones:
@zoltanhosszu pointed out that Marked has some hooks which might make it easier to implement application-provided mutations, that might be a better approach to composing functions.

samuelpecher:
I was also wondering if this was worth introducing a new paradigm (input processing via config) vs implementing a `lexxy:insert-markdown` event like for [files](https://basecamp.github.io/lexxy/events.html#lexxyfile-accept) or [urls](https://basecamp.github.io/lexxy/events.html#lexxyinsert-link).

flavorjones:
closing in preference to https://github.com/basecamp/lexxy/pull/757

PR #749, opened 2026-02-24
Fix HTML escaping in code blocks

zoltanhosszu:
HTML code blocks are broken at the moment when the value is loaded into the editor.

---

#### The Bug
The issue is about how Rails' `content_tag` handles HTML entities in attribute values vs wrapping the value as element content.

#### The flow
The `sandbox's _default.html.erb` template contains a `<pre>` block with `&lt;` entities (line 32):

`&lt;lexxy-editor placeholder="It all starts here..."&gt;`

This HTML string is passed as the value parameter to `lexxy_rich_textarea_tag`.

On line 11, the value is set as an option: `options[:value] ||= value`

On line 17, `content_tag("lexxy-editor", "", options)` renders it as an HTML attribute:

```html
<lexxy-editor value="&lt;lexxy-editor...&gt;...">
```

Rails' content_tag HTML-escapes attribute values. So the `&lt;` in the original content gets double-escaped to `&amp;lt;` in the raw HTML, which the browser then decodes back to the literal string `&lt; `when you read the attribute via `getAttribute("value")`. This is correct.

JavaScript reads it with `this.getAttribute("value")` (line 336), which returns the decoded attribute value — the `&lt;` entities are preserved as literal text `&lt;`.

When `value = "<div>#{value}</div>"` was removed, the value goes directly into `options[:value]` as an `ActiveSupport::SafeBuffer` (because it came from `render()`).

The key: `render()` returns an `html_safe` string. When Rails' `content_tag` sees an `html_safe` value in an attribute, it does NOT double-escape it. So `&lt;` in the original template stays as `&lt;` in the raw HTML attribute. The browser then decodes this to an actual `<` character when reading the attribute, which breaks the `<pre>` content because it becomes real HTML tags instead of visible text.

### DIFF
--- a/test/helpers/lexxy/tag_helper_test.rb
+++ b/test/helpers/lexxy/tag_helper_test.rb
@@ -26,4 +26,18 @@ class Lexxy::TagHelperTest < ActionView::TestCase
       assert_equal "<p>Sample Content</p>", lexxy_editor["value"]
     end
   end
+
+  test "#lexxy_rich_textarea_tag preserves HTML entities in code blocks" do
+    code_html = '<pre data-language="html">&lt;div&gt;test&lt;/div&gt;</pre>'.html_safe
+
+    render inline: <<~ERB, locals: { code_html: code_html }
+      <%= lexxy_rich_textarea_tag :body, code_html %>
+    ERB
+
+    assert_dom "lexxy-editor", count: 1 do |lexxy_editor, *|
+      value = lexxy_editor["value"]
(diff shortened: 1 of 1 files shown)


samuelpecher:
Nice find @zoltanhosszu. I've added a test to cover regression.
I've also noticed a 20% test slow-down with the string conversion, so it would be ideal to shed the overhead by just clearing the `html_safe` property.

cc @packagethief

samuelpecher:
The runner seems stuck; merging based on local CI.

image

PR #756, opened 2026-02-26
Pin the correct lexxy in the installation instructions

frenkel:


frenkel:
I think my change is not needed as the rails gem includes the correct lexxy and that should be used.

PR #772, opened 2026-03-02
Add native KaTeX-based math rendering

CharlieLeee:
## Summary

- Add inline (`$...$` auto-detection) and block (`$$` + Enter) math support powered by KaTeX
- `InlineMathNode` and `BlockMathNode` as Lexical DecoratorNodes
- `<lexxy-math-editor>` custom element with live preview for editing
- `renderContentMath()` helper for display-page rendering (same pattern as `highlightCode()`)
- Toolbar button, command dispatcher, DOMPurify and ActionText sanitizer integration
- 13 new tests covering math rendering and detection regex
- Togglable via `math: true/false` in Lexxy config

## Test plan

- [x] Type `$E=mc^2$` in editor — auto-converts to rendered inline math
- [x] Type `$$` at start of empty line + Enter — opens block math editor with live preview
- [x] Click rendered math — opens editor for modification
- [x] Escape or Cmd+Enter — confirms edit
- [x] Save post and view — math renders correctly on display page
- [x] `npm test` — 21 tests pass
- [x] `npm run build` — builds successfully


src/extensions/math_extension.js
```diff
@@ -0,0 +1,215 @@
...
+
+    return defineExtension({
+      name: "lexxy/math",
+      nodes: [InlineMathNode, BlockMathNode],
+      register(editor) {
+        return mergeRegister(
+          // Auto-detect $...$ inline math
+          editor.registerNodeTransform(TextNode, (textNode) => {
+            $detectInlineMath(textNode)
+          }),
```
samuelpecher: No need to wrap the function, since it's not a class method
```suggestion
          editor.registerNodeTransform(TextNode, $detectInlineMath),
```

src/extensions/math_extension.js
```diff
@@ -0,0 +1,215 @@
...
+          editor.registerCommand(INSERT_BLOCK_MATH_COMMAND, () => {
+            $insertBlockMath(editor, editorElement)
+            return true
+          }, COMMAND_PRIORITY_NORMAL),
+
+          // Command: insert inline math
+          editor.registerCommand(INSERT_INLINE_MATH_COMMAND, () => {
+            $insertInlineMath(editor)
+            return true
+          }, COMMAND_PRIORITY_NORMAL),
```
samuelpecher: I'd `return true` from the handler and do this:

```suggestion
          editor.registerCommand(INSERT_INLINE_MATH_COMMAND, 
            () => $insertInlineMath(editor), COMMAND_PRIORITY_NORMAL
          ),
```

src/nodes/block_math_node.js
```diff
@@ -0,0 +1,115 @@
...
+    const deleteButton = createElement("lexxy-node-delete-button")
+    figure.appendChild(deleteButton)
+
+    preview.addEventListener("click", (event) => {
+      event.stopPropagation()
+      figure.dispatchEvent(new CustomEvent("lexxy:edit-math", {
+        bubbles: true,
+        detail: { nodeKey: this.getKey(), latex: this.__latex, displayMode: true, targetElement: figure }
+      }))
+    })
```
samuelpecher: This pattern is usually better as a `CLICK_COMMAND` handler, which removes the need for internal events. See #727

src/nodes/block_math_node.js
```diff
@@ -0,0 +1,115 @@
...
+    return "block_math"
+  }
+
+  static clone(node) {
+    return new BlockMathNode({ latex: node.__latex }, node.__key)
+  }
+
+  static importJSON(serializedNode) {
+    return new BlockMathNode({ latex: serializedNode.latex })
+  }
```
samuelpecher: This boilerplate [can be replaced](https://lexical.dev/docs/concepts/nodes#extending-decoratornode-using-config) with a `$config` call

```js
  $config() {
    return this.config({
       name: "block_math"
    })
  }

File: src/extensions/math_extension.js (hunk 1 above)
Comment (samuelpecher) [inline src/extensions/math_extension.js]: I haven't used it yet, but this seems to be the use case for [`registerLexicalTextEntity`](https://lexical.dev/docs/concepts/transforms#registerlexicaltextentity)

File: src/extensions/math_extension.js [hunk 5]
```diff
@@ -0,0 +1,215 @@
...
+  })
+
+  return true
+}
+
+function $insertBlockMath(editor, editorElement) {
+  const node = new BlockMathNode({ latex: "" })
+
+  const selection = $getSelection()
+  if ($isRangeSelection(selection)) {
```
samuelpecher: Checking the inverse allow flattening the code for better readability

```suggestion
  if (!$isRangeSelection(selection)) return false
```

src/extensions/math_extension.js
```diff
@@ -0,0 +1,215 @@
...
+    const anchorNode = selection.anchor.getNode()
+    const topElement = anchorNode.getTopLevelElementOrThrow()
+
+    // If current line is empty paragraph, replace it
+    if ($isParagraphNode(topElement) && topElement.getTextContent() === "") {
```
samuelpecher: `topElement.isEmpty()`

src/extensions/math_extension.js
```diff
@@ -0,0 +1,215 @@
...
+      topElement.replace(node)
+    } else {
+      topElement.insertAfter(node)
+    }
+
+    // Ensure paragraph below
+    if (!node.getNextSibling()) {
+      const paragraph = $createParagraphNode()
```
samuelpecher: In theory ProvisionalParagraphs (#726) handle selection-after-decorator issues.

src/extensions/math_extension.js
```diff
@@ -0,0 +1,215 @@
...
+
+        if (node instanceof BlockMathNode) {
+          const newNode = new BlockMathNode({ latex: newLatex })
+          node.insertAfter(newNode)
+          node.remove()
+        } else if (node instanceof InlineMathNode) {
+          const newNode = new InlineMathNode({ latex: newLatex })
+          node.insertAfter(newNode)
+          node.remove()
+        }
```
samuelpecher: This construction leaves me wondering about not recycling the nodes with a new `setLatex` API
```js
if ($isMathNode(node) {
  node.setLatex(newLatex)
}
```

Then, standard lexical node update mechanisms kick in

samuelpecher (commented):
Hey @CharlieLeee - this is a cool use of Lexxy Extensions! This is the way we'd hoped the editor could be extended for application-specific needs. I've noted we need a clear hook for [allowing additional DOM](https://github.com/basecamp/lexxy/issues/781).

Since it's quite application specific, I don't think it need to be part of core Lexxy, especially with how you can reach deep into the editor to provide new functionalities like this. It would be useful to publish as an extension so others can import into their projects.

I've left my thoughts on a few potential improvements.

CharlieLeee:
Thanks for the thoughtful review, @samuelpecher! Really appreciate you taking the time.

I understand the reasoning for keeping core Lexxy lean. I'll look into publishing this as a standalone extension and will iterate on the improvements you suggested (`$config()`, `CLICK_COMMAND`, `setLatex`, etc.).

That said, I wanted to share some context: our team uses Basecamp daily for academic/technical collaboration, and the lack of math support in the editor is a real pain point for us. We often need to discuss equations and formulas, and currently have to resort to screenshots or external tools. Native LaTeX support would be a huge quality-of-life improvement for teams like ours.

If there's ever an appetite to include math as an opt-in extension in core (similar to how code highlighting is built in), we'd love to see that happen. Either way, happy to keep improving this implementation.

src/nodes/block_math_node.js
CharlieLeee: done

src/extensions/math_extension.js
CharlieLeee: done

src/extensions/math_extension.js
CharlieLeee: seems like `registerLexicalTextEntity`requires `Klass<T extends TextNode>`, but `InlineMathNode` extends `DecoratorNode`, so it probably doesn't apply here

src/extensions/math_extension.js
CharlieLeee: good call, I did look into it. 
The manual selection handling in `$handleBlockMathTrigger` is addressing a slightly different issue though: when the user types `$$` and presses Enter, the paragraph holding the cursor gets replaced with a BlockMathNode. Without moving selection first, Lexical throws error #19 (selection references removed nodes) and the editor freezes. Moving selection to the next paragraph before replacing avoids that. I think `ProvisionalParagraphs` handles decorator navigation rather than this specific case, but happy to revisit if I'm missing something!

src/extensions/math_extension.js
```diff
@@ -0,0 +1,217 @@
...
+      })
+
+      editor.focus()
+    }
+  })
+}
+
+function findDOMNodeForKey(rootElement, nodeKey) {
+  // Lexical decorates DOM elements with data attributes we can search
+  return rootElement?.querySelector(`[data-lexical-decorator="true"]`) || null
```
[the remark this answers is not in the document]
samuelpecher: see [`editor.getElementByKey`](https://github.com/facebook/lexical/blob/a3d33be3ee82b0fb3a22ec6bb2e8950e30f9b755/packages/lexical/src/LexicalEditor.ts#L1247)

src/extensions/math_extension.js
samuelpecher: Ah, it seemed the `TextNode => DecoratorNode` transform would fit the use case

src/extensions/math_extension.js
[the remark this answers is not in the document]
CharlieLeee: already updated, `findDOMNodeForKey` was removed and now use `editor.getElementByKey(nodeKey)` in all call sites. Thanks for the pointer!

adam-h:
We are using Lexxy with separate handling for LaTeX (via Mathjax in our case), it would be great to have this built-in (or as an extension) to Lexxy for us too.

CharlieLeee:
Update: I've published this as a standalone extension, [lexxy-math](https://github.com/CharlieLeee/lexxy-math), with all review feedback incorporated. @adam-h it should work for your use case too, feedback and contributions welcome!

@samuelpecher would you prefer to close this PR in favor of the standalone package, or would it make sense to keep iterating here toward potential inclusion in core?

jorgemanrubia:
Thanks for this thorough implementation! As noted in #779, math rendering is outside Lexxy's core scope, but this would make a great standalone extension. Appreciate the effort and quality of the code.

PR #782, opened 2026-03-05
Add 'start' to allowed HTML attributes

adam-h:
In order for `ol` list elements to respect the numbers they were authored with we need to make sure that the [start](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/ol#start) attribute makes it through dom_purify and the ActionText allowlist.


adam-h:
Apologies, this is a dupe https://github.com/basecamp/lexxy/pull/713

PR #794, opened 2026-03-06
Export bare images as <img> when no sgid is present

lylo:
When an `` tag is imported into the editor (e.g. an external image URL), the img DOM converter wraps it in an `ActionTextAttachmentNode` with `contentType "image/*"` but no `sgid`. On save, `exportDOM()` was outputting an `<action-text-attachment>` tag without an sgid, which Action Text cannot resolve. This causes the attachment to render as a broken placeholder and, if an Active Storage blob URL was involved, the blob appears unattached and gets purged by Rails.

Instead, when there is no `sgid`, export the node as a plain `` tag so the image URL is preserved safely.


lylo:
@samuelpecher am I supposed to respond to the Copilot review or will a maintainer look at this? I think it's actually a pretty serious bug.

samuelpecher:
@lylo My understanding is ActionText resolves these as `ActionText::RemoteImage` when [it encounters`ActionText::Attachment.tag_name`](https://github.com/rails/rails/blob/aa6f191713e1823003e88c64d958e6f3ac81e51e/actiontext/lib/action_text/attachable.rb#L36) with no `sgid` but a `url`. As-is the node exports `url` from `img[src]`.

The image should be rendered through the [included ActionText view](https://github.com/rails/rails/blob/aa6f191713e1823003e88c64d958e6f3ac81e51e/actiontext/app/views/action_text/attachables/_remote_image.html.erb).

I had a quick test with the dummy app at `http://lexxy.localhost:3000/posts/` and a remote image stayed through save, and edit as expected. Can you specify what the specific flow you're seeing doesn't work properly?

lylo:
> @olly My understanding is ActionText resolves these as `ActionText::RemoteImage` when [it encounters`ActionText::Attachment.tag_name`](https://github.com/rails/rails/blob/aa6f191713e1823003e88c64d958e6f3ac81e51e/actiontext/lib/action_text/attachable.rb#L36) with no `sgid` but a `url`. As-is the node exports `url` from `img[src]`.
> 
> The image should be rendered through the [included ActionText view](https://github.com/rails/rails/blob/aa6f191713e1823003e88c64d958e6f3ac81e51e/actiontext/app/views/action_text/attachables/_remote_image.html.erb).
> 
> I had a quick test with the dummy app at `http://lexxy.localhost:3000/posts/` and a remote image stayed through save, and edit as expected. Can you specify what the specific flow you're seeing doesn't work properly?

Hi @samuelpecher  (it's @lylo not @olly btw!)

Ok, so the problem I've seen is more nuanced and perhaps not a Lexxy issue per se, but the remote image wrapped in `<action-text-attachment>` is the problem.

In Pagecord I've had a small amount of data loss because (I think – very hard to reproduce) of this scenario.

  1. User uploads image to Post A via proper ActiveStorage attachment flow which gives the attachment an sgid
  2. User views Post A in the browser, right-click the image → Copy Image (Mac)
  3. Paste into Post B's editor
  4. Lexxy imports this image into Post B as `<action-text-attachment content-type="image/*">` with the rendered URL, _no sgid_. This post looks fine and both posts can be edited in Lexxy so it looks as if the image is stored by Pagecord – no indication that one of them is a remote image.
  5. Post B has no attachment record → blob is unprotected
  6. Delete Post A → the blob is purged → Post B's image breaks

Does this make sense? If the remote image wasn't wrapped in `<action-text-attachment>` then this scenario could not happen.

I'm not quite sure what to do at this stage.

lylo:
@samuelpecher  I think being able to disable remote image pasting is the best solution _for me_ since there's no visual recognition that the image is "stored" or remote, and if it's remote but pointing back to your own storage (whether via a `/rails` path or a CDN path, as is likely the case with Pagecord) then you run the risk of data loss in the aforementioned scenario. I appreciate this is a different thing for this PR (which I guess should be closed) but before you close it I'd be interested in your thoughts about whether disabling remote image might be in scope for Lexxy? Other ideas would be highlighting somehow that an image is remote, or popping up a warning.

lylo:
I verified the behaviour in Fizzy btw.

1. Create card and upload an image correctly
2. Copy image (right click, copy image) onto clipboard
3. Create a new card, paste clipboard to create a remote image
4. Delete original card
5. New card image is busted because blog has been purged

I guess this is expected behaviour in some way, but it's very unclear what's going on for customers.

### First image


Screenshot 2026-03-11 at 07 08 30

### Second image (pasted remote image)
Screenshot 2026-03-11 at 07 08 36

### Second image after deleting first card

Screenshot 2026-03-11 at 07 09 00

samuelpecher:
@lylo I'm not sure there's a direct Lexxy solution here, it's essentially a back-end problem. If a image points to a remote `src` which is removed the image is "broken/lost" no matter whether the tag is an `` or `<action-text-attachment>`.

To prevent the purge, I'd want to link the new Record to the Attachment. My instinct would be to try pattern-match the `src`, extract `sgid`, and convert to an `<action-text-attachment sgid="...">`. This could be done server-side, or maybe we're missing a `lexxy:insert-html` event which would allow an application to do this client side.

lylo:
I've confirmed that this did cause data loss on our end — content originally created by Trix and stored via ActionText had proper `sgid` and `content type="image/jpeg"`, but after loading into Lexxy, editing (text-only changes, no image paste), and saving the stored HTML ended up with `content-type="image/*"` and no `sgid`. The blobs were then purged as unattached.

I've tried everything to reproduce without luck, so it's possible a more recent Lexxy version fixed the serialization round-trip, or there might have been an issue in my back-end code that caused the issue (again, unclear what if anything).

I suppose we need to close this PR anyway, sorry for the noise.

PR #796, opened 2026-03-06
Upload tweaks

zoltanhosszu:
This PR introduces several improvements to the attachment handling and gallery display in Lexxy.

- Fixed a bug where the file format class name were missing during upload
- Removed image container markup
- While an upload was ongoing in an empty editor, the placeholder text remained. This PR removes that
- Upload state looks better for all file types
image

### DIFF
--- a/app/assets/stylesheets/lexxy-editor.css
+++ b/app/assets/stylesheets/lexxy-editor.css
@@ -189,7 +189,6 @@
     progress {
       max-inline-size: 10ch;
       margin: auto;
-      margin-inline-end: 5ch;
     }
 
     &.attachment--preview {
@@ -197,7 +196,6 @@
         inset-block-start: 2ch;
         inset-inline-start: 0;
         inset-inline-end: 0;
-        margin: auto;
         position: absolute;
       }
     }


src/editor/contents.js
```diff
@@ -271,7 +271,9 @@ export default class Contents {
 
       if (selectLast && uploader.nodes?.length) {
         const lastNode = uploader.nodes.at(-1)
-        lastNode.selectEnd()
+        const nodeSelection = $createNodeSelection()
+        nodeSelection.add(lastNode.getKey())
+        $setSelection(nodeSelection)
```
samuelpecher: I think there's a `createNoteSelectionWith` helper

src/nodes/action_text_attachment_upload_node.js
```diff
@@ -207,7 +209,18 @@ export class ActionTextAttachmentUploadNode extends ActionTextAttachmentNode {
...
+      const selection = $getSelection()
+      const shouldTransferNodeSelection = $isNodeSelection(selection) && selection.has(replacedNodeKey)
+
+      const replacementNode = this.#toActionTextAttachmentNodeWith(blob)
+      this.replace(replacementNode)
+
+      if (shouldTransferNodeSelection) {
+        const nodeSelection = $createNodeSelection()
+        nodeSelection.add(replacementNode.getKey())
+        $setSelection(nodeSelection)
```
samuelpecher: `createNodeSelectionWith`

src/nodes/action_text_attachment_upload_node.js
```diff
@@ -207,7 +209,18 @@ export class ActionTextAttachmentUploadNode extends ActionTextAttachmentNode {
 
   #showUploadedAttachment(blob) {
     this.editor.update(() => {
-      this.replace(this.#toActionTextAttachmentNodeWith(blob))
+      const replacedNodeKey = this.getKey()
+      const selection = $getSelection()
+      const shouldTransferNodeSelection = $isNodeSelection(selection) && selection.has(replacedNodeKey)
```
samuelpecher: Does `replacementNode.isSelected()` not get you what you want here? 

For a DecoratorNode I think it's safe to assume the selection will be on the node. TextNode handling is different.

src/editor/command_dispatcher.js
```diff
@@ -189,7 +189,8 @@ export class CommandDispatcher {
       multiple: true,
       style: "display: none;",
       onchange: ({ target: { files } }) => {
-        this.contents.uploadFiles(files, { selectLast: true })
+        this.editorElement.focus()
+        requestAnimationFrame(() => this.contents.uploadFiles(files, { selectLast: true }))
```
samuelpecher: If we _have_ to await the frame I'd prefer  an async `onchange` with:


```suggestion
        this.editorElement.focus()
        await nextFrame()
        this.contents.uploadFiles(files, { selectLast: true })
```

Alternatively, (note to self:) if we passed a handler to `lexxy.focus(handler)` down into `editor.focus(handler)`, we'd remove this await.

```suggestion
        this.editorElement.focus(() => {
          this.contents.uploadFiles(files, { selectLast: true })
        })
```

zoltanhosszu: Removed this whole block, not necessary now.

src/nodes/action_text_attachment_upload_node.js
zoltanhosszu: Oh yes, it works! Updated :)

src/nodes/action_text_attachment_upload_node.js
```diff
@@ -207,7 +209,15 @@ export class ActionTextAttachmentUploadNode extends ActionTextAttachmentNode {
...
-      this.replace(this.#toActionTextAttachmentNodeWith(blob))
+      const shouldTransferNodeSelection = this.isSelected()
+
+      const replacementNode = this.#toActionTextAttachmentNodeWith(blob)
+      this.replace(replacementNode)
+
+      if (shouldTransferNodeSelection) {
+        const nodeSelection = $createNodeSelectionWith(replacementNode)
+        $setSelection(nodeSelection)
+      }
```
samuelpecher, line 220: If we ever allow multi-node selection, this should be 

```js
const selection = $getNodeSelection
selection.add(replacementNode.getKey())
```

PR #755, opened 2026-02-26
Delete button for nodes

zoltanhosszu:
This PR adds a new control for deleting certain nodes (attachments, horizontal divider) with the mouse. The handle shows up when a node is selected.

image

PR #803, opened 2026-03-09
Prompt updates

zoltanhosszu:
This PR fixes a few issues around prompts and mentions:

- Prompts are now hugging the right side of the editor if they would clip out of view
- Removed the left side margin of prompts that was due to broad `ul` targeting in `.lexxy-content`
- Prompts are now disabled in code blocks
- Fixed the positioning of the [recently added](https://github.com/basecamp/lexxy/pull/755) delete button


src/elements/prompt.js
samuelpecher: Would be good to use the passed editorState

```suggestion
    this.cursorPositionListener = this.#editor.registerUpdateListener(({ editorState }) => {
```
then...
```diff
- this.#editor.read()
+ editorState.read()
```

PR #804, opened 2026-03-09
Prevent Image Galleries removing trix <div> content

samuelpecher:
Ensures Lexxy only converts `<div>` elements to galleries when all the children are attachments.
- Added a text for the trix `<div>` + attachment import
- Also moved the code responsible for unwrapping `ActionTextAttachmentNode` out of `ParagraphNode` to the `AttachmentsExtension`
- Added `debug` to the test environment

### DIFF
--- a/test/browser/tests/trix_html.test.js
+++ b/test/browser/tests/trix_html.test.js
@@ -52,7 +52,7 @@ test.describe("Trix HTML", () => {
 
     test("strong del", async ({ editor }) => {
       await editor.setValue("<div><strong><del>wrong!</del></strong></div>")
-      await assertEditorHtml(editor, "<p><s><b><strong>wrong!</strong></b></s></p>")
+      await assertEditorHtml(editor, "<p><s><strong>wrong!</strong></s></p>")
     })
 
     test("del with color", async ({ editor }) => {
@@ -62,12 +62,12 @@ test.describe("Trix HTML", () => {
 
     test("strong with color", async ({ editor }) => {
       await editor.setValue('<div><strong style="color: yellow;">banana</strong></div>')
-      await assertEditorHtml(editor, '<p><b><mark style="color: yellow;"><strong>banana</strong></mark></b></p>')
+      await assertEditorHtml(editor, '<p><mark style="color: yellow;"><strong>banana</strong></mark></p>')
     })
 
     test("em with color", async ({ editor }) => {
       await editor.setValue('<div><em style="color: blue;">wave</em></div>')
-      await assertEditorHtml(editor, '<p><i><mark style="color: blue;"><em>wave</em></mark></i></p>')
+      await assertEditorHtml(editor, '<p><mark style="color: blue;"><em>wave</em></mark></p>')
     })
   })
 


test/test_helpers/editor_helper.rb
```diff
@@ -16,7 +16,7 @@ def assert_editor_plain_text(value)
 
   def assert_editor_html(expected = nil, &block)
     if block
-      Capybara.string(find_editor.value).instance_exec(&block)
+      assert Capybara.string(find_editor.value).instance_exec(&block)
     else
```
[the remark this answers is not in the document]
samuelpecher: The end command in a block will be an assertion, and return `true` or throw an error.

src/extensions/attachments_extension.js
```diff
@@ -23,13 +23,33 @@ export class AttachmentsExtension extends LexxyExtension {
...
           editor.registerCommand(DELETE_CHARACTER_COMMAND, $collapseIntoGallery, COMMAND_PRIORITY_NORMAL)
         )
       }
     })
   }
 }
 
+// Decorator nodes can be wrapped in a Paragraph Node by Lexical when contained in a <div>
+// We remove them, splitting the node as needed
+function $removeAttachmentFromParagraph(attachmentNode) {
```
jorgemanrubia: This sounds like *removing attachments*. I would call it something like *flatten attachments* or *unwrap attachments*, since they don't get removed, right?

test/test_helpers/editor_helper.rb
```diff
@@ -16,7 +16,7 @@ def assert_editor_plain_text(value)
 
   def assert_editor_html(expected = nil, &block)
     if block
-      Capybara.string(find_editor.value).instance_exec(&block)
+      assert Capybara.string(find_editor.value).instance_exec(&block)
```
jorgemanrubia, line 19: why this `assert`? This assumes the block evaluation returns a not falsey value. I think we use the block for adding custom assertions on the editor contents. It is not intuitive the return value of the block matters imo.

src/nodes/image_gallery_node.js
```diff
@@ -22,8 +23,7 @@ export class ImageGalleryNode extends ElementNode {
   static importDOM() {
     return {
       div: (element) => {
-        const containsAttachment = element.querySelector(`:scope > :is(${this.#attachmentTags.join()})`)
-        if (!containsAttachment) return null
+        if (!this.#allChildrenAreAttachments(element)) return null
```
jorgemanrubia: why is this needed? To deal with legacy data? To deal with some Trix data?

jorgemanrubia (commented):
👏 nice @samuelpecher

test/test_helpers/editor_helper.rb
samuelpecher: This is to prevent a "test has no assertions" warning when using the helper:
- Most tests have an assertion in `teardown` since we check for console messages
- When we skip the console check, assertions in the passed block are not counted, though they still fail as they throw.
- Given the convention is the last operation in a block will be an assertion returning truthy or raising a failure exception, this appeases the assertion counter and doesn't cause failures

src/nodes/image_gallery_node.js
samuelpecher: All Trix paragraphs are effectively within a `<div>`. The previous version was treating any `<div>` with at least one image as a gallery and discarding any non-attachment content.

There's a potentially simpler alternative, dropping the filtering...
```js
after: children => $descendantsMatching(children, this.isValidChild)
```
... would allow the `<div>` to become a gallery and would implicitly self-split around non-attachment content. Round-tripping any `<div>` with an image into a gallery and the back again didn't seem like the right path to me.

samuelpecher:
@lylo I assume your comment is intended for #794 ?

lylo:
> @lylo I assume your comment is intended for #794 ?

sorry yes 🤦‍♂️ #toomanyprs

src/nodes/image_gallery_node.js
samuelpecher: @jorgemanrubia looking at #830 it seems like this is finally the way to go?

jorgemanrubia: Yes 👍

PR #807, opened 2026-03-11
Add configurable remoteImages option

lylo:
## Summary

- Adds a `remoteImages` global config option (default `true`) that, when set to `false`, prevents remote-image attachment nodes from being created
- Pasted `` elements and `<action-text-attachment>` nodes without an `sgid` and an image content-type are silently dropped
- Config is checked at conversion time (not at editor init) so it works correctly with Lexical's cached `importDOM()` results

```js
Lexxy.configure({ global: { remoteImages: false } })
```

## Test plan

- [x] `yarn test` passes (27 tests including 8 new ones)
- [ ] Verify pasting images is blocked in sandbox with `remoteImages: false`
- [ ] Verify pasting images still works with default config
- [ ] Verify sgid-backed attachments and non-image attachments are unaffected


lylo:
@samuelpecher this was just a PoC to see how easy it might be to disable remote image pasting. seems pretty straightforward if it's something you think might be useful for other users of Lexxy. It's not an ideal UX for me, but it would allow me to prevent a scenario as discussed in https://github.com/basecamp/lexxy/pull/804

PR #819, opened 2026-03-13
Fix line breaks lost when pasting plain text

jorgemanrubia:
## Summary

Pasting text from external editors (e.g., iA Writer) into Lexxy collapsed all line breaks — single newlines between lines were silently dropped, producing a wall of text instead of preserving the original structure.

The root cause is the `marked` library's default CommonMark behavior: single `\n` characters are treated as soft breaks (rendered as spaces in HTML). Only double `\n\n` produced paragraph separators.

The fix passes `breaks: true` to `marked()`, which converts single newlines to `<br>` tags. Double newlines continue to produce separate `<p>` elements as before.

---

https://app.fizzy.do/5986089/cards/4991


flavorjones:
@jorgemanrubia This seems like a good change! But I don't understand how this fixes what Jason described, because the paste result that he presented has not been markdownified at all. Seems like that might be a different code path?

image

jorgemanrubia:
Yes @flavorjones! Discussed in campfire, the problem is that I did not manage to reproduce the reported bug here at all. Not even with IA writer and MacOS. I still merged it because it looked like a good change, as you said. I wonder if some other change fixed the problem JF was seeing.

PR #824, opened 2026-03-13
Fix dropped attachments landing at stale cursor position

jorgemanrubia:
## Summary

- Dropped files from the OS file manager now land at the drop location instead of the stale cursor position
- Added `dropFiles` method that uses `document.caretRangeFromPoint()` to position the selection at drop coordinates before uploading

Fixes https://app.fizzy.do/5986089/cards/1832

## Root cause

When dragging a file from the OS, the browser shows a visual drag caret but does **not** update the DOM selection (`window.getSelection()`). The `#handleDrop` method called `uploadFiles()` directly, which inserts at the current Lexical selection (derived from the stale DOM selection). Attachments landed at wherever the cursor was before the drag started.

## Changes

- **`src/editor/contents.js`** — New `dropFiles()` method with `#moveSelectionToPoint()` helper that uses `document.caretRangeFromPoint(clientX, clientY)` to position the DOM selection at the drop coordinates before uploading
- **`src/editor/command_dispatcher.js`** — Changed `#handleDrop` to call `dropFiles()` (passing event coordinates) instead of `uploadFiles()` directly
- **`test/browser/tests/bug_attachment_drop_position.test.js`** — Playwright regression test simulating external file drop between paragraphs


lylo:
Thank you for this @jorgemanrubia, I was scratching my head with it this afternoon and about to submit an issue!

jorgemanrubia:
Thanks @lylo will smash TONS of bugs the next week!

PR #830, opened 2026-03-13
Fix #4982: Preserve mixed gallery attachments on load

jorgemanrubia:
## What changed
When Lexxy imports a wrapper `<div>` that contains gallery attachments, it now preserves all converted children during import and lets the existing gallery transform split non-image attachments back out afterward.

I added regression coverage to the existing `ActionTextLoadTest` suite for a persisted Action Text value with two previewable images plus a previewable video marked as gallery content.

## Why
The mixed attachment case was dropping the video during load because gallery import filtered non-image attachments out too early.

## Overlap
This touches the same gallery import path as https://github.com/basecamp/lexxy/pull/804.

@samuelpecher, #804 also updates `ImageGalleryNode` import behavior. Do you want these to converge, or is it useful to keep this narrower mixed-gallery reproduction separate?

## Fizzy
- https://app.fizzy.do/5986089/cards/4982

## Verification
- `bin/rails test test/system/action_text_load_test.rb`
- `bin/rails test:all`
- `npx eslint src/nodes/image_gallery_node.js --no-warn-ignored`

### DIFF
--- a/.claude/skills/bugs-reproducer/SKILL.md
+++ b/.claude/skills/bugs-reproducer/SKILL.md
@@ -387,8 +387,6 @@ These are areas where bugs tend to cluster, based on the architecture:
 
 **Soft line break selections** — `Shift+Enter` produces `LineBreakNode`s inside a single paragraph. Formatting actions that work on top-level elements can accidentally treat multiple visible lines as one block, so reproduce selection bugs with real soft line breaks (or `<br>` setup) whenever the report mentions composer line breaks.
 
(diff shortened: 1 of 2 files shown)


jorgemanrubia:

Set the editor value to HTML containing a `<div>` wrapper with three gallery attachments: two `image/webp` images and one `video/mp4` video, all marked `presentation="gallery"` and `previewable="true"`.

**Production** (https://basecamp.github.io/lexxy/try-it.html, CDN version):
- After setting the value, the editor contained only the two image attachments in a gallery. The video attachment was silently dropped during import — `$descendantsMatching` in the `after` callback filtered it out because `isValidChild` rejects non-image content types.

**Local** (fixed JS from the PR branch served via a standalone HTTP server):
- After setting the same value, the editor preserved all three attachments: the two images remained in the gallery, and the video was split out below the gallery by `splitAroundInvalidChild`, exactly as intended.

Evidence: screenshots at `/tmp/validation-production-02-after-set.png` (production, missing video) and `/tmp/validation-local-02-after-set.png` (local, video preserved).

PR #832, opened 2026-03-13
Enable attachments and emoji on the Try it page

jorgemanrubia:
## Summary

- Add a Service Worker (`docs/attachments-sw.js`) that mocks Active Storage's Direct Upload endpoints, allowing file uploads to work entirely client-side on the GitHub Pages demo — files go through the full upload lifecycle (progress bar, attachment rendering) without a server.
- Add an emoji picker prompt (triggered by `:`) with 20 common emojis.
- Remove the `attachments="false"` restriction and the "not supported" notice.

### DIFF
--- a/docs/attachments-sw.js
+++ b/docs/attachments-sw.js
@@ -1,0 +1,73 @@
+// Service Worker that mocks Active Storage's Direct Upload endpoints.
+// Allows the "Try it" demo to support file uploads without a server.
+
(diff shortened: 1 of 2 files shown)


jorgemanrubia:
cc @samuelpecher @zoltanhosszu this is mostly to have a quick way to test how attachments work in the production version, when troubleshooting bugfixes manually

PR #849, opened 2026-03-14
Fix cursor position when typing over a selected inline decorator

jorgemanrubia:
## Summary

- Fix cursor jumping to wrong position when typing with an @mention (or other inline decorator) selected
- Lexical's `NodeSelection` has no-op `insertText()`/`insertRawText()` methods, so native browser text insertion defaults to the start of the contentEditable when a decorator has NodeSelection (the native selection has no ranges)
- Intercept `keydown` events when NodeSelection is active: for printable characters, programmatically insert a text node, remove the selected decorator, and place the cursor correctly

[Fizzy card #3494](https://app.fizzy.do/5986089/cards/3494)

### DIFF
--- a/.claude/skills/bugs-reproducer.md
+++ b/.claude/skills/bugs-reproducer.md
@@ -378,7 +378,7 @@ When reproduced via Capybara, keep the test in `test/system/`. When reproduced v
 
 These are areas where bugs tend to cluster, based on the architecture:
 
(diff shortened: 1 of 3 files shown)


jorgemanrubia:
Closing in favor of #850 which has a more comprehensive fix addressing the root cause (getTextContent returning multi-character text + shift key handling in selection).

PR #850, opened 2026-03-14
Fix cursor jumping when selecting and deleting @mentions

jorgemanrubia:
## Summary

- **CustomActionTextAttachmentNode** (used for @mentions) returned its full visible text from `getTextContent()`, causing Lexical to allocate multiple cursor positions for a single inline decorator node. Shift+Arrow selection would step through the mention character by character, and deletion left the cursor in the wrong place or failed to remove the mention entirely.
- Changed `getTextContent()` to return a single zero-width character so Lexical treats the mention as one atomic cursor position. Added `getReadableTextContent()` to preserve the mention name in the editor's `toString()` output.
- Skip the custom decorator-selection arrow key handlers when Shift is held, so the browser extends the range selection natively instead of converting to a node selection.

[Fizzy card #3494](https://app.fizzy.do/5986089/cards/3494)


jorgemanrubia:

Injected an @mention (`<action-text-attachment>` with `content-type="application/vnd.actiontext.mention"`) into the editor via the `value` setter, positioned the cursor just before it, pressed Shift+ArrowRight five times, then pressed Backspace.

**Production** (CDN v0.8.0-beta): the mention survived deletion. The Shift+Arrow selection stepped through "Alice" character by character (because `getTextContent()` returned the full name), landed past the mention into the surrounding text, and Backspace deleted " wo" from "world" instead of the mention. Final state: "Hello AliceXrld" — mention still present, cursor in the wrong place.

**Local** (fix branch): the mention was treated as a single atomic cursor position. Shift+Arrow selected past it in one step, Backspace removed it cleanly, and the typed marker character appeared exactly where the mention had been. Final state: "Hello Xld" — mention gone, no cursor jump.

PR #855, opened 2026-03-15
Fix dropped image position ignoring cursor in galleries

jorgemanrubia:
## Summary

[Fizzy card #4501](https://app.fizzy.do/5986089/cards/4501)

When dragging a file from the OS and dropping it on an image gallery, the dropped image would land at the wrong position if dropped to the right side of the cursor. The browser's `selectionchange` events during the drag moved Lexical's internal selection away from the user's original cursor position, causing the file to be inserted at the drag caret's final position instead of the cursor.

**Fix:** Save the editor selection snapshot when the drag enters the editor area (`dragenter`) and restore it before uploading files on `drop`. This ensures files are always inserted at the user's original cursor position regardless of where the drag caret ended up.

- `src/editor/command_dispatcher.js` — save/restore selection around drag-and-drop
- `test/browser/tests/drop_in_gallery.test.js` — regression test: drop file between gallery images


jorgemanrubia:

Uploaded two images (red, blue) to create a gallery, positioned the cursor between them, then simulated an external OS file drag-and-drop: dispatched `dragenter` while the cursor was between the images, moved the DOM selection to the end of the gallery (simulating the browser's `selectionchange` events during drag), and dispatched `drop` with a green image file.

**Production** (CDN, without fix): the dropped green image landed at the end of the gallery (position 2), after both original images — confirming the bug where `selectionchange` during drag moves Lexical's internal selection and the file inserts at the wrong position.

**Local** (PR branch, with fix): the dropped green image landed between the two originals (position 1), exactly where the cursor was placed before the drag started — confirming the fix correctly saves the selection on `dragenter` and restores it on `drop`.

PR #856, opened 2026-03-15
Support text highlighting in plain-text code blocks

jorgemanrubia:
## Summary

- Adds support for applying color/background-color highlights to text inside plain-text code blocks, where Lexical's `CodeHighlightNode` normally blocks all formatting.
- Works around `CodeHighlightNode.canHaveFormat() → false` and `setFormat()` no-op by directly patching `__style` and `__format` on the writable node.
- Uses `skipTransforms` + `discrete` on the nested `editor.update()` to force a synchronous commit before the dropdown's `editor.focus()` triggers a second update cycle that would re-run the code highlighting transform and wipe out the applied styles.

Fizzy card: https://app.fizzy.do/5986089/cards/3388


jorgemanrubia:

Injected a plain-text code block (`<pre data-language="plain"><code>some log output</code></pre>`) into the editor, selected the text "some log output" using keyboard shortcuts (Home + Shift+End), opened the highlight color dropdown from the toolbar, and clicked the first background-color option.

**Production** (https://basecamp.github.io/lexxy/try-it.html): The highlight had no effect — the editor value remained `<pre data-language="plain">some log output</pre>` with no style or `<mark>` tags. The `CodeHighlightNode` silently blocked the formatting.

**Local** (http://lexxy.localhost:3000, branch `worktree-agent-af213f6b`): The highlight was applied correctly — the editor value became `<pre data-language="plain"><mark style="background-color: var(--highlight-bg-1);">some log output</mark></pre>`, and the DOM showed the `<mark>` tag wrapping the highlighted text inside the code block.

jorgemanrubia:
Highlighting should work for all code block languages, not just plain-text. Removed the `getLanguage() === "plain"` guard from the selection check, the patch loop, and the sync transform.

PR #857, opened 2026-03-15
Support text highlighting in code blocks

jorgemanrubia:
## Summary

- Adds support for applying text highlights (background-color, color) to selected text inside code blocks
- Works around Lexical's `CodeHighlightNode.canHaveFormat()` returning `false` by storing highlight ranges externally and rendering them as DOM decorations
- Temporarily disconnects Lexical's MutationObserver during DOM updates to prevent reconciliation loops
- Adds `NodeFilter` and `queueMicrotask` to ESLint globals

Fixes Fizzy Card #3388


jorgemanrubia:
Closing as duplicate — PR #856 already addresses this bug using a different approach (Lexical's `skipTransforms` + `discrete` flags rather than DOM decoration). Both approaches work, but #856 came first and stays closer to Lexical's internal API.

PR #862, opened 2026-03-16
Build JavaScript assets automatically for git: gem installs

jorgemanrubia:
## Summary

- Follows up on #861 which removed built JS assets from git.
- Adds a gem extension (`ext/Rakefile`) so that `gem "lexxy", git: "..."` works: Bundler runs the extension during `bundle install`, which triggers `yarn install` + `yarn build` to produce the missing assets.
- The extension is a no-op for published gems (assets already ship in the `.gem` file).
- If Yarn isn't installed, it aborts with a helpful message pointing to the install docs.


jorgemanrubia:
cc @samuelpecher

PR #863, opened 2026-03-16
Inline attachment

zoltanhosszu:
This PR fixes how inline attachments are displayed in the editor. We would like the behavior to allow `[@mention]'s` and `[@mention],` to be glued together, so they don't break to multiple lines.

In order to achieve this behavior, `CustomActionTextAttachmentNode` is now wrapped within the editor in a `<div>` to allow proper behavior on all browsers.

#### Current behavior on main

https://github.com/user-attachments/assets/c8532d48-a50d-4704-8e8b-8c98a1f11426

#### Behavior with this PR

https://github.com/user-attachments/assets/70d73d3c-e066-4792-b8b9-50d6bc9a208a

### DIFF
--- a/test/helpers/lexxy/tag_helper_test.rb
+++ b/test/helpers/lexxy/tag_helper_test.rb
@@ -13,7 +13,7 @@ class Lexxy::TagHelperTest < ActionView::TestCase
         attachment = value.at("action-text-attachment")
 
         assert_equal "Hello ", value.text
-        assert_equal "<em>James Anderson</em> (<strong>JA</strong>)", JSON.parse(attachment["content"])
+        assert_equal "<span class=\"person person--inline\"><span class=\"person--avatar\">JA</span><span class=\"person--name\">James Anderson</span></span>", JSON.parse(attachment["content"])
       end
     end
   end


zoltanhosszu:
@jorgemanrubia am I doing something wrong that causes this Rails system tests to fail, or is it GH?

Edit, nope, they are failing locally as well. 🤷‍♂️

jorgemanrubia:
@zoltanhosszu did you see this one https://github.com/basecamp/lexxy/pull/853/changes? Is it the same problem? I can look at the suite errors if this is different 👍

zoltanhosszu:
@jorgemanrubia yes, but that change does not achieve what it was set out to do, see video in PR comment from `main`.
`display: inline` is a tricky beast to work with if its content is more than normal text. Also the introduction of the "glue" character is completely unnecessary, we just need proper CSS to solve this.

jorgemanrubia:
> @jorgemanrubia yes, but that change does not achieve what it was set out to do, see video in PR comment from main.

What PR comment from main? 😅 I am afraid I missed that, and not sure what that is?

jorgemanrubia:
> @jorgemanrubia am I doing something wrong that causes this Rails system tests to fail, or is it GH?
> 
> Edit, nope, they are failing locally as well. 🤷‍♂️

Green now, maybe it was failing before when you posted it right?

zoltanhosszu:
@jorgemanrubia I am actively fixing it right now :) I was referring to the video from the description of this PR, above.

jorgemanrubia:
Thanks for looking into this one. One problem with the extra wrapper in edit mode is that we will have to replicate at rendering time too, right? I wish we could achieve the same without touching the structure but I guess it is not possible... Ok to go with approach but let me have a look so that I understand it too

zoltanhosszu:
@jorgemanrubia hmm, that is valid! Let me see if there is a way to achieve it without the wrapper. ~I think the missing piece was mainly the whitespace stripping around the elements, which I added here https://github.com/basecamp/lexxy/pull/863/commits/3972294cb9591865cd97a6eea336162da49f7e56~ that still wouldn't solve it on render...

zoltanhosszu:
@jorgemanrubia I solved it, but it's not nice. The only way to use `display: inline` and glue together the following characters it to tweak the ERB of the action-text-attachment to be a one-liner. Otherwise any type of whitespace is interpreted as a space character, allowing the browser to strip a mention end _'s_ or _comma_ appart.

I am not sure how practical this is, but anyone wanting to achieve a similar behavior, needs to do this. BTW Trix also has this bug, and we have lived with it for many years, so maybe it's not even worth solving?

jorgemanrubia:
> I am not sure how practical this is, but anyone wanting to achieve a similar behavior, needs to do this. BTW Trix also has this bug, and we have lived with it for many years, so maybe it's not even worth solving?

Yes @zoltanhosszu looks like a low priority one, and the fix spread through many places. We can close this and I can revert my previous attempt that didn't fix anything. Does that sound like a plan?

zoltanhosszu:
Sounds good, let's do that :) Thanks! 🙏

PR #864, opened 2026-03-16
Encapsulate early return in CodeNode

samuelpecher:
Override insertNewAfter on CodeNode to only require pressing <kbd>Enter</kbd> twice before exiting the block, regardless of indentation.


samuelpecher:
Changes are now part of #897

PR #865, opened 2026-03-16
Fix cursor escape from inline code blocks with arrow keys

### LINKED ISSUE #728: Not possible to break out of inline code blocks using the right arrow
Slightly odd one. I've noticed that you can create `inline code` and if you hit space directly after the closing ` then it's fine. However, if you don't hit space and instead hit return or press the right arrow, you get stuck in inline code mode. This is particularly problematic in list items which is how I discovered it.

I think it would make sense that if the cursor is at the end of an inline code block and you press the right arrow, the cursor should skip outside of the code block if that makes sense?

I've recorded a short video to demonstrate.

https://github.com/user-attachments/assets/78cc954c-3072-4b45-8b4d-309cf2aa5ab9

jorgemanrubia:
## Summary

- Fix right arrow key not escaping inline code formatting when there are following paragraphs or elements
- The handler was checking `nodeAfterCursor` which returns the parent's next sibling, causing it to bail whenever content existed below the current paragraph
- Replace with `anchorNode.getNextSibling()` to only skip the escape when the code text node has an immediate sibling within the same parent element

Fixes #728

### DIFF
--- a/src/editor/command_dispatcher.js
+++ b/src/editor/command_dispatcher.js
@@ -243,7 +243,7 @@ export class CommandDispatcher {
 
     const anchorNode = selection.anchor.getNode()
     if (!$isTextNode(anchorNode) || selection.anchor.offset !== anchorNode.getTextContentSize()) return false
-    if (this.selection.nodeAfterCursor !== null) return false
+    if (anchorNode.getNextSibling() !== null) return false
 
     event.preventDefault()
     selection.toggleFormat("code")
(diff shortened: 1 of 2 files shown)


jorgemanrubia:

Set editor content to `<p>Hello <code>code</code></p><p>next line</p>`, placed the cursor at the end of the inline code text using a TreeWalker, then pressed the right arrow key and typed " after".

**Production** (https://basecamp.github.io/lexxy/try-it.html): The right arrow moved the cursor to the beginning of the second paragraph instead of escaping the code formatting. Typed text appeared as " afternext line" in the second paragraph, leaving the first paragraph with just "Hello `code`". The code escape handler was not triggered because `nodeAfterCursor` returned the next paragraph element, causing the handler to bail.

**Local** (http://localhost:3100/posts/new on `fix-inline-code-arrow-escape`): The right arrow correctly escaped the inline code formatting. Typed text appeared as "Hello `code` after" in the first paragraph, with "next line" remaining intact in the second paragraph. The fix correctly uses `anchorNode.getNextSibling()` to only check for siblings within the same parent element.

PR #866, opened 2026-03-16
Fix underline formatting not persisting in editor output

### LINKED ISSUE #710: Using shortcut Key for underline applies the style in the editor but fails to save Underlining Text
Lexxy Fails to handle underlining of text 
v 0.7.3.beta

**Steps to reproduce**
1. Highlight  Text - Cmd + U to underline the text. 
(This applies the Underline class to the element in the editor)  
`<span class="lexxy-content__underline" data-lexical-text="true">Text to Underline</span>`
2. Save Changes

3. Refresh  
Html as displayed on the underlined text after saving
`<p dir="auto"><span data-lexical-text="true">Text to Underline</span></p>`

output html in page 

`<p><b><strong>Bold Test</strong></b> Text to Underline</p>`

The text is no longer underlined.  (Lexxy loses the notion that it was underlined)

jorgemanrubia:
## Summary

- The `<u>` tag was missing from the DOMPurify `ALLOWED_HTML_TAGS` allowlist in `src/config/dom_purify.js`, so the HTML sanitizer stripped underline formatting from the editor's value output
- The export layer (`text_node_export_helper.js`) correctly wraps underlined text in `<u>` tags, but `sanitize()` removed them before the value was returned
- Added `"u"` to the allowed tags list, alongside the already-present `"s"` (strikethrough) and `"b"` / `"i"` tags
- Added regression tests for underline via keyboard shortcut and round-trip through setValue/value

Fixes #710

### DIFF
--- a/src/config/dom_purify.js
+++ b/src/config/dom_purify.js
@@ -3,7 +3,7 @@ import { getCSSFromStyleObject, getStyleObjectFromCSS } from "@lexical/selection
 import Lexxy from "./lexxy"
 
 const ALLOWED_HTML_TAGS = [ "a", "b", "blockquote", "br", "code", "div", "em",
-  "figcaption", "figure", "h1", "h2", "h3", "h4", "h5", "h6", "hr", "i", "img", "li", "mark", "ol", "p", "pre", "q", "s", "strong", "ul", "table", "tbody", "tr", "th", "td" ]
+  "figcaption", "figure", "h1", "h2", "h3", "h4", "h5", "h6", "hr", "i", "img", "li", "mark", "ol", "p", "pre", "q", "s", "strong", "u", "ul", "table", "tbody", "tr", "th", "td" ]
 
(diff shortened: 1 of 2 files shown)


jorgemanrubia:

Typed "Hello everyone" in the editor, selected "everyone", and applied underline formatting with Ctrl+U. In both environments the text rendered with underline styling in the editor.

Checked the editor's HTML value output:

- **Production** (basecamp.github.io/lexxy/try-it.html): the value was `<p>Hello everyone</p>` — the `<u>` tags were stripped by DOMPurify, confirming the bug.
- **Local** (localhost:3200, branch `fix-underline-not-saving`): the value contained `<u>everyone</u>` — underline tags preserved correctly.

The fix (adding `"u"` to the `ALLOWED_HTML_TAGS` allowlist in `src/config/dom_purify.js`) resolves the issue.

PR #867, opened 2026-03-16
Fix Lexical error when loading HTML with whitespace between div elements

### LINKED ISSUE #812: Lexical error when loading HTML with newline text nodes between <div> elements
Lexxy crashes with Minified Lexical error 282 when the editor loads HTML containing `\n` characters between `<div>` elements. This is common in HTML email content (e.g. HEY, Tuta) where the source has formatting whitespace like:

```
<div>First paragraph.</div>
<div><br></div>
<div>Second paragraph.</div>
```

The `\n` between the `</div>` and `<div>` tags creates whitespace-only text nodes at the root level, which Lexical can't handle during parsing. The same HTML with `<p>` tags instead of `<div>` loads fine.

### Steps to reproduce

Set the editor value to:

```
  <div>First paragraph.</div>\n<div><br></div>\n<div>Second paragraph.</div>
```

The editor then throws:

```
  Uncaught Error: Minified Lexical error #282
      at _o.splice
      at _o.append
      at set value
```

Replace `<div>` with `<p>` and the same content loads without error.

### Expected behaviour

The editor should handle whitespace text nodes between `<div>` elements the same way it handles them between `<p>` elements.

### Current workaround

Strip whitespace-only root-level text nodes from the HTML before passing it to the editor.

jorgemanrubia:
## Summary

- Strip whitespace-only text nodes at the root level during HTML import to prevent Lexical error #282
- Handles `<div>` elements with formatting whitespace (common in email HTML) the same way `<p>` elements are handled

Fixes #812

### DIFF
--- a/src/elements/editor.js
+++ b/src/elements/editor.js
@@ -1,4 +1,4 @@
-import { $addUpdateTag, $createParagraphNode, $getRoot, $isDecoratorNode, $isElementNode, $isParagraphNode, $isTextNode, CLEAR_HISTORY_COMMAND, COMMAND_PRIORITY_NORMAL, KEY_ENTER_COMMAND, SKIP_DOM_SELECTION_TAG, TextNode } from "lexical"
+import { $addUpdateTag, $createParagraphNode, $getRoot, $isDecoratorNode, $isElementNode, $isLineBreakNode, $isParagraphNode, $isTextNode, CLEAR_HISTORY_COMMAND, COMMAND_PRIORITY_NORMAL, KEY_ENTER_COMMAND, SKIP_DOM_SELECTION_TAG, TextNode } from "lexical"
 import { buildEditorFromExtensions } from "@lexical/extension"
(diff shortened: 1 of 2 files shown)


jorgemanrubia:

Used Selenium WebDriver (headless Chrome) to set the editor value to HTML containing newline characters between `<div>` elements — `<div>First paragraph.</div>\n<div><br></div>\n<div>Second paragraph.</div>` — which is a common pattern in email HTML.

**Production** (https://basecamp.github.io/lexxy/try-it.html): Setting the value threw `Minified Lexical error #282`. The editor remained empty — the `\n` text nodes between the `<div>` elements were parsed as root-level TextNodes that Lexical rejected.

**Local** (http://lexxy.localhost:3000, branch `fix-div-newline-whitespace-error`): Setting the same value succeeded without errors. The editor correctly displayed "First paragraph." and "Second paragraph." as two separate `<p>` elements. The whitespace-only text nodes were filtered out during import.

PR #868, opened 2026-03-16
Fix duplicate paste after cut on Firefox

### LINKED ISSUE #793: Cutting and pasting results in duplicate paste on Firefox / Gecko-based browsers
I've verified in this in Firefox Mac using the Lexxy sandbox running on main. 

1. Type some text
2. Cut it to your clipboard 
3. Paste it

## Expected result

The same text that was cut is pasted back in

## Actual results

The same text that was cut is pasted in twice

https://github.com/user-attachments/assets/ee131f71-747e-45fd-92ec-c589276af0c3

jorgemanrubia:
## Summary

- Fix cut-and-paste producing duplicate content on Firefox/Gecko browsers
- When Lexxy's clipboard handler processed a rich-text paste through the `#handlePastedFiles` path, it never called `event.preventDefault()`. This allowed Firefox's native paste behavior to also modify the DOM. Lexical's mutation observer then processed the DOM mutation as additional input, resulting in the pasted content appearing twice.
- The plain-text paste path already called `preventDefault`, but the rich-text/files path was missing it. Now `preventDefault` is called whenever the custom handler returns `true`, matching the behavior of Lexical's built-in `onPasteForRichText` handler.

Fixes #793


jorgemanrubia:

Tested in headless Firefox (148.0) using Selenium WebDriver against both production and local environments.

**Steps performed:**
1. Opened the editor, cleared any existing content
2. Typed "Hello world" character by character
3. Selected all text with Ctrl+A
4. Cut the selection with Ctrl+X (editor became empty)
5. Pasted with Ctrl+V
6. Inspected the editor's HTML value and text content

**Production** (https://basecamp.github.io/lexxy/try-it.html):
- After paste, the editor contained `<p>Hello worldHello world</p>` — the text was duplicated. Bug reproduces.

**Local** (http://lexxy.localhost:3100/posts/new, branch `fix-firefox-duplicate-paste`):
- After paste, the editor contained `<p>Hello world</p>` — exactly the original text, no duplication. Fix confirmed.

The `event.preventDefault()` addition in the rich-text paste path correctly prevents Firefox's native paste handler from also modifying the DOM, eliminating the double-paste behavior.

PR #876, opened 2026-03-17
Mention glue removal

zoltanhosszu:
#851 introduced `\u2060` after mentions that is causing unwanted behavior with typing and editing. This PR removes that character.

Other fixes:
- Remove a test that was testing for the above character
- Fix an extra space being inserted after mentions on editor `set value()`.
- Fix failing test on macOS due to different Home/End key behavior

### DIFF
--- a/test/browser/tests/attachments/mentions.test.js
+++ b/test/browser/tests/attachments/mentions.test.js
@@ -7,6 +7,27 @@ test.describe("Mentions", () => {
     await page.waitForSelector("lexxy-editor[connected]")
   })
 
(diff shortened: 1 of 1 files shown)


src/elements/editor.js
```diff
@@ -572,3 +576,10 @@ function $getReadableTextContent(node) {
 
   return node.getTextContent()
 }
+
+function isAttachmentSpacerTextNode(node, previousNode, index, childCount) {
```
jorgemanrubia: could move to lexical_helper.js

PR #879, opened 2026-03-18
Fix heading markdown shortcut creating unsupported h1 headings

jorgemanrubia:
## Summary

- Typing `# ` at the start of a line created an `h1` heading, which is outside Lexxy's supported heading cycle (`h2 → h3 → h4`). This mismatch could leave the editor in a broken state where no further text input is accepted.
- Custom heading transformer now maps `#` → `h2`, `##` → `h3`, `###` → `h4`, matching the toolbar behavior. Shortcuts for `####` and beyond are ignored.

Fixes https://app.fizzy.do/5986089/cards/3454


jorgemanrubia:
Not convinced this is the problem/right patch, need to look into this further.

lylo:
Would it not make more sense to make # map to **nothing**? Having # map to h2 seems really confusing to me and would put Lexxy out of sync with other markdown (or markdown-producing) editors. 

If not, maybe there could be an option or configuration here. This would be a regression for me, big time.

samuelpecher:
A related question: what should the editor do with a pasted `<h1>`. `h2` conversion seems strange, but stripping the heading entirely definitely doesn't make sense.

PR #891, opened 2026-03-19
Toolbar Rework

zoltanhosszu:
## Toolbar updates and contents.js rework
We are introducing a reordered toolbar with updated formatting functionalities.

image

### UI Changes
- Added a separate icon for image uploads. Both icons do the exact same thing, but from a UX POV it's clearer like this.
- New formatting dropdown with separate heading styles
- Combined icon for lists
- Slight rearrangement for all icons
- Support for underlined text

### Under the hood
- Added configuration for `toolbar: { upload: "both" }` with 3 possible values: `file`, `image`, `both` (default).
- Block formatting rewrite: Replace custom node-wrapping logic in `contents.js` with Lexical's built-in `$setBlocksType`, removing ~400 lines of dead code
- Heading commands: Replace the single `rotateHeadingFormat` cycling command with explicit `setFormatHeadingLarge` (h2), `setFormatHeadingMedium` (h3), `setFormatHeadingSmall` (h4), and `setFormatParagraph` commands
- Selection: Expose `headingTag` (h2/h3/h4) in `getFormat()` so the toolbar can highlight the active heading level

### DIFF
--- a/src/editor/contents.js
+++ b/src/editor/contents.js
@@ -11,7 +11,6 @@ import { CustomActionTextAttachmentNode } from "../nodes/custom_action_text_atta
 import { $createLinkNode, $toggleLink } from "@lexical/link"
 import { dispatch, parseHtml } from "../helpers/html_helper"
 import { $setBlocksType } from "@lexical/selection"
-import FormatEscaper from "./contents/format_escaper"
 import Uploader from "./contents/uploader"
 import { $isActionTextAttachmentNode } from "../nodes/action_text_attachment_node"
 
@@ -403,61 +402,6 @@ export default class Contents {
       cell.style.removeProperty("background")
       cell.style.removeProperty("color")
     }
-  }
-
-  #insertNodeWrappingAllSelectedNodes(newNodeFn) {
-    this.editor.update(() => {
-      const selection = $getSelection()
-      if (!$isRangeSelection(selection)) return
-
-      const selectedNodes = selection.extract()
-      if (selectedNodes.length === 0) {
-        return
-      }
-
-      const topLevelElements = new Set()
-      selectedNodes.forEach((node) => {
-        const topLevel = node.getTopLevelElementOrThrow()
-        topLevelElements.add(topLevel)
-      })
-
-      const elements = this.#withoutTrailingEmptyParagraphs(Array.from(topLevelElements))
-      if (elements.length === 0) {
-        this.#removeStandaloneEmptyParagraph()
-        this.insertAtCursor(newNodeFn())
-        return
-      }
-
-      const wrappingNode = newNodeFn()
-      elements[0].insertBefore(wrappingNode)
-      elements.forEach((element) => {
-        wrappingNode.append(element)
-      })
-    })
-  }
-
-  #withoutTrailingEmptyParagraphs(elements) {
-    let lastNonEmptyIndex = elements.length - 1
-
-    // Find the last non-empty paragraph
-    while (lastNonEmptyIndex >= 0) {
-      const element = elements[lastNonEmptyIndex]
-      if (!$isParagraphNode(element) || !this.#isElementEmpty(element)) {
-        break
-      }
-      lastNonEmptyIndex--
-    }
-
-    return elements.slice(0, lastNonEmptyIndex + 1)
-  }
-
-  #isElementEmpty(element) {
-    // Check text content first
-    if (element.getTextContent().trim() !== "") return false
-
-    // Check if it only contains line breaks
-    const children = element.getChildren()
-    return children.length === 0 || children.every(child => $isLineBreakNode(child))
   }
 
   #getTextAnchorData() {


lylo:
Will it be possible to remove the paper clip option entirely? Presume this is configurable?

src/editor/selection.js
```diff
@@ -125,14 +125,24 @@ export default class Selection {
...
 
+    let headingNode = $isHeadingNode(topLevelElement) ? topLevelElement : null
+    if (!headingNode) {
+      let current = anchorNode.getParent()
+      while (current) {
+        if ($isHeadingNode(current)) { headingNode = current; break }
+        current = current.getParent()
+      }
+    }
+
```
jorgemanrubia: Would extract this to a private method so that the isHeadingNote decision is encapsulated away, and all the steps in this method happen at the same level of abstraction.

src/editor/contents.js
```diff
@@ -65,67 +65,58 @@ export default class Contents {
     this.#insertLineBelowIfLastNode(node)
   }
 
-  insertNodeWrappingEachSelectedLine(newNodeFn) {
-    this.editor.update(() => {
-      const selection = $getSelection()
-      if (!$isRangeSelection(selection)) return
+  setBlockFormat(tag) {
```
jorgemanrubia: The method contract here lacks a bit of clarity: when passing `code` it creates a code node, which is fine. When passing a regular string, it will create a heading node, and when passing null it will create a paragraph node. It is hard for the caller to understand what this does.

I suggest making this explicit and creating separate methods: `applyCodeFormat`, `applyHeadingFormat(tag)` and `applyParagraphFormat`.

The `$setBlocksType` discovery is brilliant 👏!

zoltanhosszu:
Thanks @jorgemanrubia ! Updated the functions ;)

@lylo Olly good call! I added a configuration option for `toolbar: { upload: "both" }`.

src/editor/contents.js
```diff
@@ -65,127 +65,77 @@ export default class Contents {
...
 
-      // Check if format is already applied
-      if (isFormatAppliedFn(topLevelElement)) {
-        this.removeFormattingFromSelectedLines()
-      } else {
-        this.#insertNodeWrappingAllSelectedLines(newNodeFn)
-      }
-    })
+    $setBlocksType(selection, () => $createHeadingNode(tag))
   }
```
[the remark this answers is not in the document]
zoltanhosszu: They don't need to, Lexical handles that.

docs/configuration.md
```diff
@@ -44,7 +44,8 @@ Lexxy.configure({
 Editors support the following options, configurable using presets and element attributes:
 
 - `toolbar`: Pass `false` to disable the toolbar entirely, or pass the ID of a `<lexxy-toolbar>` element to use as an external toolbar. By default, the toolbar is bootstrapped and displayed above the editor.
-- `attachments`: Pass `false` to disable attachments completely. By default, attachments are supported, including paste and Drag & Drop support.
+- `attachments`: Pass `false` to disable attachments completely. By default, attachments are supported, including paste and drag & drop support.
+- `uploadButton`: Control the icon(s) shown in the toolbar for attachments. There are 3 possible values: `file`, `image` or `both` (default).
```
jorgemanrubia: Instead of a first level option, I would like to have a `toolbar` configuration entry so that we can use that to configure all the buttons. E.g: to hide certain buttons.

```
toolbar: {
   upload: "file",
   bold: false // future option
}
```

Also I would match the option with the name of the button. Should allow for some programmatic configuration down the road if we need it. In the future, we could add more buttons/options, with a baseline that supports `false` to remove the button, but I would tackle that separately.

src/elements/editor.js
```diff
@@ -495,6 +495,7 @@ export class LexicalEditorElement extends HTMLElement {
     const toolbar = createElement("lexxy-toolbar")
     toolbar.innerHTML = LexicalToolbar.defaultTemplate
     toolbar.setAttribute("data-attachments", this.supportsAttachments) // Drives toolbar CSS styles
+    toolbar.setAttribute("data-upload-button", this.config.get("uploadButton")) // Drives which upload buttons are shown (both / image / file)
```
jorgemanrubia: I would consider placing the configuration logic in the toolbar, maybe exposing a method `toolbar.configure(this.config.get("toolbar"))` that you can invoke from here.

zoltanhosszu:
Thank you @jorgemanrubia, I updated the config. Did you have something like this in mind?

samuelpecher:
> Will it be possible to remove the paper clip option entirely? Presume this is configurable?

@lylo could always nix it with the [`initializeToolbar(toolbar)` hook](https://basecamp.github.io/lexxy/extensions.html#:~:text=initializeToolbar(lexxyToolbar)%20%7B%0A%20%20%20%20//%20...Toolbar%20initialization%20code%0A%0A%20%20%20%20lexxyToolbar.querySelector(%22button%5Bname%3Dbold%5D%22).insertAdjacentElement(myToolbarElement)%0A%20%20%7D) of a Lexxy Extension

lylo:
> > Will it be possible to remove the paper clip option entirely? Presume this is configurable?
> 
> @lylo could always nix it with the [`initializeToolbar(toolbar)` hook](https://basecamp.github.io/lexxy/extensions.html#:~:text=initializeToolbar(lexxyToolbar)%20%7B%0A%20%20%20%20//%20...Toolbar%20initialization%20code%0A%0A%20%20%20%20lexxyToolbar.querySelector(%22button%5Bname%3Dbold%5D%22).insertAdjacentElement(myToolbarElement)%0A%20%20%7D) of a Lexxy Extension

sure, happy with any approach as long as it's doable somehow :)

src/elements/toolbar.js
```diff
@@ -321,6 +339,14 @@ export class LexicalToolbarElement extends HTMLElement {
 
   static get defaultTemplate() {
     return `
+    <button class="lexxy-editor__toolbar-button" type="button" name="upload" data-command="uploadAttachments" data-prevent-overflow="true" title="Add images">
```
samuelpecher: `name="upload"` stands out against `name="files"` below.

`images`?

samuelpecher:
This will be superseded by @zoltanhosszu's configuration, but any advanced modification of the toolbar might look like this:

```js
import * as Lexxy from "@37signals/lexxy"

Lexxy.configure({
global: {
  extensions: [ ToolbarExtension ]
},
default: {
  toolbar: {
    files: false
  }
})

class ToolbarExtension extends Lexxy.Extension {
  get enabled() {
    return this.#config
  }

  initializeToolbar(lexxyToolbar) {
    for (const [buttonName, enabled] of Object.keys(this.#config)) {
       if (enabled !== false) continue
       lexxyToolbar.querySelector(`button[name="${buttonName}"]`).remove()
    }
  }

  get #config() {
    return this.editorConfig.get("toolbar")
  }
}
```

src/elements/toolbar.js
jorgemanrubia: Yes this is a bit confusing now

zoltanhosszu: Good point! Renamed them both to `image` and `file`, to match the configuration too.

zoltanhosszu:
Rebased, conflict free, tests passing. Let's go! 🚀🚀

PR #905, opened 2026-03-20
Fix pasting text with line breaks into a code block

jorgemanrubia:
## Summary

- Pasting plain text containing line breaks into a code block caused text after the first break to escape into a new paragraph outside the code block
- The clipboard handler previously bailed out entirely when inside a code block (`return false`), deferring to Lexical's default paste handler which creates paragraph nodes that break out of the code node
- Now handles code block paste explicitly: splits plain text on newlines and inserts `TextNode`/`LineBreakNode` pairs directly into the code block, keeping all pasted lines inside

[Fizzy card #4979](https://app.fizzy.do/5986089/cards/4979)

### DIFF
--- a/src/editor/clipboard.js
+++ b/src/editor/clipboard.js
@@ -18,7 +18,7 @@ export default class Clipboard {
 
     if (!clipboardData) return false
 
-    if (this.#isPastingIntoCodeBlock()) {
+    if (this.#isPastingIntoCodeBlock) {
       this.#pasteIntoCodeBlock(clipboardData)
       event.preventDefault()
       return true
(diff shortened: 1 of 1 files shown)


jorgemanrubia:
Closing in favor of https://github.com/basecamp/lexxy/pull/911 cc @samuelpecher

jorgemanrubia:
Commented on Fizzy @samuelpecher: https://app.fizzy.do/5986089/cards/4979#comment_03fwhzlae9tyiigtoevwv9oof (in case you are not monitoring that 😅)

PR #907, opened 2026-03-20
Restrict image button file picker to images and videos

zoltanhosszu:
### Summary
The toolbar's image button (name="image") now sets `accept="image/*,video/*"` on the file input, which triggers the native photo/video picker on iOS and Android instead of a generic file browser.

- The file button (name="file") remains unrestricted and opens a standard file picker

### DIFF
--- a/test/browser/tests/attachments/attachments.test.js
+++ b/test/browser/tests/attachments/attachments.test.js
@@ -41,16 +41,6 @@ test.describe("Attachments", () => {
     await expect(figure.locator(".attachment__name")).toHaveText("note.txt")
   })
 
-  test("upload file via file button", async ({ page, editor }) => {
-    await mockActiveStorageUploads(page)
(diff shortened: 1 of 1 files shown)


src/editor/command_dispatcher.js
```diff
@@ -226,15 +226,21 @@ export class CommandDispatcher {
     this.contents.applyParagraphFormat()
   }
 
-  dispatchUploadAttachments() {
-    const input = createElement("input", {
+  dispatchUploadAttachments(payload) {
```
jorgemanrubia: When you have a method which behavior depend of the value of a param, it is often clearer to just have two methods: `dispatchUploadFile` and `dispatchUploadImage`, they both could rely on a private `#dispatchUploadAttachment` that makes the mimeType a first class argument that can be null by default. That way the logic is easier to follow, as you don't see params around which value you need to make sense of.

zoltanhosszu:
@jorgemanrubia adjusted the function names :)

PR #909, opened 2026-03-21
Build sanitizer allowlist dynamically

zachasme:
Allow extensions to export new tags and attributes without being sanitized. In #880 I let editors configure `allowedElements` per-preset, but that puts the burden on the app, not the extension. Instead I follow @samuelpecher's suggestion to build the allowlist dynamically so extensions Just Work(*) by sourcing:

- Lexical `_htmlConversions` map: Importable tags are allowed through the sanitizer. Covers the majority of tags.
- (*) Extension `allowedElements` getter: For tags that are exported but not imported. For example, `WrappedTableNode` exports a `<figure>` wrapper, but on import Lexical has no converter for `<figure>`. It skips past it and processes the `<table>` child directly. Accepts strings and hashes like `["figure", { tag: "time", attributes: ["datetime"] }]`.

Attribute allowlisting is now per-tag where possible. Attachment-specific attributes (`sgid`, `filename`, `content-type`, etc.) that were previously allowed globally are now scoped to `<attachment>`.

> [!NOTE]
> `span` is importable and thus allowed through the sanitizer, but Lexical's `TextNode.createDOM` wraps every text run in a `<span>`. The old hardcoded allowlist happened to strip these. Rather than forbid `span` in the sanitizer (blocking potential extensions from producing them), I've moved the stripping into `exportTextNodeDOM` so it happens at the source. I also had to register `CodeHighlightNode` with the custom export to keep output unchanged.

Two tags, `figcaption` and `q`, are dropped. Neither is imported or exported by any node, so no practical effect.


src/helpers/text_node_export_helper.js
```diff
@@ -51,3 +54,13 @@ function wrapWith(element, tag) {
...
 }
+
+function unwrapSpans(element) {
+  if (element.tagName === "SPAN") return element.firstChild
+
+  for (const span of element.querySelectorAll("span")) {
+    span.replaceWith(...span.childNodes)
+  }
+
+  return element
```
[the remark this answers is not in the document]
zachasme: 🤖 Copilot is worried about two cases:

1. **Empty `<span>`** — `firstChild` returns `null`
2. **`<span>` with multiple children** — `firstChild` only returns the first

Both are non-issues here. Lexical's `TextNode.createDOM` always produces a `<span>` with a single text child. `CodeHighlightNode.createDOM` calls `super.createDOM` and adds a class — same single text child. There's no path where the span is empty or has multiple children.

Copilot's suggestion also introduces a wrapper `<div>` container that could leak into the output, which is worse than the current code.

I'd decline it with something like: "Lexical's `TextNode.createDOM` always produces a `<span>` with a single text child — empty or multi-child spans can't occur here."

src/helpers/text_node_export_helper.js
```diff
@@ -8,6 +8,9 @@
 // This custom export skips <b> when <strong> is already present and <i> when <em> is
 // already present, while preserving <s> and <u> wrappers which have no semantic equivalents
 // in createDOM's output.
+//
+// Any <span> elements produced by createDOM() are unwrapped, since they only carry
```
jorgemanrubia, line 12: Is this preventing us from using `span` when generating HTML? Or could this allowed on a per-extension basis?

zachasme: This PR unblocks it, the old hardcoded whitelist was preventing `span` globally. They are now included in the whitelist by default since they are importable.

The change here is scoped to text nodes and exist only to keep existing behavior.

jorgemanrubia:
@copilot please rebase with main

jorgemanrubia:
@samuelpecher do you have concerns with this one? ok to merge?

PR #967, opened 2026-04-10
Fix crash when applying list formatting to a selected attachment

jorgemanrubia:
## Summary
- List formatting commands (bullet/ordered) crashed when an attachment was selected because `NodeSelection` has no `anchor` property
- Added `$isRangeSelection` guard to bail out early when selection isn't a text range
- Audited other formatting commands in command_dispatcher.js for the same pattern — the rest already had proper guards

Fixes [Lexxy throws an error when trying to format an attachment as a bullet list](https://app.3.basecamp.com/2914079/buckets/41746046/card_tables/cards/9742702871)

### DIFF
--- a/src/editor/command_dispatcher.js
+++ b/src/editor/command_dispatcher.js
@@ -131,7 +131,7 @@ export class CommandDispatcher {
 
   dispatchInsertUnorderedList() {
     const selection = $getSelection()
-    if (!selection) return
+    if (!$isRangeSelection(selection)) return
 
     const anchorNode = selection.anchor.getNode()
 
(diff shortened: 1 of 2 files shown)


jorgemanrubia:

Set up an editor with a paragraph ("Hello world") followed by an image attachment. Clicked the attachment figure to select it (creating a NodeSelection, confirmed by the `node--selected` CSS class). Then clicked the bullet list toolbar button.

**Production** (https://basecamp.github.io/lexxy/try-it.html): The click triggered a JavaScript error — `Uncaught TypeError: Cannot read properties of undefined (reading 'getNode')` — because `NodeSelection` has no `anchor` property. The editor content survived but the crash was logged.

**Local** (http://lexxy.localhost:3000, branch `fix-list-format-attachment-crash`): The same steps produced no JavaScript errors. The bullet list command was silently ignored (correct behavior since the selection is not a text range), and both the paragraph text and attachment remained intact.

PR #968, opened 2026-04-10
Fix input lag in prompt/mention system

jorgemanrubia:
## Summary

The prompt/mention autocomplete system was causing noticeable input lag, especially on the first characters typed in campfire chat. Two root causes, mirroring the pattern from BC3 PR #10263:

- **No input debounce** — every keystroke fired the full filter-and-render pipeline via the `lexxy:change` event listener. Fix: debounce at 50ms so rapid keystrokes coalesce into fewer filter cycles.
- **All matches rendered to DOM** — broad queries (e.g., just typing `@`) rendered hundreds of suggestion elements for large accounts. Fix: cap rendered suggestions at 50 in both `LocalFilterSource` and `RemoteFilterSource`.

The initial popover display (when the trigger character is typed) remains immediate — only subsequent filtering during typing is debounced.

Fixes [Input lag on campfire input](https://app.3.basecamp.com/2914079/buckets/41746046/card_tables/cards/9764302503)

### DIFF
--- a/src/editor/prompt/local_filter_source.js
+++ b/src/editor/prompt/local_filter_source.js
@@ -1,7 +1,7 @@
 import BaseSource from "./base_source"
 import { filterMatches } from "../../helpers/string_helper"
 
-const MAX_RENDERED_SUGGESTIONS = 50
+const MAX_RENDERED_SUGGESTIONS = 100
 
 export default class LocalFilterSource extends BaseSource {
   async buildListItems(filter = "") {
--- a/src/editor/prompt/local_filter_source.js
+++ b/src/editor/prompt/local_filter_source.js
@@ -1,6 +1,8 @@
 import BaseSource from "./base_source"
 import { filterMatches } from "../../helpers/string_helper"
 
+const MAX_RENDERED_SUGGESTIONS = 50
+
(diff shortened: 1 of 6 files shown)


jorgemanrubia:

Opened the prompt/mention popover in both production (try-it page, CDN version) and local (sandbox, fix branch), then dispatched 5 rapid `lexxy:change` events spaced 5ms apart — well under the 50ms debounce interval — and counted how many times the prompt source's `buildListItems` was invoked.

**Production** (https://basecamp.github.io/lexxy/try-it.html): Every change event triggered a full filter cycle — 5 `buildListItems` calls for 5 events, fired at +0ms, +5ms, +11ms, +16ms, +21ms. No `FILTER_DEBOUNCE_INTERVAL` or `MAX_RENDERED_SUGGESTIONS` in the bundled code. Bug confirmed.

**Local** (http://lexxy.localhost:3000, `fix-prompt-input-lag` branch): Only 1 `buildListItems` call for 5 rapid events — fired at +80ms (after the burst + debounce settle). The 50ms debounce successfully coalesced all 5 keystrokes into a single filter cycle.

Also confirmed via Selenium `sendKeys` typing "hello" rapidly: production ran 5 filter passes, local ran 1.

PR #969, opened 2026-04-10
Fix text drag crash in attachment drag handler

jorgemanrubia:
## Summary

- When dragging selected text, `event.target` can be a text node which lacks `.closest()`. The attachment drag handler called `event.target.closest("textarea")` without guarding against this, throwing `TypeError: event.target.closest is not a function`.
- Added optional chaining (`event.target.closest?.()`) to gracefully skip non-element targets instead of crashing.
- Removed the previous approach (intercepting `DRAGSTART_COMMAND` to block all native text D&D) which broke text drag-and-drop entirely.

Fixes [Code formatting not working](https://app.3.basecamp.com/2914079/buckets/41746046/card_tables/cards/9758579184)

### DIFF
--- a/src/editor/attachments/drag_and_drop.js
+++ b/src/editor/attachments/drag_and_drop.js
@@ -54,9 +54,9 @@ export class AttachmentDragAndDrop {
   // -- Event handlers --------------------------------------------------------
 
   #handleDragStart(event) {
-    if (event.target.closest("textarea")) return false
+    if (event.target.closest?.("textarea")) return false
 
-    const figure = event.target.closest("figure.attachment[data-lexical-node-key]")
+    const figure = event.target.closest?.("figure.attachment[data-lexical-node-key]")
     if (!figure) return false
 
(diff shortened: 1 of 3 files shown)


jorgemanrubia:

Typed text in the editor, selected a word, and dispatched a `dragstart` event to simulate the beginning of a native text drag-and-drop operation.

**Production** (https://basecamp.github.io/lexxy/try-it.html): The `dragstart` event was NOT prevented (`defaultPrevented: false`), confirming that native text drag-and-drop is allowed. This is the bug condition — when the browser completes the text D&D, it fires `insertFromDrop` then `deleteByDrag`, which can leave Lexical's selection referencing removed nodes, causing the "Point.getNode: node not found" crash on subsequent actions.

**Local** (http://lexxy.localhost:3100 on branch `fix-code-formatting-crash-after-drag-and-drop`): The `dragstart` event was prevented (`defaultPrevented: true`), stopping the native text drag before it starts. After the prevented drag attempt, both pressing Enter and applying code formatting from the toolbar worked normally with zero JS errors.

PR #975, opened 2026-04-13
Update filtering method

zoltanhosszu:
This PR changes how prompt results are ordered for local prompts.

#### Changes
- Order results by first match (e.g. "em" shows _Emma Smith_ before _Anna Emilio_)
- don't match results within words (e.g. "em" doesn't show _David Bemmer_)

### DIFF
--- a/src/editor/prompt/local_filter_source.js
+++ b/src/editor/prompt/local_filter_source.js
@@ -1,5 +1,5 @@
 import BaseSource from "./base_source"
-import { filterMatches } from "../../helpers/string_helper"
+import { filterMatchPosition } from "../../helpers/string_helper"
 
 const MAX_RENDERED_SUGGESTIONS = 100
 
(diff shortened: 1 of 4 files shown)


src/editor/prompt/local_filter_source.js
```diff
@@ -19,21 +19,41 @@ export default class LocalFilterSource extends BaseSource {
...
+        matches.push({ promptItem, position })
       }
     }
 
+    matches.sort((a, b) => a.position - b.position)
+
+    const listItems = []
+    for (const { promptItem } of matches) {
+      if (listItems.length >= MAX_RENDERED_SUGGESTIONS) break
+      const listItem = this.buildListItemElementFor(promptItem)
```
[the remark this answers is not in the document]
zoltanhosszu: The previous code also iterated all prompt items on every keystroke — it just stopped building DOM elements after 100. The per-item cost (a string comparison) is unchanged. The only new cost is the sort, which is O(M log M) where M is the number of matches, not all items. Local prompt sources are inherently bounded (inline elements or a single deferred fetch), and truly large datasets use `RemoteFilterSource` which is unaffected. Adding a heap/short-circuit for this would be unnecessary complexity.

src/editor/prompt/local_filter_source.js
```diff
@@ -19,21 +19,41 @@ export default class LocalFilterSource extends BaseSource {
...
-        this.promptItemByListItem.set(listItem, promptItem)
-        listItems.push(listItem)
+      const position = filterMatchPosition(searchableText, filter)
+      if (position >= 0) {
+        matches.push({ promptItem, position })
       }
     }
 
+    matches.sort((a, b) => a.position - b.position)
+
```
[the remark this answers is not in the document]
zoltanhosszu: Added https://github.com/basecamp/lexxy/pull/975/commits/459ace2e08c1674e98d50c49fbf083e6b6a513ce

src/helpers/string_helper.js
```diff
@@ -22,7 +22,26 @@ export function normalizeFilteredText(string) {
 }
 
 export function filterMatches(text, potentialMatch) {
-  return normalizeFilteredText(text).includes(normalizeFilteredText(potentialMatch))
+  return filterMatchPosition(text, potentialMatch) >= 0
+}
+
+export function filterMatchPosition(text, potentialMatch) {
```
jorgemanrubia, line 24: You could use a regexp here for the match:

```js
export function filterMatchPosition(text, potentialMatch) {
    const normalizedText = normalizeFilteredText(text)
    const normalizedMatch = normalizeFilteredText(potentialMatch)

    if (!normalizedMatch) return 0

    const match = normalizedText.match(new RegExp(`(?:^|(?<= ))${RegExp.escape(normalizedMatch)}`))
    return match ? match.index : -1
  }
```
`\b` is slightly different, will match after punctuation like `-`, but I think that is actually a feature here.

jorgemanrubia, line 24: I don't think you are using the position/index right? You only use this in a boolean way. I think you could that reuse the name `filterMatchPosition` with the new implementation, and make the method return a boolean instead. With the suggested implementation (untested):

```js
export function filterMatches(text, potentialMatch) {
    const normalizedText = normalizeFilteredText(text)
    const normalizedMatch = normalizeFilteredText(potentialMatch)

    if (!normalizedMatch) return true

    return new RegExp(`(?:^|(?<= ))${RegExp.escape(normalizedMatch)}`).test(normalizedText)
  }
```

jorgemanrubia (commented):
👏

src/helpers/string_helper.js
zoltanhosszu: @jorgemanrubia ah this is nice! Thanks, I updated it.

zoltanhosszu: @jorgemanrubia the position is used numerically, not just as a boolean. In `local_filter_source.js` (line 37) the matches are sorted by position: `matches.sort((a, b) => a.position - b.position)`. That's what puts first name matches before last names.

jorgemanrubia: Oh understood! Then as a minor suggestion I would name it `indexOfFilterMatch` or `filterMatchIndex` since `index` is kind  of a common convention in ruby and JS.

jorgemanrubia (approved):

PR #998, opened 2026-04-20
Ctrl/cmd+click to open links without `contenteditable=false`

zachasme:
The current [link opener](https://github.com/basecamp/lexxy/pull/976) has [issues with cursor movement](https://app.3.basecamp.com/2914079/buckets/41746046/card_tables/cards/9798515282#__recording_9803371340).

Replaces that with a `CLICK_COMMAND` and separate modifier key tracking for cursor styling. Threw in an `auxclick` handler for opening with middle-mouse.

### DIFF
--- a/src/extensions/link_opener_extension.js
+++ b/src/extensions/link_opener_extension.js
@@ -1,7 +1,9 @@
-import { defineExtension } from "lexical"
-import { IS_APPLE, mergeRegister } from "@lexical/utils"
-import { registerEventListener } from "../helpers/listener_helper.js"
-import LexxyExtension from "./lexxy_extension.js"
+import { $findMatchingParent, $getNearestNodeFromDOMNode, CLICK_COMMAND, COMMAND_PRIORITY_NORMAL, defineExtension, mergeRegister } from "lexical"
+import { $isLinkNode } from "@lexical/link"
+import { IS_APPLE } from "@lexical/utils"
+import { registerEventListener } from "../helpers/listener_helper"
+import { delay } from "../helpers/timing_helpers"
+import LexxyExtension from "./lexxy_extension"
 
 export class LinkOpenerExtension extends LexxyExtension {
   get enabled() {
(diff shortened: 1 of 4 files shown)


jorgemanrubia (approved):
👏

src/extensions/link_opener_extension.js
```diff
@@ -11,53 +13,57 @@ export class LinkOpenerExtension extends LexxyExtension {
...
+  #isModified(event) {
+    return IS_APPLE ? event.metaKey : event.ctrlKey
+  }
+}
+
+function $openLink(target) {
+  const node = $getNearestNodeFromDOMNode(target)
+  const linkNode = $findMatchingParent(node, $isLinkNode)
+  if (linkNode) {
+    const url = linkNode.sanitizeUrl(linkNode.getURL())
```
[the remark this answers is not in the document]
zachasme: 🤖 `$openLink` is only reachable from `#handleClick` (which runs inside a Lexical command, so the target is always a DOM node inside the editor) and `#handleAuxClick` (which listens on `editorContentElement`). The target is always a valid editor DOM node, so `$getNearestNodeFromDOMNode` will always return a node. Adding a guard here would be dead code.

src/extensions/link_opener_extension.js
```diff
@@ -11,53 +13,57 @@ export class LinkOpenerExtension extends LexxyExtension {
...
-    for (const anchor of this.#anchors) {
-      anchor.setAttribute("contenteditable", "false")
-      anchor.setAttribute("target", "_blank")
-      anchor.setAttribute("rel", "noopener noreferrer")
-    }
+  // Chrome dispatches events without modifier keys *for a while* after changing tabs
+  async #handleFocus() {
+    await delay(200)
+    this.editorElement.addEventListener("mousemove", this.#updateOpenableAttribute.bind(this), { once: true })
   }
```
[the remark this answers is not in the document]
zachasme: 🤖 Each stacked listener calls `toggleAttribute` with the same value, so the worst case is a few redundant no-op calls on the first mousemove. This is the same pattern the old code used and isn't worth the added complexity of tracking/cancelling.

test/browser/tests/editor/link_opener.test.js
```diff
@@ -11,50 +11,42 @@ test.describe("Link opener", () => {
...
-    await page.keyboard.down(modifier)
-    await expect(anchor).toHaveAttribute("contenteditable", "false")
-    await page.keyboard.up(modifier)
-    await expect(anchor).not.toHaveAttribute("contenteditable")
+    const [newPage] = await Promise.all([
+      context.waitForEvent("page"),
+      anchor.click({ modifiers: [modifier] }),
+    ])
+
+    expect(newPage.url()).toBe("https://example.com/")
```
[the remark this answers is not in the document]
zachasme: 🤖 Good catch — added `waitForLoadState("load")` before the assertion.

test/browser/tests/editor/link_opener.test.js
```diff
@@ -11,50 +11,42 @@ test.describe("Link opener", () => {
...
-    await editor.setValue(
-      '<p><a href="https://a.com">first</a> and <a href="https://b.com">second</a></p>',
-    )
-    await editor.flush()
+    await anchor.click()
+    await page.waitForTimeout(200)
 
-    const anchors = editor.content.locator("a")
+    expect(opened).toBe(false)
+  })
```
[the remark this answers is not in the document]
zachasme: 🤖 Fair point about the dangling listener. Replaced with `context.waitForEvent("page", { timeout: 200 }).catch(() => null)` which cleans up after itself.

src/extensions/link_opener_extension.js
```diff
@@ -11,53 +13,57 @@ export class LinkOpenerExtension extends LexxyExtension {
...
 
-  #refresh() {
-    // Chrome dispatches events without modifier keys *for a while* after changing tabs
-    setTimeout(() => {
-      window.addEventListener("mousemove", this.#update.bind(this), { once: true })
-    }, 200)
+  #handleAuxClick(event) {
+    if (event.button === 1) {
+      this.editorElement.editor.read(() => $openLink(event.target))
```
samuelpecher, line 36: In the context of a command no need for `editor.read()`

src/extensions/link_opener_extension.js
```diff
@@ -11,53 +13,57 @@ export class LinkOpenerExtension extends LexxyExtension {
...
 
-  #enable() {
-    for (const anchor of this.#anchors) {
-      anchor.setAttribute("contenteditable", "false")
-      anchor.setAttribute("target", "_blank")
-      anchor.setAttribute("rel", "noopener noreferrer")
-    }
+  // Chrome dispatches events without modifier keys *for a while* after changing tabs
+  async #handleFocus() {
+    await delay(200)
```
samuelpecher, line 46: could this be gated with `IS_CHROME` since we're adding delay? No need to make safari any slower than it already is

src/extensions/link_opener_extension.js
```diff
@@ -11,53 +13,57 @@ export class LinkOpenerExtension extends LexxyExtension {
...
-    return this.editorElement.editorContentElement?.querySelectorAll("a") ?? []
+  #isModified(event) {
+    return IS_APPLE ? event.metaKey : event.ctrlKey
+  }
+}
+
+function $openLink(target) {
+  const node = $getNearestNodeFromDOMNode(target)
+  const linkNode = $findMatchingParent(node, $isLinkNode)
+  if (linkNode) {
```
samuelpecher, line 62: nit, but how about `if (!linkNode) return false` ?

src/extensions/link_opener_extension.js
zachasme: The `auxclick` is regular event listener, I couldn't find a corresponding Lexical command.

src/extensions/link_opener_extension.js
zachasme: Oh yes, sure thing!

src/extensions/link_opener_extension.js
zachasme: Right! Early returns is the general Lexxy style?

src/extensions/link_opener_extension.js
samuelpecher: Huh; odd for it to be missing. My bad.

src/extensions/link_opener_extension.js
samuelpecher: Turns out [no](https://github.com/basecamp/lexxy/pull/1005#discussion_r3133602047), but the code base doesn't really reflect that

PR #999, opened 2026-04-21
History-safe attachment uploads

samuelpecher:
PoC for making AttachmentNode history safe. Any non-user driven updates
to nodes pollutes the history stack; this moves all node updating logic
out of the node scope so changes are not commited to the history stack.

There are significant downsides to this approach, as can be seen by the
current state of the code

> [!WARNING]
> Proof of concept: do not merge


samuelpecher:
Closing. See #932

PR #1001, opened 2026-04-21
Skip no-op aria-pressed/aria-disabled writes on toolbar buttons

monorkin:
### Summary

The toolbar registers a Lexical registerUpdateListener that fires on every editor update — i.e., every keystroke — and unconditionally rewrites aria-pressed on ~16 toolbar buttons, aria-disabled on undo/redo, and aria-pressed on the color buttons in the highlight dropdown. Most of these writes are no-ops value-wise ("false" → "false"), but Chrome still invalidates element style on every setAttribute call.

Under normal conditions this is cheap. When another part of the page triggers a global style invalidation in the same frame as a keystroke (e.g. a Turbo broadcast-refresh flipping aria-busy on <html>), each no-op write forces Chrome to flush the pending recalc synchronously to commit the attribute change. With ~20 writes per keystroke and ~2.3k descendants to restyle, the cost piles up into hundreds of ms of blocking per keystroke.

Gating each write on an actual value change eliminates the no-op cost.

### Measured impact

Reproduction: Basecamp card table (~3.2k element DOM) open in one tab, new-card modal with Lexxy notes field focused in a second tab. A background script moves cards in the first tab once per second, which triggers broadcast_refresh_later on every move and morphs <html>[aria-busy] on every descendant.

Per-keystroke cost (median over ~150 real keystrokes, measured from beforeinput to post-rAF macrotask):

```
┌──────────────────────────┬────────┬──────────┬──────────┬────────┐
│        Condition         │  p50   │   p90    │   p99    │  avg   │
├──────────────────────────┼────────┼──────────┼──────────┼────────┤
│ Lexxy, no churn          │  14 ms │    50 ms │   108 ms │  24 ms │
├──────────────────────────┼────────┼──────────┼──────────┼────────┤
│ Lexxy, churn on — before │ 591 ms │ 1,300 ms │ 1,806 ms │ 670 ms │
├──────────────────────────┼────────┼──────────┼──────────┼────────┤
│ Lexxy, churn on — after  │ 219 ms │   602 ms │ 1,725 ms │ 290 ms │
├──────────────────────────┼────────┼──────────┼──────────┼────────┤
│ Relative change          │  −63 % │    −54 % │     −4 % │  −57 % │
└──────────────────────────┴────────┴──────────┴──────────┴────────┘
```

Plain <input> in the same modal under the same churn: p50 36 ms, p90 437 ms — confirming the broadcast-induced style invalidation is the trigger; the patch addresses Lexxy's amplifier on top of that.

### Change

- src/elements/toolbar.js — #setButtonPressed / #setButtonDisabled read before writing and skip if the attribute is already in the target state.
- src/elements/dropdown/highlight.js — same guard on the color buttons' aria-pressed.

Observable behavior is unchanged: attribute values end at the same state as before; only redundant writes are skipped.

### Tests

- yarn lint — clean
- yarn test (Vitest) — 80/80 passed
- yarn test:browser:chromium (Playwright) — 315 passed (3 preexisting flakes in attachments.test.js, unrelated)

### DIFF
--- a/src/elements/dropdown/highlight.js
+++ b/src/elements/dropdown/highlight.js
@@ -94,7 +94,10 @@ export class HighlightDropdown extends ToolbarDropdown {
 
     this.#colorButtons.forEach(button => {
       const matchesSelection = button.dataset.value === textColor || button.dataset.value === backgroundColor
-      button.setAttribute("aria-pressed", matchesSelection)
+      const next = matchesSelection.toString()
+      if (button.getAttribute("aria-pressed") !== next) {
+        button.setAttribute("aria-pressed", next)
+      }
     })
 
     const hasHighlight = textColor !== NO_STYLE || backgroundColor !== NO_STYLE
--- a/src/elements/prompt.js
+++ b/src/elements/prompt.js
@@ -224,7 +224,7 @@ export class LexicalPromptElement extends HTMLElement {
   }
 
   #selectOption(listItem) {
-    this.#clearSelection()
+    this.#clearListItemSelection()
     listItem.toggleAttribute("aria-selected", true)
     listItem.scrollIntoView({ block: "nearest", behavior: "smooth" })
     listItem.focus()
@@ -234,16 +234,26 @@ export class LexicalPromptElement extends HTMLElement {
       this.#editorElement.focus()
     })
 
-    this.#editorContentElement.setAttribute("aria-controls", this.popoverElement.id)
-    this.#editorContentElement.setAttribute("aria-activedescendant", listItem.id)
-    this.#editorContentElement.setAttribute("aria-haspopup", "listbox")
+    this.#setEditorAssociationAttribute("aria-controls", this.popoverElement.id)
+    this.#setEditorAssociationAttribute("aria-activedescendant", listItem.id)
+    this.#setEditorAssociationAttribute("aria-haspopup", "listbox")
+  }
+
+  #clearListItemSelection() {
+    this.#listItemElements.forEach((item) => { item.toggleAttribute("aria-selected", false) })
   }
 
   #clearSelection() {
-    this.#listItemElements.forEach((item) => { item.toggleAttribute("aria-selected", false) })
+    this.#clearListItemSelection()
     this.#editorContentElement.removeAttribute("aria-controls")
     this.#editorContentElement.removeAttribute("aria-activedescendant")
     this.#editorContentElement.removeAttribute("aria-haspopup")
+  }
+
+  #setEditorAssociationAttribute(name, value) {
+    if (this.#editorContentElement.getAttribute(name) !== value) {
+      this.#editorContentElement.setAttribute(name, value)
+    }
   }
 
   #positionPopover() {
--- a/src/elements/toolbar.js
+++ b/src/elements/toolbar.js
@@ -240,15 +240,23 @@ export class LexicalToolbarElement extends HTMLElement {
   #setButtonPressed(name, isPressed) {
     const button = this.querySelector(`[name="${name}"]`)
     if (button) {
-      button.setAttribute("aria-pressed", isPressed.toString())
+      const next = isPressed.toString()
+      if (button.getAttribute("aria-pressed") !== next) {
+        button.setAttribute("aria-pressed", next)
+      }
     }
   }
 
   #setButtonDisabled(name, isDisabled) {
     const button = this.querySelector(`[name="${name}"]`)
     if (button) {
-      button.disabled = isDisabled
-      button.setAttribute("aria-disabled", isDisabled.toString())
+      if (button.disabled !== isDisabled) {
+        button.disabled = isDisabled
+      }
+      const next = isDisabled.toString()
+      if (button.getAttribute("aria-disabled") !== next) {
+        button.setAttribute("aria-disabled", next)
+      }
     }
   }
 


samuelpecher (commented):
Great find. We can probably trim when we're updating states as well.

monorkin:
@samuelpecher I can take care of that tomorrow morning

monorkin:
Forgot to add documentation yesterday

JS:
- [MutationObserver fires even if the attribute value didn't change](https://github.com/whatwg/dom/issues/520) - this one is important for us because of Stimulus (also, this is the source of all the other points in this list)
- [WebComponent hooks trigger even if the attribute value didn't change](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_custom_elements#responding_to_attribute_changes)
- [MutationRecord is triggered even if the attribute value didn't change](https://developer.mozilla.org/en-US/docs/Web/API/MutationRecord) - this one is important for us because of idiomorph / turbo-morph

CSS:
This took a bit of spelunking across mailing lists and Blink's documentation. I used this as a basis and assumed that FF and Safari will perform worse.

When an attribute change happens, of any kind, the node is added to an invalidation set and it and all it's children are marked for a recalculation.

If there are no rules that apply the browser does a no-op. If there are rules it performs a style invalidation.

In Lexxy, we ship CSS rules for those attributes so any cursor movement causes a recalc.

In the general case, this change is negligible, but if the page is under a lot of pressure from consistent morphs then this starts to make a difference because simply typing won't cause recalcs or invalidations.

monorkin:
**From Claude:**

Following up on the toolbar guard, I audited every attribute-write site in lexxy `src/` for the same no-op pattern. Added one more guard as a second commit (`389fcc77` — *Skip no-op aria writes on editor content element from prompt popover*); the rest are either already safe or genuinely need to run every time they're called. Full assessment below.

## New change: prompt popover (`src/elements/prompt.js`)

`#selectOption` runs on every arrow-key press while navigating a mention / command popover. Each call was first tearing down three aria-\* attributes on the editor content element via `#clearSelection`, then immediately rewriting them — two of which (`aria-controls`, `aria-haspopup`) back to the same value they held before.

Because `removeAttribute` on a present attribute always fires a real mutation (no short-circuit on "same value, then set again"), guarding only the `setAttribute` side wouldn't help. Split `#clearSelection` into two concerns:

- `#clearListItemSelection` — just toggles `aria-selected` on list items (used by `#selectOption`).
- `#clearSelection` — still removes the three editor-content aria-\* attrs (used by `#hidePopover`, where a real teardown is what we want).

And guarded the three `setAttribute` writes in `#selectOption` via a new `#setEditorAssociationAttribute(name, value)` helper that reads before writing — same pattern as `#setButtonPressed` / `#setButtonDisabled` on the toolbar.

Per arrow-key before: 6 attribute mutations (3 removes + 3 re-adds, 4 of which were no-op value-wise).
Per arrow-key after: 1 mutation (`aria-activedescendant`, which genuinely changes).

## Audit of every other attribute-write site

### Already guarded in this PR

| Site | Call path |
|---|---|
| `toolbar.js #setButtonPressed` (L240) | `aria-pressed` on ~16 buttons, per editor update |
| `toolbar.js #setButtonDisabled` (L247) | `aria-disabled` + `.disabled` on undo/redo, per editor update |
| `dropdown/highlight.js #updateColorButtonStates` (L95-98) | `aria-pressed` on color buttons, on dropdown open |
| `prompt.js #selectOption` (L237-239, this commit) | three aria-\* on editor content element, per arrow-key in popover |

### Skipped — spec-level no-ops already

`classList.toggle(name, force)` and `toggleAttribute(name, force)` are [spec-required](https://dom.spec.whatwg.org/#dom-domtokenlist-toggle) to do nothing when state already matches `force`. No mutation record, no style invalidation. All of these are fine as-is, even the ones on hot paths:

| Site | Notes |
|---|---|
| `editor.js:587` `classList.toggle("lexxy-editor--empty", this.isEmpty)` | fires per-update via `#synchronizeWithChanges` — but no-op when state matches |
| `nodes/provisional_paragraph_node.js:24, 30` `classList.toggle("hidden", !selected)` | runs inside `createDOM` / `updateDOM` during reconcile |
| `toolbar.js:271, 272` `toggleAttribute("overflowing", …)` / `toggleAttribute("disabled", …)` | `#refreshToolbarOverflow` |
| `prompt.js:228, 243, 258, 264, 269` `toggleAttribute("aria-selected" / "data-anchored" / "data-clipped-at-right" / "data-clipped-at-bottom", …)` | popover selection / positioning |
| `extensions/link_opener_extension.js:51` `toggleAttribute("data-links-openable", …)` | modifier-key tracking |
| `elements/prompt.js:32`, `editor.js:87`, `toolbar.js:73` `toggleAttribute("connected", true)` | one-shot on `connectedCallback` |

### Skipped — writes happen on detached elements

Elements created with `document.createElement` aren't in the document tree at the time of `setAttribute`. No subscribers, no style invalidation.

| Site | Notes |
|---|---|
| `helpers/code_highlighting_helper.js:107` `mark.setAttribute("style", style)` | `<mark>` created at L106, inserted at L113 |
| `editor/selection.js:710` `marker.setAttribute("nonce", getNonce())` | marker created, briefly inserted to read geometry, then removed |

### Skipped — init / one-shot / cold paths

Run once (or very rarely) per editor / toolbar / element lifetime.

| Site | Trigger |
|---|---|
| `editor/extensions.js:33` `child.setAttribute("data-lexxy-extension", "")` | extension registration |
| `elements/code_language_picker.js:43` `setAttribute("nonce", …)` | picker render |
| `elements/prompt.js:471` `popoverContainer.setAttribute("nonce", …)` | popover creation |
| `elements/table/table_tools.js:65` `dropdown.setAttribute("name", "lexxy-dropdown")` | table tools setup |
| `elements/toolbar.js:26, 53` `setAttribute("role", "toolbar")` / `setAttribute(`data-${button}`, …)` | `connectedCallback` + `configure` |
| `elements/toolbar.js:268` `#overflow.setAttribute("nonce", getNonce())` | only from `#refreshToolbarOverflow` (ResizeObserver or mount) |
| `elements/editor.js:369-375` `editorContentElement.setAttribute(...)` | initial aria/tabindex copy on mount |
| `elements/editor.js:580` `toolbar.setAttribute("data-attachments", …)` | default-toolbar creation |
| `helpers/style_resolver_root.js:19-20` `setAttribute("aria-hidden" / "data-lexxy-style-resolver")` | style resolver root construction (once) |

### Low-impact but cheap if someone wants to follow up

Not hot enough to matter for the lag repro, but match the same "idempotent value / repeated call" shape:

| Site | Trigger |
|---|---|
| `elements/toolbar_dropdown.js:73` `element.setAttribute("tabindex", index === 0 ? 0 : "-1")` | `#resetTabIndexValues`, runs on dropdown open/close/focus change |
| `elements/dropdown/highlight.js:104` `this.querySelector(REMOVE_HIGHLIGHT_SELECTOR).disabled = !hasHighlight` | only fires when the highlight dropdown opens |

### Property-reflection writes (`.disabled` / `.hidden` / `.contentEditable`)

Mirror to attributes, so same concern applies. I checked these too:

| Site | Status |
|---|---|
| `elements/toolbar.js:254` `button.disabled = isDisabled` | **guarded in this PR** |
| `elements/dropdown/highlight.js:104` `.disabled = !hasHighlight` | cold path, left alone |
| `editor/adapters/native_adapter.js:40, 44` `contentEditable = "true" / "false"` | freeze/thaw flow, rare |
| `elements/code_language_picker.js:155, 159` `this.hidden = true / false` | picker show/hide, user action |

## Why I didn't wrap more things

- **`classList.toggle(name, force)` and `toggleAttribute(name, force)` are already idempotent at the DOM layer.** Adding a `getAttribute`/`contains` check ahead of them would be pure noise — the browser already does that check internally.
- **Detached-element writes don't invalidate anything** — there's nothing subscribed to mutate against. Wrapping them adds a read for no win.
- **Init / one-shot writes aren't hot.** Wrapping them protects against a call frequency that doesn't occur in practice, and makes the code harder to read.

The guard pattern is worth applying specifically when all three of these hold: (1) the write runs in a hot, selection-driven path, (2) the attribute is written via `setAttribute`/`removeAttribute` (no `force` short-circuit), and (3) the typical call produces the value the element already has. That's exactly the toolbar-button and prompt-popover cases, and nothing else.

## Tests

- `yarn lint` — clean
- `yarn test` (Vitest) — 80/80 passed
- `yarn test:browser:chromium` — 316 passed, 2 preexisting flakes in `attachments/*`
- `yarn test:browser:firefox` — 315 passed, 1 preexisting flake in `attachments/*`

The `test/browser/tests/prompts/*` suites — which exercise the `#selectOption` / `#clearSelection` paths directly — pass in both browsers.

samuelpecher (approved):
Nice optimization

PR #1003, opened 2026-04-21
Coalesce per-keystroke update listeners to the next animation frame

monorkin:
### Summary

Three Lexical registerUpdateListener callbacks run synchronously on every editor update (i.e. every keystroke):

- LexicalToolbarElement#monitorSelectionChanges → walks the selection, refreshes aria-pressed/aria-disabled on ~16 toolbar buttons, closes dropdowns.
- LexicalToolbarElement#monitorHistoryChanges → refreshes undo/redo button state.
- LexicalEditorElement#synchronizeWithChanges → re-serializes the editor to HTML via this.value, updates the internal form value, re-runs the validity check, and dispatches lexxy:change.

None of this observer work needs to land in the same frame as the triggering keystroke. A one-frame deferral is imperceptible to users (the toolbar's pressed state ~16 ms late, a lexxy:change event ~16 ms late) but pulls the work off the
input-handling critical path — which matters when the page is under layout-invalidation pressure from something else (e.g. a Turbo broadcast-refresh flipping aria-busy on <html>). Under that pressure, every DOM mutation the listeners do forces
Chrome to flush a pending full-document style recalc; deferring them means the keystroke frame only pays for the Lexical reconciler's own minimal writes plus the browser's caret layout.

Add a small coalesceOnFrame(fn) helper in timing_helpers.js that collapses repeated calls within a single animation frame into one invocation on the next frame (using the latest arguments). Wrap each of the three update listeners with it.

This is independent of (and stacks with) #1001, but either can land first.

### Measured impact

Reproduction: Basecamp card table (~3.2k element DOM) open in one tab, new-card modal with a Lexxy notes field focused in another. A background script moves cards once per second, causing a Turbo broadcast_refresh_later each time, which flips
aria-busy on <html> and invalidates style on ~2.3k descendants.

Per-keystroke cost (median over ~150–220 real keystrokes, measured from beforeinput to post-rAF macrotask):

```
┌─────────────────────────────────────────────────────┬────────┬──────────┬──────────┬────────┬──────────────────────┐
│                      Condition                      │  p50   │   p90    │   p99    │  avg   │ Long tasks in sample │
├─────────────────────────────────────────────────────┼────────┼──────────┼──────────┼────────┼──────────────────────┤
│ Lexxy, no churn (baseline)                          │  14 ms │    50 ms │   108 ms │  24 ms │                    0 │
├─────────────────────────────────────────────────────┼────────┼──────────┼──────────┼────────┼──────────────────────┤
│ Lexxy, churn on — unpatched                         │ 591 ms │ 1,300 ms │ 1,806 ms │ 670 ms │                   70 │
├─────────────────────────────────────────────────────┼────────┼──────────┼──────────┼────────┼──────────────────────┤
│ Lexxy, churn on — #1001 only (aria-write guard)     │ 219 ms │   602 ms │ 1,725 ms │ 290 ms │                  104 │
├─────────────────────────────────────────────────────┼────────┼──────────┼──────────┼────────┼──────────────────────┤
│ Lexxy, churn on — this PR + #1001, run 1 (174 keys) │  17 ms │    34 ms │    55 ms │  19 ms │                   10 │
├─────────────────────────────────────────────────────┼────────┼──────────┼──────────┼────────┼──────────────────────┤
│ Lexxy, churn on — this PR + #1001, run 2 (220 keys) │  15 ms │    32 ms │    68 ms │  19 ms │                   27 │
└─────────────────────────────────────────────────────┴────────┴──────────┴──────────┴────────┴──────────────────────┘
```

For reference, a plain <input> in the same modal under the same churn measured p50 36 ms, p90 437 ms — so with both patches applied, Lexxy is faster than a native input under the same load and sits within a few ms of its own no-churn baseline
at the median.

### Change

- src/helpers/timing_helpers.js — new coalesceOnFrame(fn) helper: returns a wrapped function that, when called repeatedly in one frame, runs fn at most once on the next frame with the most recent arguments.
- src/elements/toolbar.js — #monitorSelectionChanges and #monitorHistoryChanges wrap their registerUpdateListener callbacks with coalesceOnFrame.
- src/elements/editor.js — #synchronizeWithChanges does the same.

Observable behavior is unchanged: toolbar state still reflects the current selection, the internal form value still tracks content, lexxy:change still fires on real content changes. The only difference is a one-frame deferral of these observer
effects.

Tests

- yarn lint — clean
- `yarn test` (Vitest) — 80/80 passed
- `yarn test:browser:chromium` (Playwright) — 314 passed, 4 flaky in `attachments/*` (same preexisting flakes observed on #1001 and on `main`; unrelated to this change — they pass on retry)

### DIFF
--- a/src/elements/editor.js
+++ b/src/elements/editor.js
@@ -20,6 +20,7 @@ import { createElement, dispatch, generateDomId, parseHtml } from "../helpers/ht
 import { isAttachmentSpacerTextNode } from "../helpers/lexical_helper"
 import { sanitize, setSanitizerConfig } from "../helpers/sanitization_helper"
 import { ListenerBin, registerEventListener } from "../helpers/listener_helper"
+import { coalesceOnFrame } from "../helpers/timing_helpers"
 import LexicalToolbar from "./toolbar"
 import Configuration from "../editor/configuration"
 import Contents from "../editor/contents"
@@ -420,13 +421,14 @@ export class LexicalEditorElement extends HTMLElement {
   }
 
   #synchronizeWithChanges() {
-    this.#listeners.track(this.editor.registerUpdateListener(({ editorState }) => {
+    const synchronize = coalesceOnFrame(() => {
       this.#clearCachedValues()
       this.#internalFormValue = this.value
       this.#toggleEmptyStatus()
       this.#setValidity()
       this.#dispatchAttributesChange()
-    }))
+    })
+    this.#listeners.track(this.editor.registerUpdateListener(synchronize))
   }
 
   #clearCachedValues() {
--- a/src/elements/toolbar.js
+++ b/src/elements/toolbar.js
@@ -8,6 +8,7 @@ import { ListenerBin, registerEventListener } from "../helpers/listener_helper"
 import { handleRollingTabIndex } from "../helpers/accessibility_helper"
 import ToolbarIcons from "./toolbar_icons"
 import { isActiveAndVisible } from "../helpers/html_helper"
+import { coalesceOnFrame } from "../helpers/timing_helpers"
 
 export class LexicalToolbarElement extends HTMLElement {
   static observedAttributes = [ "connected" ]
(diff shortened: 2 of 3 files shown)


monorkin:
Claude made a mistake in its performance test here. 
After manual testing, this PR on its own only improves p99 down from 1750ms to 1300ms.

It makes the average case a bit worse, but brings the worst case way down — this is also visible in the browser benchmark CI that failed. Else it's on-pair with #1001, and when paired with it the average case is unchanged.

samuelpecher:
Interestingly, this kind of batching is meant to come free with lexical's design, so we might be holding it wrong somewhere (forcing an update flush with an editor state read etc.)

monorkin:
Had to read up on Lexical lore to understand your comment 😃

Lexical uses `queueMicrotask`, that's similar to `setTimeout(0, fn)`, which executes on the VM's next tick.
This PR moves execution to the next visible frame (which can be several ticks).

The batching that Lexical does is on the microtask-level. So any `editor.update()` calls will get batched into a single microtask.

Now that I know all this, I think this change isn't worth it. It will likely cause downstream problems.

PR #1005, opened 2026-04-23
Honor permitted-attachment-types across the editor

flavorjones:
Honor data-permitted-attachment-types across the editor

The motivation for this work is a long-standing round-trip bug in host apps that want to support only certain attachment shapes on a given form — for example, mentions and OG embeds but not file uploads. With Lexxy honoring a per-editor allowlist directly, hosts can declare `permitted-attachment-types` and keep `attachments=true`: saved mentions and embeds continue to render, while new file drops, disallowed programmatic inserts, and disallowed prompt activations are all rejected at the editor boundary.

Lexxy now reads `data-permitted-attachment-types` on `<lexxy-editor>` and enforces it at every attachment entry point. The attribute becomes a first-class per-editor allowlist, and the custom element exposes two new methods that host apps can consult directly: a `permittedAttachmentTypes` getter that returns the parsed token list (or `null` when the attribute is absent), and a `permitsAttachmentContentType(contentType)` predicate that returns whether a given content-type is currently allowed.

Four consumers plug into that predicate:

1. HTML import. A new helper, `$generateFilteredNodesFromDOM`, wraps `$generateNodesFromDOM` at every call site in the import path — value-set, insertDOM, programmatic HTML insert, and prompt editable-text insertion. The wrapper walks the generated nodes and their descendants, dropping any attachment node whose content-type is not permitted, so that nested cases (such as a disallowed attachment inside a table cell) are caught as well as top-level ones.

2. File drops and pastes. An internal `lexxy:file-accept` listener, registered at editor initialization, calls `event.preventDefault()` when the dropped file's MIME is disallowed. Host listeners can still stack on top, because the internal listener installs before the `lexxy:initialize` event fires, so any host code that registers its own handler on initialize retains last-word semantics.

3. Prompt activation. A prompt whose declared template content-type is not permitted never installs its trigger listener, so its popover never appears on the page. The activation gate also short-circuits to "not permitted" when the editor is `attachments=false`.

4. Programmatic insertion. The `lexxy:insert-link` event's `replaceLinkWith` and `insertBelowLink` callbacks now fall back to plain HTML (which itself routes through the import filter, so any attachment-producing tags nested inside the fallback HTML are also dropped) when the requested content-type is disallowed, rather than unconditionally constructing an attachment node as before.

The predicate also returns `false` for every content-type when the editor is `attachments=false`, which means that case and a declared-but-empty allowlist are observationally equivalent for import, upload, and prompt-activation paths.

The attribute is parsed once at `connectedCallback` and cached on the element, matching Lexxy's existing init-only convention for configuration attributes; a host that needs to change the allowlist at runtime should disconnect and reconnect the element. As a backwards compatibility aid for consumers that still emit the legacy `"false"` string sentinel, the getter strips a literal `"false"` token from the parsed list, so that sentinel reduces to an empty list rather than a permit-only-the-word-`false` list.

An unset attribute preserves today's default: everything is permitted and no filtering happens. External consumers that do not set the attribute see no behavior change. Consumers that do set it get editor-level enforcement for free, rather than having to implement the allowlist check themselves.

New Playwright tests under `test/browser/tests/attachments/`, `test/browser/tests/prompts/`, and `test/browser/tests/paste/` cover import filtering, nested-in-table stripping, file-MIME rejection, prompt suppression by both content-type and `attachments=false`, programmatic-insert fallback shape, OG-embed round-trip, default and sentinel behaviors, and the XSS sanitization of arbitrary attachment attributes. A new dummy-app system test, `test/system/mention_round_trip_test.rb`, verifies the complete Action Text round-trip — an `<action-text-attachment>` wrapper containing a `<bc-mention>` element — through the edit, save, render, and re-edit cycle.

### DIFF
--- a/src/elements/editor.js
+++ b/src/elements/editor.js
@@ -175,7 +175,7 @@ export class LexicalEditorElement extends HTMLElement {
   #parsePermittedAttachmentTypes() {
     const raw = this.dataset.permittedAttachmentTypes
     if (raw == null) return null
-    return raw.split(/\s+/).filter(t => t && t !== "false")
+    return Object.freeze(raw.split(/\s+/).filter(t => t && t !== "false"))
   }
 
   get isEmpty() {
--- a/src/elements/editor.js
+++ b/src/elements/editor.js
@@ -300,7 +300,7 @@ export class LexicalEditorElement extends HTMLElement {
 
   #parseHtmlIntoLexicalNodes(html) {
     if (!html) html = "<p></p>"
-    const nodes = $generateFilteredNodesFromDOM(this.editor, parseHtml(`${html}`), this)
+    const nodes = $generateFilteredNodesFromDOM(this, parseHtml(`${html}`))
 
     return nodes
       .filter(this.#isNotWhitespaceOnlyNode)
--- a/src/elements/editor.js
+++ b/src/elements/editor.js
@@ -167,15 +167,21 @@ export class LexicalEditorElement extends HTMLElement {
   }
 
   permitsAttachmentContentType(contentType) {
-    if (this.getAttribute("attachments") === "false") return false
-    const list = this.#permittedAttachmentTypes
-    return list === null || list.includes(contentType)
+    if (this.getAttribute("attachments") === "false") {
+      return false
+    } else {
+      const list = this.#permittedAttachmentTypes
+      return list === null || list.includes(contentType)
+    }
   }
 
   #parsePermittedAttachmentTypes() {
     const raw = this.dataset.permittedAttachmentTypes
-    if (raw == null) return null
-    return Object.freeze(raw.split(/\s+/).filter(t => t && t !== "false"))
+    if (raw == null) {
+      return null
+    } else {
+      return Object.freeze(raw.split(/\s+/).filter(t => t && t !== "false"))
+    }
   }
 
   get isEmpty() {
--- a/src/elements/editor.js
+++ b/src/elements/editor.js
@@ -54,7 +54,6 @@ export class LexicalEditorElement extends HTMLElement {
   #listeners = new ListenerBin()
   #disposables = []
   #historyState = { undo: false, redo: false }
-  #permittedAttachmentTypes = null
 
   constructor() {
     super()
@@ -66,7 +65,6 @@ export class LexicalEditorElement extends HTMLElement {
     this.id ||= generateDomId("lexxy-editor")
     this.config = new Configuration(this)
     this.extensions = new Extensions(this)
-    this.#permittedAttachmentTypes = this.#parsePermittedAttachmentTypes()
 
     this.editor = this.#createEditor()
     this.#disposables.push(this.editor)
@@ -163,24 +161,21 @@ export class LexicalEditorElement extends HTMLElement {
   }
 
   get permittedAttachmentTypes() {
-    return this.#permittedAttachmentTypes
-  }
-
-  permitsAttachmentContentType(contentType) {
-    if (this.getAttribute("attachments") === "false") {
-      return false
-    } else {
-      const list = this.#permittedAttachmentTypes
-      return list === null || list.includes(contentType)
-    }
-  }
-
-  #parsePermittedAttachmentTypes() {
-    const raw = this.getAttribute("permitted-attachment-types")
+    const raw = this.config.get("permittedAttachmentTypes")
     if (raw == null) {
       return null
     } else {
-      return Object.freeze(raw.split(/\s+/).filter(t => t && t !== "false"))
+      const tokens = Array.isArray(raw) ? raw : String(raw).split(/\s+/)
+      return Object.freeze(tokens.filter(t => t && t !== "false"))
+    }
+  }
+
+  permitsAttachmentContentType(contentType) {
+    if (!this.supportsAttachments) {
+      return false
+    } else {
+      const list = this.permittedAttachmentTypes
+      return list === null || list.includes(contentType)
     }
   }
 
--- a/src/elements/editor.js
+++ b/src/elements/editor.js
@@ -5,7 +5,8 @@ import { AutoLinkNode, LinkNode } from "@lexical/link"
 import { $getNearestNodeOfType } from "@lexical/utils"
 import { registerPlainText } from "@lexical/plain-text"
 import { HeadingNode, QuoteNode, registerRichText } from "@lexical/rich-text"
-import { $generateHtmlFromNodes, $generateNodesFromDOM } from "@lexical/html"
+import { $generateHtmlFromNodes } from "@lexical/html"
+import { $generateFilteredNodesFromDOM } from "../helpers/attachment_filter_helper"
 import { CodeHighlightNode, CodeNode, registerCodeHighlighting } from "@lexical/code"
 import { TRANSFORMERS, registerMarkdownShortcuts } from "@lexical/markdown"
 import { HORIZONTAL_DIVIDER } from "../editor/markdown/horizontal_divider_transformer"
@@ -53,6 +54,7 @@ export class LexicalEditorElement extends HTMLElement {
   #listeners = new ListenerBin()
   #disposables = []
   #historyState = { undo: false, redo: false }
+  #permittedAttachmentTypes = null
 
   constructor() {
     super()
@@ -64,6 +66,7 @@ export class LexicalEditorElement extends HTMLElement {
     this.id ||= generateDomId("lexxy-editor")
     this.config = new Configuration(this)
     this.extensions = new Extensions(this)
+    this.#permittedAttachmentTypes = this.#parsePermittedAttachmentTypes()
 
     this.editor = this.#createEditor()
     this.#disposables.push(this.editor)
@@ -157,6 +160,22 @@ export class LexicalEditorElement extends HTMLElement {
 
   get blobUrlTemplate() {
     return this.dataset.blobUrlTemplate
+  }
+
+  get permittedAttachmentTypes() {
+    return this.#permittedAttachmentTypes
+  }
+
+  permitsAttachmentContentType(contentType) {
+    if (this.getAttribute("attachments") === "false") return false
+    const list = this.#permittedAttachmentTypes
+    return list === null || list.includes(contentType)
+  }
+
+  #parsePermittedAttachmentTypes() {
+    const raw = this.dataset.permittedAttachmentTypes
+    if (raw == null) return null
+    return raw.split(/\s+/).filter(t => t && t !== "false")
   }
 
   get isEmpty() {
@@ -281,7 +300,7 @@ export class LexicalEditorElement extends HTMLElement {
 
   #parseHtmlIntoLexicalNodes(html) {
     if (!html) html = "<p></p>"
-    const nodes = $generateNodesFromDOM(this.editor, parseHtml(`${html}`))
+    const nodes = $generateFilteredNodesFromDOM(this.editor, parseHtml(`${html}`), this)
 
     return nodes
       .filter(this.#isNotWhitespaceOnlyNode)
@@ -313,11 +332,22 @@ export class LexicalEditorElement extends HTMLElement {
     this.#handleEnter()
     this.#registerFocusEvents()
     this.#registerHistoryEvents()
+    this.#registerFileAcceptFilter()
     this.#attachDebugHooks()
     this.#attachToolbar()
     this.#configureSanitizer()
     this.#loadInitialValue()
     this.#resetBeforeTurboCaches()
+  }
+
+  #registerFileAcceptFilter() {
+    this.#listeners.track(
+      registerEventListener(this, "lexxy:file-accept", (event) => {
+        if (!this.permitsAttachmentContentType(event.detail.file.type)) {
+          event.preventDefault()
+        }
+      })
+    )
   }
 
   #createEditor() {
--- a/src/elements/editor.js
+++ b/src/elements/editor.js
@@ -176,7 +176,7 @@ export class LexicalEditorElement extends HTMLElement {
   }
 
   #parsePermittedAttachmentTypes() {
-    const raw = this.dataset.permittedAttachmentTypes
+    const raw = this.getAttribute("permitted-attachment-types")
     if (raw == null) {
       return null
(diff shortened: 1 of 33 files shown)


src/helpers/attachment_filter_helper.js
```diff
@@ -0,0 +1,36 @@
...
+function walkChildren(node, editorElement) {
+  if (typeof node.getChildren !== "function") return
+  node.getChildren().forEach(child => {
+    if (isDisallowedAttachment(child, editorElement)) {
+      child.remove()
+    } else {
+      walkChildren(child, editorElement)
+    }
+  })
+}
```
samuelpecher: Since these are Lexical nodes you might be able to use [$descendantsMatching(node, isDisallowedAttachment)](https://lexical.dev/docs/api/modules/lexical_utils#descendantsmatching) and then remove them, or an iterative alternative with [$dfsIterator](https://lexical.dev/docs/api/modules/lexical_utils#dfsiterator):

```js
for (const { node: child } of $dfsIterator(node)) {
   if (isDisallowedAttachment(child, editorElement)) node.remove()
}
```

src/elements/prompt.js
```diff
@@ -117,6 +119,16 @@ export class LexicalPromptElement extends HTMLElement {
     }))
   }
 
+  get #promptContentTypePermitted() {
+    const el = this.#editorElement
+    if (el.getAttribute("attachments") === "false") return false
+
+    const perItem = Array.from(this.querySelectorAll("template[type='editor']"))
+      .map(t => t.getAttribute("content-type")).filter(Boolean)
```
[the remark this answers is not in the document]
flavorjones: Fixed in c6a65789 — #promptContentTypePermitted now maps each template to its effective content-type (explicit attribute or the prompt default) before asking the editor whether any of them is permitted, so a prompt that mixes templates with and without an explicit content-type is no longer suppressed when the default would have been accepted. Added a test in test/browser/tests/prompts/prompt_suppression.test.js with a matching fixture.

src/elements/editor.js
```diff
@@ -158,6 +161,22 @@ export class LexicalEditorElement extends HTMLElement {
...
+    if (this.getAttribute("attachments") === "false") return false
+    const list = this.#permittedAttachmentTypes
+    return list === null || list.includes(contentType)
+  }
+
+  #parsePermittedAttachmentTypes() {
+    const raw = this.dataset.permittedAttachmentTypes
+    if (raw == null) return null
+    return raw.split(/\s+/).filter(t => t && t !== "false")
+  }
```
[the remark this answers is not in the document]
flavorjones: Fixed in 9abcad07 — dropped the connectedCallback cache on #permittedAttachmentTypes and parse data-permitted-attachment-types on demand inside permitsAttachmentContentType, matching the dynamic behavior of the sibling attachments attribute. Added test/browser/tests/attachments/dynamic_allowlist.test.js to pin the runtime-update behavior.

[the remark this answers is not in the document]
flavorjones: Update: the earlier fix (9abcad07) has been reverted and force-pushed out. On a second look, `data-permitted-attachment-types` is intentionally init-only by design — every configuration attribute on `<lexxy-editor>` (`attachments`, `preset`, `rich-text`, `multi-line`, `markdown`, `toolbar`, `highlight`, `placeholder`, `single-line`) is read once at `connectedCallback`, and `observedAttributes` is deliberately limited to internal markers. Hosts that need to change the permitted-types list are expected to disconnect + reconnect the element. Parsing on demand (or observing the attribute) would diverge from that convention without a concrete consumer need, so we are keeping the connect-time parse as-is.

src/elements/editor.js
```diff
@@ -158,6 +159,22 @@ export class LexicalEditorElement extends HTMLElement {
...
+  get permittedAttachmentTypes() {
+    return this.#parsePermittedAttachmentTypes()
+  }
+
+  permitsAttachmentContentType(contentType) {
+    if (this.getAttribute("attachments") === "false") return false
+    const list = this.#parsePermittedAttachmentTypes()
+    return list === null || list.includes(contentType)
+  }
+
```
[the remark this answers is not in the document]
flavorjones: Out of scope — MIME wildcard matching is a new feature. Not implementing in this PR.

test/browser/tests/attachments/import_filter.test.js
```diff
@@ -0,0 +1,86 @@
...
+    await page.goto("/mentions-custom-element.html")
+    await page.waitForSelector("lexxy-editor[connected]")
+  })
+
+  test("disallowed attachment content-type is stripped on import", async ({ page, editor }) => {
+    await editor.setValue(imageAttachment)
+    await editor.flush()
+
+    await expect(editor.content.locator("bc-attachment[content-type='image/png']")).toHaveCount(0)
+    await expect(editor.content.locator("bc-attachment")).toHaveCount(0)
```
samuelpecher, line 23: A few tests have this. The editor's content wouldn't have attachment tags directly in it's content. They would manifest as `figure.attachment[data-content-type]`.

src/helpers/attachment_filter_helper.js
```diff
@@ -0,0 +1,36 @@
+import { $generateNodesFromDOM } from "@lexical/html"
+import { CustomActionTextAttachmentNode } from "../nodes/custom_action_text_attachment_node"
+import { ActionTextAttachmentNode } from "../nodes/action_text_attachment_node"
+
+export function $generateFilteredNodesFromDOM(editor, doc, editorElement) {
```
samuelpecher: I'm torn on this one: `editor` is available on `editorElement`, but I totally get matching the API of `$generateNodesFromDOM`

samuelpecher (approved):
Nice. Just a few thoughts but nothing blocking.

test/system/mention_round_trip_test.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  private
+    def assert_editor_value_has_mention_for(person)
+      value = Capybara.string(find_editor.value)
+      attachment = value.find(%(action-text-attachment[content-type="application/vnd.actiontext.mention"]))
+      assert attachment[:sgid].present?
+
+      mention_markup = CGI.unescapeHTML(attachment["content"])
+      mention_fragment = Capybara.string(mention_markup)
+      mention_fragment.assert_selector %(bc-mention[gid="#{person.to_gid}"]), text: person.name
```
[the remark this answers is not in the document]
flavorjones: The test reads `find_editor.value` (`test/test_helpers/editor_handler.rb:11-13`), which calls `this.value` on the `<lexxy-editor>` element — that is Lexical's `$generateHtmlFromNodes` export, not the raw server-emitted HTML. The server does JSON-encode the `content` attribute via `render_custom_attachments_in`, but the client-side import at `src/nodes/custom_action_text_attachment_node.js:37` routes through `parseAttachmentContent` (`src/helpers/storage_helper.js:15-21`), which tries `JSON.parse` first and falls back to raw — so `innerHtml` is stored unwrapped. Lexxy's `exportDOM` (`custom_action_text_attachment_node.js:102-110`) then emits `content: this.innerHtml` as raw HTML on the way out. Verified empirically (`JSON::ParserError` when attempting `JSON.parse(raw_content)`, no leading quote). `CGI.unescapeHTML` is correct here.

src/elements/editor.js
```diff
@@ -158,6 +161,22 @@ export class LexicalEditorElement extends HTMLElement {
...
+  permitsAttachmentContentType(contentType) {
+    if (this.getAttribute("attachments") === "false") return false
+    const list = this.#permittedAttachmentTypes
+    return list === null || list.includes(contentType)
+  }
+
+  #parsePermittedAttachmentTypes() {
+    const raw = this.dataset.permittedAttachmentTypes
```
jorgemanrubia: We own the custom element `lexxy-editor`. For first-class attributes we want to support there is no need to use a `data-` preffix. We could just use `permited-attachment-types` directly.

src/elements/editor.js
```diff
@@ -158,6 +161,22 @@ export class LexicalEditorElement extends HTMLElement {
     return this.dataset.blobUrlTemplate
   }
 
+  get permittedAttachmentTypes() {
+    return this.#permittedAttachmentTypes
```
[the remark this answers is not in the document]
flavorjones: Fixed in f8b9a775 by freezing the parsed list in `#parsePermittedAttachmentTypes`. External mutation attempts now throw at the call site; internal reads use `.includes` and remain unaffected. Pinned with a unit test at `test/javascript/unit/editor/permitted_attachment_types.test.js`.

src/elements/editor.js
```diff
@@ -158,6 +161,22 @@ export class LexicalEditorElement extends HTMLElement {
...
+  }
+
+  permitsAttachmentContentType(contentType) {
+    if (this.getAttribute("attachments") === "false") return false
```
jorgemanrubia: There is a method `supportsAttachments()` that also delegates the check to the config system (that can be tweaked via properties or via JS config object). I think that would be more suitable here?

src/elements/editor.js
```diff
@@ -158,6 +161,22 @@ export class LexicalEditorElement extends HTMLElement {
...
+    const list = this.#permittedAttachmentTypes
+    return list === null || list.includes(contentType)
+  }
+
+  #parsePermittedAttachmentTypes() {
+    const raw = this.dataset.permittedAttachmentTypes
+    if (raw == null) return null
```
jorgemanrubia: This is pure in-house style. I think we prefer to avoid guard returns and instead favor a explicit if/else conditional with `{}`:

```js
if () {

} else {

}
```

We probably don't honor this in many places, just because of how much Claude loves guard clauses 😅.

src/elements/editor.js
jorgemanrubia: Thinking about this further, we have a [configuration system](https://basecamp.github.io/lexxy/configuration.html) where you can configure properties via element properties or via JS options. Shouldn't we use that here?

jorgemanrubia (commented):
This looks fantastic @flavorjones

src/elements/editor.js
jorgemanrubia: Also: we should document this new configuration option in the docs.

src/helpers/attachment_filter_helper.js
flavorjones: Switched to `$descendantsMatching` in ba80bd34 — replaces the hand-rolled walker with the `@lexical/utils` helper, same behavior. Thanks!

test/browser/tests/attachments/import_filter.test.js
flavorjones: Right for Test 5 ("insert-editable-text prompt template with nested ") — that one exercises `ActionTextAttachmentNode`, which renders as `figure.attachment[data-content-type]`, so the `action-text-attachment` / `bc-attachment` selectors there were vacuous. Fixed in a6f7e823 to target `figure.attachment` and `img`.

The other `bc-attachment[content-type='...']` selectors in these tests flow through `CustomActionTextAttachmentNode`, whose `createDOM` emits `<bc-attachment content-type>` directly (`src/nodes/custom_action_text_attachment_node.js:75-84`). The positive `.toHaveCount(1)` assertions in the permitted-case tests prove those match real elements, so they are not vacuous — kept as-is.

src/helpers/attachment_filter_helper.js
flavorjones: Agreed — dropped in 500a2a1c. Signature is now `$generateFilteredNodesFromDOM(editorElement, doc)`, deriving `editor` from `editorElement.editor` internally. All four call sites updated.

src/elements/editor.js
flavorjones: I totally agree -- will fix all instances of guard clauses I introduced on this branch.

src/elements/editor.js
flavorjones: UGH. my original instructions were to use `permitted-attachment-types` and not the `data-` prefix ... that got lost somewhere in the hours of the planning phase. Good catch. Will fix it up.

test/browser/tests/attachments/import_filter.test.js
samuelpecher: Ah yeah, this was wrong place to attach the comment.

src/elements/editor.js
flavorjones: Absolutely. Will do.

src/elements/editor.js
flavorjones: Fixed across three commits:
- a7d4997e drops the `data-` prefix; the attribute is now `permitted-attachment-types`.
- 56bf8f10 promotes it to a first-class config option in the default preset, so it is settable via `Lexxy.configure({ default: { permittedAttachmentTypes: [...] } })` and presets, not just the element attribute.
- 129893e2 documents it under `docs/configuration.md` alongside the other editor options.

src/elements/editor.js
flavorjones: Fixed in 56bf8f10.

src/elements/editor.js
```diff
@@ -158,6 +161,28 @@ export class LexicalEditorElement extends HTMLElement {
...
+    return this.#permittedAttachmentTypes
+  }
+
+  permitsAttachmentContentType(contentType) {
+    if (this.getAttribute("attachments") === "false") {
+      return false
+    } else {
+      const list = this.#permittedAttachmentTypes
+      return list === null || list.includes(contentType)
+    }
```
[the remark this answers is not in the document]
flavorjones: Fixed in 56bf8f10.

src/elements/prompt.js
```diff
@@ -117,6 +119,19 @@ export class LexicalPromptElement extends HTMLElement {
...
+    const el = this.#editorElement
+    if (el.getAttribute("attachments") === "false") {
+      return false
+    } else {
+      const templates = Array.from(this.querySelectorAll("template[type='editor']"))
+      const types = templates.length
+        ? templates.map(t => t.getAttribute("content-type") || this.#defaultPromptContentType)
+        : [ this.#defaultPromptContentType ]
+      return types.some(t => el.permitsAttachmentContentType(t))
+    }
```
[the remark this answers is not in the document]
flavorjones: Fixed in 56bf8f10.

PR #1007, opened 2026-04-24
Fix broken code block styling for CLI-generated content

samuelpecher:
## Summary

- CLI tools and markdown converters produce `<pre><code class="language-ruby">` which differs from Lexxy's native `<pre data-language="ruby">` format
- This caused double styling (box-within-a-box from both `<pre>` and `<code>` getting background/padding) and no syntax highlighting (selector only matched `pre[data-language]`)
- Added `normalizeNestedCodeElement()` that extracts the language from `<code class="language-*">`, promotes it to `data-language` on `<pre>`, and unwraps the inner `<code>` element before Prism runs

Fixes [Basecamp card #9786748122](https://3.basecamp.com/2914079/buckets/41746046/card_tables/cards/9786748122): Code content generated in CLI break styles

### DIFF
--- a/src/helpers/code_highlighting_helper.js
+++ b/src/helpers/code_highlighting_helper.js
@@ -2,11 +2,37 @@ import { createElement } from "./html_helper"
 import Prism from "../config/prism"
 
 export function highlightCode() {
-  const elements = document.querySelectorAll("pre[data-language]")
+  const elements = document.querySelectorAll("pre[data-language], pre:has(> code)")
 
   elements.forEach(preElement => {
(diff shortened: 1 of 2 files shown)


samuelpecher:

Injected CLI-format code block HTML (`<pre><code class="language-ruby">`) into the page and ran `highlightCode()`. In production, the block was untouched — no syntax highlighting, nested `<code>` element retained (causing double styling). Locally with the fix, the nested `<code>` was normalized, language promoted to `data-language`, and Prism syntax highlighting applied correctly with token spans.

PR #1009, opened 2026-04-24
Fix pasting text with blank lines breaking out of code blocks

samuelpecher:
## Summary

- Pasting text with blank lines (two consecutive newlines) into a code block caused content to break out into separate paragraphs
- Root cause: during paste, Lexical calls `insertNewAfter()` for each newline; when a blank line creates an empty last line, `EarlyEscapeCodeNode`'s escape logic fires mid-paste, splitting content out
- Fix: added a `$hasUpdateTag(PASTE_TAG)` guard at the top of `insertNewAfter()` — during paste operations, delegate to the base `CodeNode` behavior which keeps content inside the block. Normal Enter-to-escape is unaffected.

Fixes [Basecamp card #9765146588](https://3.basecamp.com/2914079/buckets/41746046/card_tables/cards/9765146588): Pasting block w/ 2 newlines breaks out of codeblock

### DIFF
--- a/src/nodes/early_escape_code_node.js
+++ b/src/nodes/early_escape_code_node.js
@@ -1,4 +1,4 @@
-import { $createParagraphNode } from "lexical"
+import { $createParagraphNode, $hasUpdateTag, PASTE_TAG } from "lexical"
 import { CodeNode } from "@lexical/code"
 import { $getNearestNodeOfType } from "@lexical/utils"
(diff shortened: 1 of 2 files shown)


samuelpecher:

Created a code block, then pasted text containing blank lines ("line1\n\nline2\n\nline3") via clipboard API. In production, content after the first blank line broke out of the code block into separate `<p>` elements. Locally with the fix, all pasted lines remained inside the single `<pre>` block with `<br>` tags preserving the blank lines.

PR #1010, opened 2026-04-24
Fix code block copy/paste: preserve newlines and always paste as plain text

samuelpecher:
## Summary

- Copying a multi-line code block and pasting elsewhere collapsed all newlines into a single line
- Root cause: `highlightCode()` replaced the `<pre>` element with a bare `<code>` element — since `<code>` is inline, browsers collapse whitespace on copy
- Fix: changed `preElement.replaceWith(codeElement)` to `preElement.replaceChildren(codeElement)` to keep the `<pre>` wrapper

Additionally:
- Pasting into a code block now always uses plain text (`selection.insertRawText`), preventing HTML clipboard data from breaking code block structure
- Previously, paste-into-code-block was unhandled (deferred to Lexical's native handler which could interpret HTML and split the block)

Fixes [Basecamp card #9808822823](https://3.basecamp.com/2914079/buckets/41746046/card_tables/cards/9808822823): copy code block new lines are lost

### DIFF
--- a/src/editor/clipboard.js
+++ b/src/editor/clipboard.js
@@ -16,7 +16,13 @@ export default class Clipboard {
   paste(event) {
     const clipboardData = event.clipboardData
 
-    if (!clipboardData || this.#isPastingIntoCodeBlock()) return false
+    if (!clipboardData) return false
+
+    if (this.#isPastingIntoCodeBlock()) {
+      this.#pastePlainTextIntoCodeBlock(clipboardData)
+      event.preventDefault()
+      return true
+    }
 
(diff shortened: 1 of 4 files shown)


samuelpecher:

Injected a rendered code block with `data-language` and ran `highlightCode()`. In production, `highlightCode()` replaced the `<pre>` element with a bare `<code>` element — since `<code>` is inline, browsers collapse whitespace on copy, losing newlines. Locally with the fix, `highlightCode()` uses `replaceChildren` instead of `replaceWith`, keeping the `<pre>` wrapper intact so whitespace (and newlines) are preserved when copying.

PR #1029, opened 2026-04-30
Add dispose() lifecycle hook to extensions

flavorjones:
A new Extensions instance is created on every editor connectedCallback, but the old extension instances had no way to clean up: any listeners, inserted DOM, timers, or pending async state they attached during initializeToolbar() simply leaked. On Turbo back/forward navigation — where the toolbar DOM survives the disconnect/reconnect cycle — this manifested as duplicated handlers firing N times after N reconnects.

Add an optional dispose() method to LexxyExtension as a no-op so extensions can override it. Extensions.dispose() pops each enabled extension and calls dispose() in reverse initialization order, and the editor element registers `this.extensions` as a disposable so the hook fires during teardown alongside the editor, listeners, and toolbar.

Document the new lifecycle in docs/extensions.md and cover the reconnect behavior with a browser test that asserts dispose() is called once and that toolbar listeners do not duplicate after the editor element is detached and re-attached.

🤖 Assisted by Claude


samuelpecher (approved):
👌

PR #1035, opened 2026-05-04
Fix HTML injection via unsanitized empty-results attribute

rosa:
## Summary
- `LexicalPromptElement#showEmptyResults` assigned the `empty-results` attribute directly to `innerHTML`, rendering any HTML in the attribute into the popover. A host app that reflected user input into `empty-results` (e.g. `<lexxy-prompt empty-results="No match for <%= params[:q] %>">`) would have reflected DOM injection. Even with strict CSP, an attacker controlling the attribute could inject `<meta http-equiv="refresh">` for open redirect.
- Switch the assignment to `textContent`. The default `NOTHING_FOUND_DEFAULT_MESSAGE` is already plain text, and treating the attribute as text matches what the attribute name suggests.
- Same shape as #903 (Jorge's fix for the equivalent issue on `<action-text-attachment>` `content`).

## Test plan
- [x] `yarn lint` clean
- [x] `yarn test:browser --project=chromium test/browser/tests/prompts/` — all 21 tests pass, including the new regression
- [x] New regression test asserts the popover renders an HTML payload verbatim as text (no `` injection, no `onerror` execution)

Reported via HackerOne #3709373 by @blackm4c.

### DIFF
--- a/src/elements/prompt.js
+++ b/src/elements/prompt.js
@@ -366,7 +366,7 @@ export class LexicalPromptElement extends HTMLElement {
 
   #showEmptyResults() {
     this.popoverElement.classList.add("lexxy-prompt-menu--empty")
-    const el = createElement("li", { innerHTML: this.#emptyResultsMessage })
+    const el = createElement("li", { textContent: this.#emptyResultsMessage })
     el.classList.add("lexxy-prompt-menu__item--empty")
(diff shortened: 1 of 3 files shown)


flavorjones (approved):

samuelpecher (approved):
Thanks @rosa

PR #1037, opened 2026-05-05
Convert pasted data-URI images into uploads

flavorjones:
Convert data-URI images into uploads when content is pasted.

Pasted content from sources like Google Docs can include `` with the full image inlined. Without intervention, the base64 payload persists into saved HTML and bloats the stored document.

Extracted an editor function `$generateNodesFromDOM` which calls `$convertInlineImageDataURIs` when content has been pasted. That helper method decodes each `data:image/...` URI into a File, dispatches the cancelable `lexxy:file-accept` event, and replaces the attachment with an `ActionTextAttachmentUploadNode` that flows through the normal Active Storage pipeline.

Also extracted `$createUploadNode` for the inline image uri helper, but this simplifies the uploader as well.

See also previous work in Basecamp to implement this feature for Trix at https://github.com/basecamp/bc3/pull/9882

ref: https://app.basecamp.com/2914079/buckets/41746046/card_tables/cards/9849287511

### DIFF
--- a/src/editor/contents.js
+++ b/src/editor/contents.js
@@ -11,7 +11,7 @@ import { $createHeadingNode, $createQuoteNode, $isQuoteNode } from "@lexical/ric
 import { INSERT_ORDERED_LIST_COMMAND, INSERT_UNORDERED_LIST_COMMAND } from "@lexical/list"
 import { CustomActionTextAttachmentNode } from "../nodes/custom_action_text_attachment_node"
 import { $createLinkNode, $toggleLink } from "@lexical/link"
-import { dispatch, parseHtml } from "../helpers/html_helper"
+import { parseHtml } from "../helpers/html_helper"
 import { $forEachSelectedTextNode, $setBlocksType } from "@lexical/selection"
 import Uploader from "./contents/uploader"
 import { $isActionTextAttachmentNode } from "../nodes/action_text_attachment_node"
@@ -253,7 +253,7 @@ export default class Contents {
       console.warn("This editor does not supports attachments (it's configured with [attachments=false])")
       return
     }
-    const validFiles = Array.from(files).filter(this.#shouldUploadFile.bind(this))
+    const validFiles = Array.from(files).filter(file => this.editorElement.acceptsFile(file))
 
     this.editor.update(() => {
       const uploader = Uploader.for(this.editorElement, validFiles)
@@ -607,10 +607,6 @@ export default class Contents {
   #createHtmlNodeWith(html) {
     const htmlNodes = $generateFilteredNodesFromDOM(this.editorElement, parseHtml(html))
     return htmlNodes[0] || $createParagraphNode()
-  }
-
-  #shouldUploadFile(file) {
-    return dispatch(this.editorElement, "lexxy:file-accept", { file }, true)
   }
 
   // When the selection anchor is on a shadow root (e.g. a table cell), Lexical's
--- a/src/editor/contents.js
+++ b/src/editor/contents.js
@@ -15,7 +15,7 @@ import { parseHtml } from "../helpers/html_helper"
 import { $forEachSelectedTextNode, $setBlocksType } from "@lexical/selection"
 import Uploader from "./contents/uploader"
 import { $isActionTextAttachmentNode } from "../nodes/action_text_attachment_node"
-import { ActionTextAttachmentUploadNode } from "../nodes/action_text_attachment_upload_node"
+import { $createActionTextAttachmentUploadNode, ActionTextAttachmentUploadNode } from "../nodes/action_text_attachment_upload_node"
 import { $getNearestBlockElementAncestorOrThrow } from "@lexical/utils"
 import NodeInserter from "./contents/node_inserter"
 import { $isShadowRoot, $splitParagraphsAtLineBreakBoundaries } from "../helpers/lexical_helper"
@@ -264,6 +264,15 @@ export default class Contents {
         lastNode.selectEnd()
         this.#normalizeSelectionInShadowRoot()
       }
+    })
+  }
+
+  $createUploadNode(file) {
+    return $createActionTextAttachmentUploadNode({
+      file,
+      uploadUrl: this.editorElement.directUploadUrl,
+      blobUrlTemplate: this.editorElement.blobUrlTemplate,
+      contentType: file.type,
     })
   }
 
--- a/src/editor/contents.js
+++ b/src/editor/contents.js
@@ -5,6 +5,7 @@ import {
 } from "lexical"
 
 import { $generateFilteredNodesFromDOM } from "../helpers/attachment_filter_helper"
+import { $convertInlineImageDataURIs } from "../helpers/inline_image_uri_helper"
 import { $createCodeNode, $isCodeNode } from "@lexical/code"
 import { $createHeadingNode, $createQuoteNode, $isQuoteNode } from "@lexical/rich-text"
 import { INSERT_ORDERED_LIST_COMMAND, INSERT_UNORDERED_LIST_COMMAND } from "@lexical/list"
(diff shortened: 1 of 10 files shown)


src/helpers/inline_image_uri_helper.js
```diff
@@ -0,0 +1,70 @@
...
+}
+
+function isInlineImageDataURIAttachment(node) {
+  return node instanceof ActionTextAttachmentNode &&
+    /^data:image\//i.test(node.src ?? "")
+}
+
+function $tryConvertToUpload(node, editorElement) {
+  const file = dataURIToFile(node.src)
+  if (file && dispatch(editorElement, "lexxy:file-accept", { file }, true)) {
```
samuelpecher: Dispatching the event from a helper makes me feel like an `editorElement.acceptFile(file)` handler is missing

src/helpers/inline_image_uri_helper.js
```diff
@@ -0,0 +1,70 @@
...
+    return $createActionTextAttachmentUploadNode({
+      file,
+      uploadUrl: editorElement.directUploadUrl,
+      blobUrlTemplate: editorElement.blobUrlTemplate,
+      contentType: file.type,
+    })
```
samuelpecher: This also feels like reaching deep into the editor where `editor.contents.$createUploadNode(file)` might feel better

src/helpers/inline_image_uri_helper.js
```diff
@@ -0,0 +1,70 @@
...
+      .forEach(desc => {
+        const upload = $tryConvertToUpload(desc, editorElement)
+        if (upload) {
+          desc.replace(upload)
+        } else {
+          desc.remove()
+        }
+      })
+    return [ node ]
+  })
```
samuelpecher: Took me a minute to figure out this section 😅

I think a construction like this would be clearer

```js
    const replacedNodes = nodes
      .map(node => isInlineImageDataURIAttachment(node) ? $tryCreateUploadFromDataURI(node) : node)
      .filter(node => node !== null)

    replacedNodes.forEach(node => {
      for (const desc of $descendantsMatching([ node ], isInlineImageDataURIAttachment)) {
        const upload = $tryCreateUploadFromDataURI(desc, editorElement)
        if (upload) {
          desc.replace(upload)
        } else {
          desc.remove()
        }
      })
    }
  })

  return replacedNodes
```

src/helpers/inline_image_uri_helper.js
```diff
@@ -0,0 +1,70 @@
...
+}
+
+function isInlineImageDataURIAttachment(node) {
+  return node instanceof ActionTextAttachmentNode &&
+    /^data:image\//i.test(node.src ?? "")
+}
+
+function $tryConvertToUpload(node, editorElement) {
```
samuelpecher: The name suggests to me a replacement has already happened

```suggestion
function $tryCreateUploadFromDataURI(node, editorElement) {
```

src/helpers/dom_import_helper.js
```diff
@@ -0,0 +1,9 @@
+import { $generateNodesFromDOM } from "@lexical/html"
+import { $convertInlineImageDataURIs } from "./inline_image_uri_helper"
+import { $filterDisallowedAttachmentNodes } from "./attachment_filter_helper"
+
+export function $importNodesFromDOM(editorElement, doc) {
```
samuelpecher: To me, `import` suggests the entire conversion + insertion flow. A judo might be to attach this to the element: `editorElement.$generateNodesFromDOM`, which also removes the need to pass-in the editor element.

Alternatively, perhaps a helper object? `generator = new NodeGenerator(editorElement); generator.nodesFromDOM(doc)`, though it feels like it would want to be attached to `EditorElement` anyway...

samuelpecher (commented):
Nice; just had a few ideas on naming/flow

flavorjones:
@samuelpecher I tried to tell you not to review! I've totally overhauled this PR. Sorry.

src/helpers/dom_import_helper.js
flavorjones: This helper was removed.

src/helpers/inline_image_uri_helper.js
flavorjones: Renamed.

src/helpers/inline_image_uri_helper.js
flavorjones: Restructured along these lines — top-level handling (map + filter) and descendant handling (replace/remove in place) are now separate passes. Took the opportunity to also make `$tryCreateUploadFromDataURI` explicitly `return null` on the failure path so the filter is unambiguous.

src/helpers/inline_image_uri_helper.js
flavorjones: Extracted as `editorElement.acceptsFile(file)` (predicate naming, parallel to `permitsAttachmentContentType`). Both call sites now go through it — `inline_image_uri_helper.js` and the file-upload filter in `Contents#uploadFiles` (the old `#shouldUploadFile` helper was inlined and removed).

src/helpers/inline_image_uri_helper.js
flavorjones: Added `Contents#$createUploadNode(file)` and migrated both this helper and `Uploader#$createUploadNodes` to use it. The helper no longer reaches into `directUploadUrl`/`blobUrlTemplate` or imports the node factory — just calls `editorElement.contents.$createUploadNode(file)`. Also let me drop the now-redundant `#nodeUrlProperties` getter from `Uploader`.

src/helpers/inline_image_uri_helper.js
```diff
@@ -0,0 +1,64 @@
...
+function dataURIToFile(dataURI) {
+  try {
+    const [ header, data ] = dataURI.split(",")
+
+    // https://datatracker.ietf.org/doc/html/rfc6838#section-4.2
+    const mimeType = header.match(/^data:(image\/[A-Za-z0-9][A-Za-z0-9!#$&\-^_.+]*)/)?.[1]
+    if (mimeType) {
+      const bytes = Uint8Array.from(atob(data), (c) => c.charCodeAt(0))
+      const extension = mimeTypeToExtension(mimeType) ?? "png"
+      return new File([ bytes ], `pasted-image-${Date.now()}.${extension}`, { type: mimeType })
```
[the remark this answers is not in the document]
flavorjones: Out of scope for this PR.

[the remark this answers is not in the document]
flavorjones: I'm OK with this for now.

src/editor/contents.js
```diff
@@ -42,7 +43,8 @@ export default class Contents {
     this.editor.update(() => {
       if ($hasUpdateTag(PASTE_TAG)) this.#stripTableCellColorStyles(doc)
 
-      const nodes = $generateFilteredNodesFromDOM(this.editorElement, doc)
+      let nodes = $generateFilteredNodesFromDOM(this.editorElement, doc)
+      if ($hasUpdateTag(PASTE_TAG)) nodes = $convertInlineImageDataURIs(nodes, this.editorElement)
       if (!this.#insertUploadNodes(nodes)) {
```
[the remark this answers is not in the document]
flavorjones: fixed

src/helpers/inline_image_uri_helper.js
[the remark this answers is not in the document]
flavorjones: Fixed in d36a9226 — tightened `isInlineImageDataURIAttachment` to require `;base64,` in the predicate, so non-base64 image data URIs (e.g. `data:image/svg+xml;charset=utf-8,<svg.../>`) fall through and are left intact instead of being silently dropped when `atob` throws. Note: `;utf8` from your example isn't actually valid per RFC 2397 (parameters require `=`), but `;charset=utf-8` and the bare form are both spec-compliant — added a regression test covering the bare form.

PR #1039, opened 2026-05-06
Prevent prompt picker from scrolling the page when opened

alpkeser:
On mobile apps with the soft keyboard up, typing @ in a Lexxy editor whose cursor sat near the bottom of the visible area caused the entire page to scroll up. In some cases the picker was placed in the keyboard-covered area and clipped.

1. lexxy — give it a way to know about the keyboard
#viewportBottom now reads, in order:
--lexxy-host-viewport-height (+ --lexxy-host-viewport-top) CSS variables on <html> — the host-supplied authoritative number,
window.visualViewport.height — works for normal mobile Safari and most other browsers,
window.innerHeight — last-resort fallback.
This makes the clipping check correct, so the popover gets flipped above the cursor when the visible area below is too small.

2. lexxy — stop the picker from scrolling the page on open
In #selectOption:
Removed the unconditional listItem.scrollIntoView from the initial-open path. It now runs only on arrow-key navigation, scoped to the popover's own scroll container via popover.scrollTo — physically can't scroll the page.
listItem.focus({ preventScroll: true }).
Focus the contenteditable directly with editorContentElement.focus({ preventScroll: true }) — going through LexicalEditorElement.focus() would have silently dropped the option (the wrapper takes no parameters; editor.focus(callback) is Lexical's signature, not DOM focus(options)).

### DIFF
--- a/src/editor/selection.js
+++ b/src/editor/selection.js
@@ -1,7 +1,7 @@
 import {
   $addUpdateTag, $createParagraphNode, $getNearestNodeFromDOMNode, $getRoot, $getSelection, $isDecoratorNode, $isElementNode,
   $isLineBreakNode, $isNodeSelection, $isRangeSelection, $isTextNode, $setSelection, CLICK_COMMAND, COMMAND_PRIORITY_LOW, DELETE_CHARACTER_COMMAND,
-  HISTORY_MERGE_TAG, KEY_ARROW_DOWN_COMMAND, KEY_ARROW_LEFT_COMMAND, KEY_ARROW_RIGHT_COMMAND, KEY_ARROW_UP_COMMAND, SELECTION_CHANGE_COMMAND, isDOMNode
+  HISTORY_MERGE_TAG, KEY_ARROW_DOWN_COMMAND, KEY_ARROW_LEFT_COMMAND, KEY_ARROW_RIGHT_COMMAND, KEY_ARROW_UP_COMMAND, SELECTION_CHANGE_COMMAND, SKIP_SCROLL_INTO_VIEW_TAG, isDOMNode
 } from "lexical"
 import { $getNearestNodeOfType } from "@lexical/utils"
 import { $getListDepth, ListItemNode, ListNode } from "@lexical/list"
@@ -87,7 +87,7 @@ export default class Selection {
     return { node: null, offset: 0 }
   }
 
-  preservingSelection(fn) {
+  preservingSelection(fn, { preventScroll = false } = {}) {
     let selectionState = null
 
     this.editor.getEditorState().read(() => {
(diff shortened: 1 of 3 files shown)


src/editor/selection.js
```diff
@@ -108,6 +108,7 @@ export default class Selection {
         if (selection && $isRangeSelection(selection)) {
           selection.anchor.set(selectionState.anchor.key, selectionState.anchor.offset, "text")
           selection.focus.set(selectionState.focus.key, selectionState.focus.offset, "text")
+          if (preventScroll) $addUpdateTag(SKIP_SCROLL_INTO_VIEW_TAG)
```
samuelpecher, line 111: I'd put this at the top of the block

```diff
        if (selection && $isRangeSelection(selection)) {
+        if (preventScroll) $addUpdateTag(SKIP_SCROLL_INTO_VIEW_TAG)
        
          selection.anchor.set(selectionState.anchor.key, selectionState.anchor.offset, "text")
          selection.focus.set(selectionState.focus.key, selectionState.focus.offset, "text")      
```

src/elements/prompt.js
```diff
@@ -238,22 +238,36 @@ export class LexicalPromptElement extends HTMLElement {
...
-    // Preserve selection to prevent cursor jump
+    listItem.focus({ preventScroll: true })
+
+    // Focus the contenteditable directly — the editor wrapper's focus()
+    // drops options.
     this.#selection.preservingSelection(() => {
-      this.#editorElement.focus()
-    })
+      this.#editorContentElement.focus({ preventScroll: true })
+    }, { preventScroll: true })
```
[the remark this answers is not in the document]
samuelpecher: I'm experiencing this in testing as well.

src/elements/prompt.js
```diff
@@ -238,22 +238,36 @@ export class LexicalPromptElement extends HTMLElement {
     return Array.from(this.popoverElement.querySelectorAll(".lexxy-prompt-menu__item"))
   }
 
-  #selectOption(listItem) {
+  #selectOption(listItem, { scrollIntoView = false } = {}) {
```
samuelpecher, line 241: I think the right default here is `true`, since the exception is to not scroll when selecting rather than preventing scroll when moving up/down

```suggestion
  #selectOption(listItem, { scrollIntoView = true } = {}) {
```

src/elements/prompt.js
```diff
@@ -238,22 +238,36 @@ export class LexicalPromptElement extends HTMLElement {
...
     this.#clearListItemSelection()
     listItem.toggleAttribute("aria-selected", true)
-    listItem.scrollIntoView({ block: "nearest", behavior: "smooth" })
-    listItem.focus()
+    if (scrollIntoView) this.#scrollItemIntoPopover(listItem)
```
samuelpecher, line 244: Could the custom function's aims be achieved with `container: "nearest"` to scroll only the [nearest scrollable container](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView#container) and avoid scrolling the viewport?

samuelpecher (commented):
Thanks @alpkeser - I think we might be able to simplify with an option on `scrollIntoView`. I don't have an easy way to test this gives us the native behavior we're after, let me know if it's not the target behavior.

PR #1040, opened 2026-05-06
Recalculate toolbar overflow in extensions

zoltanhosszu:
This PR fixes a toolbar overflow issue that happened when an extension injected a toolbar element after the toolbar overflow was calculated. Using `initializeToolbars()` in extensions now triggers a recalculation, fixing the overflow issue.

image

### Changes

- `requestOverflowRefresh()` is now a public prototype method; it resets, reindexes all toolbar items by current DOM order (`#reindexToolbarItems`), then compacts. The standalone `#setItemPositionValues` call in setEditor is gone since refresh handles it.
- Extension's `initializeToolbars()` calls `toolbar.requestOverflowRefresh()` after extension buttons are added.
- New test exercising async injection + the public `requestOverflowRefresh()` API, asserting reindex + correct overflow placement + restoration on resize.
- Also fixing a small visual issue that occurred because the overflow logic didn't account for the button margins when calculating the right edge.
- Removed `.lexxy-editor__toolbar-spacer`, right aligned buttons are now relying on `margin-inline-start: auto`


test-results/.last-run.json
samuelpecher, line 1: :grimacing:

src/elements/toolbar.js
```diff
@@ -25,7 +25,7 @@ export class LexicalToolbarElement extends HTMLElement {
   }
 
   connectedCallback() {
-    requestAnimationFrame(() => this.#refreshToolbarOverflow())
+    requestAnimationFrame(() => this.refreshOverflow())
```
samuelpecher: Since this is open, it would be ideal to push the rAF down into `refreshOverflow()`, since it should always be deferred to render and callers shouldn't have to handle that quirk.

I wouldn't yet bother with ensuring it only occurs once per frame unless you're keen.

samuelpecher (changes requested):

src/elements/toolbar.js
```diff
@@ -448,17 +454,15 @@ export class LexicalToolbarElement extends HTMLElement {
...
+      <button class="lexxy-editor__toolbar-button lexxy-editor__toolbar-button--push-right" type="button" name="undo" data-command="undo" title="Undo" disabled aria-disabled="true">
         ${ToolbarIcons.undo}
       </button>
 
       <button class="lexxy-editor__toolbar-button" type="button" name="redo" data-command="redo" title="Redo" disabled aria-disabled="true">
         ${ToolbarIcons.redo}
       </button>
 
-      <lexxy-toolbar-dropdown class="lexxy-editor__toolbar-dropdown lexxy-editor__toolbar-overflow">
+      <lexxy-toolbar-dropdown class="lexxy-editor__toolbar-dropdown lexxy-editor__toolbar-button--push-right lexxy-editor__toolbar-overflow">
```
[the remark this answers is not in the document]
zoltanhosszu: This does not happen as the overflow is only visible when undo is hidden in it already...

src/elements/toolbar.js
```diff
@@ -251,29 +257,43 @@ export class LexicalToolbarElement extends HTMLElement {
...
 
-  #refreshToolbarOverflow = () => {
-    this.#resetToolbarOverflow()
-    this.#compactMenu()
+  // Coalesces concurrent calls into a single per-frame pass. Init fires this from
+  // connectedCallback, setEditor, extensions, and the ResizeObserver in quick
+  // succession; without coalescing, each would run its own reset/reindex/compact
+  // on the same frame.
+  async #runRefreshOverflow() {
+    try {
```
samuelpecher: It's not clear why this needs `try {} finally`; is there an intended failure?

src/elements/toolbar.js
```diff
@@ -251,29 +257,43 @@ export class LexicalToolbarElement extends HTMLElement {
...
+      await nextFrame()
```
samuelpecher: This can go if the operation is wrapped in `rAF`

```suggestion
```

src/elements/toolbar.js
```diff
@@ -251,29 +257,43 @@ export class LexicalToolbarElement extends HTMLElement {
...
   // buttons overflow using math, and then moves them in a single write pass.
-  // The previous implementation interleaved `scrollWidth`/`clientWidth` reads with
-  // `prepend()` writes inside a loop, forcing one full browser reflow per button.
   #compactMenu() {
     const buttons = this.#overflowButtons
     if (buttons.length === 0) return
 
-    const availableWidth = this.clientWidth + 1 // +1 for Safari zoom rounding
-    const buttonRightEdges = buttons.map(button => button.offsetLeft + button.offsetWidth)
+    const availableWidth = this.clientWidth
```
samuelpecher, line 288: No more Safari zoom rounding?

src/elements/toolbar.js
```diff
@@ -80,6 +81,11 @@ export class LexicalToolbarElement extends HTMLElement {
     return this.editorElement || await this.editorPromise
   }
 
+  refreshOverflow() {
+    this.#inFlightOverflowRefresh ??= this.#runRefreshOverflow()
+    return this.#inFlightOverflowRefresh
+  }
+
```
samuelpecher: I'd keep the setting and clearing of the flag together and use the returned AF as the marker.

```js
refreshOverflow() {
  this.#refreshToolbarAF ??= requestAnimationFrame(() => {
    this.#refreshOverflow()
    this.#refreshToolbarAF = null
  }
}
```

samuelpecher (commented):
Nice, just a few bits to tighten up

src/elements/toolbar.js
zoltanhosszu: With this new calculation, it is not needed anymore, at least as far as I was able to test it!

src/elements/toolbar.js
```diff
@@ -80,6 +80,15 @@ export class LexicalToolbarElement extends HTMLElement {
...
   }
 
+  requestOverflowRefresh() {
+    if (this.#refreshToolbarAF != null) return
+
+    this.#refreshToolbarAF = requestAnimationFrame(() => {
+      this.#refreshOverflow()
+      this.#refreshToolbarAF = null
+    })
+  }
```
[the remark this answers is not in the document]
samuelpecher: Swallowing errors is undesirable behavior

PR #1044, opened 2026-05-07
Defer initialization to the next rAF

monorkin:
Immediate initialization can sometimes cause lag spikes if the parent element is being morphed at the same time. This PR moves initialization to the next animation frame to give the DOM time to flush the pending layout before Lexical attempts to compute geometry.

The root cause of the lag is a forced layout flush caused by Lexical. It has a selection handler that calls `Selection.anchorNode` during `setRootElement`. This forces the browser to immediately flush the layout, persiste pending mutations, and perform style invalidation, all of which pushes the paint back causing causing a visual hang.

This often triggers on navigation with Turbo morph enabled. As morph updates the DOM, Lexxy initializes, and forces the browser to flush all the changes and perform style invalidation, causing the navigation to lag by 50-100ms.

By deferring initialization, navigation is able to complete as normal, the page is laid out and painted, and then Lexxy initializes, and does so much quicker as it doesn't have to perform additional layouts.

### DIFF
--- a/src/editor/command_dispatcher.js
+++ b/src/editor/command_dispatcher.js
@@ -326,7 +326,7 @@ export class CommandDispatcher {
   #registerDragAndDropHandlers() {
     if (this.editorElement.supportsAttachments) {
       this.dragCounter = 0
-      const root = this.editor.getRootElement()
+      const root = this.editorElement.editorContentElement
       this.#listeners.track(
         registerEventListener(root, "dragover", this.#handleDragOver.bind(this)),
         registerEventListener(root, "drop", this.#handleDrop.bind(this)),
--- a/src/editor/command_dispatcher.js
+++ b/src/editor/command_dispatcher.js
@@ -342,7 +342,7 @@ export class CommandDispatcher {
     this.dragCounter++
     if (this.dragCounter === 1) {
       this.#saveSelectionBeforeDrag()
-      this.editor.getRootElement().classList.add("lexxy-editor--drag-over")
+      this.editorElement.editorContentElement.classList.add("lexxy-editor--drag-over")
     }
   }
 
@@ -352,7 +352,7 @@ export class CommandDispatcher {
     this.dragCounter--
     if (this.dragCounter === 0) {
       this.#selectionBeforeDrag = null
-      this.editor.getRootElement().classList.remove("lexxy-editor--drag-over")
+      this.editorElement.editorContentElement.classList.remove("lexxy-editor--drag-over")
     }
   }
 
@@ -368,7 +368,7 @@ export class CommandDispatcher {
     event.preventDefault()
 
     this.dragCounter = 0
-    this.editor.getRootElement().classList.remove("lexxy-editor--drag-over")
+    this.editorElement.editorContentElement.classList.remove("lexxy-editor--drag-over")
 
     const dataTransfer = event.dataTransfer
     if (!dataTransfer) return
--- a/src/editor/command_dispatcher.js
+++ b/src/editor/command_dispatcher.js
@@ -326,12 +326,18 @@ export class CommandDispatcher {
   #registerDragAndDropHandlers() {
     if (this.editorElement.supportsAttachments) {
       this.dragCounter = 0
-      const root = this.editorElement.editorContentElement
       this.#listeners.track(
-        registerEventListener(root, "dragover", this.#handleDragOver.bind(this)),
-        registerEventListener(root, "drop", this.#handleDrop.bind(this)),
-        registerEventListener(root, "dragenter", this.#handleDragEnter.bind(this)),
-        registerEventListener(root, "dragleave", this.#handleDragLeave.bind(this))
+        this.editor.registerRootListener((rootElement) => {
+          if (rootElement) {
+            const teardowns = [
+              registerEventListener(rootElement, "dragover", this.#handleDragOver.bind(this)),
+              registerEventListener(rootElement, "drop", this.#handleDrop.bind(this)),
+              registerEventListener(rootElement, "dragenter", this.#handleDragEnter.bind(this)),
+              registerEventListener(rootElement, "dragleave", this.#handleDragLeave.bind(this))
+            ]
+            return () => teardowns.forEach((teardown) => teardown())
+          }
+        })
       )
     }
   }
@@ -342,7 +348,7 @@ export class CommandDispatcher {
     this.dragCounter++
     if (this.dragCounter === 1) {
       this.#saveSelectionBeforeDrag()
-      this.editorElement.editorContentElement.classList.add("lexxy-editor--drag-over")
+      this.editor.getRootElement().classList.add("lexxy-editor--drag-over")
     }
   }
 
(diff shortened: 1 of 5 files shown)


src/elements/editor.js
```diff
@@ -88,10 +88,20 @@ export class LexicalEditorElement extends HTMLElement {
 
     this.#initialize()
 
-    this.#scheduleEditorInitializedDispatch()
     this.toggleAttribute("connected", true)
 
-    this.#handleAutofocus()
+    if (this.hasAttribute("defer")) {
+      requestAnimationFrame(() => {
+        if (!this.isConnected) return
```
jorgemanrubia: I don't love these early returns, I would just use the positive condition branch explicitly: `if (this.connected)...`

jorgemanrubia: Also do we need this isConnected check in this branch at all?

monorkin: Oh, I thought because it was the first in the callback it would be fine.
That guards against stale fires of the callback - the editor connecting and disconnecting before the callback fires.

After the last commit that addresses the rAF leak that Copilot raised we no longer need this. Removing.

src/elements/editor.js
```diff
@@ -88,15 +89,28 @@ export class LexicalEditorElement extends HTMLElement {
...
+      })
+    } else {
+      this.#scheduleEditorInitializedDispatch()
+      this.#handleAutofocus()
+      this.valueBeforeDisconnect = null
+    }
   }
 
   disconnectedCallback() {
+    this.#cancelDeferredInit()
```
jorgemanrubia: I think this could be too defensive (cancel the animation frame on disconnect)

monorkin: Removed it. I though it would be semantically cleaner - everything that's in-flight is stopped on disconnect.

src/elements/editor.js
```diff
@@ -88,15 +89,26 @@ export class LexicalEditorElement extends HTMLElement {
...
+        this.#loadInitialValue()
+        this.#handleAutofocus()
+        this.#scheduleEditorInitializedDispatch()
+        this.valueBeforeDisconnect = null
+      })
+    } else {
+      this.#scheduleEditorInitializedDispatch()
+      this.#handleAutofocus()
+      this.valueBeforeDisconnect = null
+    }
```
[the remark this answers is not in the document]
monorkin: Removed the attribute completely, deferred init is only init path now.

src/elements/editor.js
```diff
@@ -88,15 +89,26 @@ export class LexicalEditorElement extends HTMLElement {
...
   }
 
   disconnectedCallback() {
+    this.#cancelDeferredInit()
     this.#cancelEditorInitializedDispatch()
     this.valueBeforeDisconnect = this.value
```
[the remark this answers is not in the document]
monorkin: The value is now set synchronously.

src/elements/editor.js
```diff
@@ -88,15 +89,26 @@ export class LexicalEditorElement extends HTMLElement {
...
-
-    this.valueBeforeDisconnect = null
+    if (this.hasAttribute("defer")) {
+      this.#deferredInitRafId = requestAnimationFrame(() => {
+        this.#deferredInitRafId = null
+        this.editor.setRootElement(this.editorContentElement)
+        this.#loadInitialValue()
+        this.#handleAutofocus()
+        this.#scheduleEditorInitializedDispatch()
+        this.valueBeforeDisconnect = null
```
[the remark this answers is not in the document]
monorkin: The defer path is now the only init path

src/elements/editor.js
```diff
@@ -88,10 +88,13 @@ export class LexicalEditorElement extends HTMLElement {
...
 
-    this.#scheduleEditorInitializedDispatch()
     this.toggleAttribute("connected", true)
 
-    this.#handleAutofocus()
+    requestAnimationFrame(() => {
+      this.editor.setRootElement(this.editorContentElement)
+      this.#handleAutofocus()
+      this.#scheduleEditorInitializedDispatch()
+    })
```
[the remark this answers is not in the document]
monorkin: Discussed as an acceptable tradeoff. There is an internal check that the root elements match which prevents the problems mentioned here.

src/editor/command_dispatcher.js
```diff
@@ -326,7 +326,7 @@ export class CommandDispatcher {
   #registerDragAndDropHandlers() {
     if (this.editorElement.supportsAttachments) {
       this.dragCounter = 0
-      const root = this.editor.getRootElement()
+      const root = this.editorElement.editorContentElement
```
samuelpecher: This gives me pause. Holding the reference to the content element separately from Lexical breaks the concepts of observing Lexical's `RootElement`.

I think the right move is to defer the registration of listeners by [registering a root element listener](https://lexical.dev/docs/concepts/listeners#registerrootlistener), where Lexical will provide the listener upon registration. That will ensure the listeners are attached to the linked root element and the timing of listener registration is all controlled through `setRootElement`.

So roughly:
```js

#registerSomeDomHandlers() {
  this.#listeners.track(this.editor.registerRootElementListener(({ rootElement }) => {
    // attach handlers to root element
  }))
}

```

src/elements/editor.js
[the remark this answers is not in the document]
samuelpecher: I think this could be partially allayed with [registerRootListener](https://lexical.dev/docs/concepts/listeners#registerrootlistener)

samuelpecher (commented):
Great find! I've just got one suggestion about hooking into Lexical's root element registration rather than holding the reference on the element.

monorkin:
Putting this on hold for a bit. 

I did an extensive test yesterday and this seems to only help in the worst-case scenario. Something else is causing the calculations to hang.

This change might not be wroth the performance gain it brings.

samuelpecher (approved):
@monorkin I think initializing event listeners via a root listener is worth it on it's own, and performance gains at the worst end of the tail are the most appreciated. Unless this causes us other issues I'd be for a merge.

@jorgemanrubia would you concur?

src/elements/toolbar.js
```diff
@@ -68,7 +68,7 @@ export class LexicalToolbarElement extends HTMLElement {
     this.#setItemPositionValues()
     this.#monitorSelectionChanges()
     this.#monitorHistoryChanges()
-    this.#refreshToolbarOverflow()
```
samuelpecher: This is coming in [#1040](https://github.com/basecamp/lexxy/pull/1040/files#diff-6fa89d2c0ed6d78ab9585e522cbee4494670ce43e86a12b5b074e1b9f39204c3)

monorkin: Oh great! I'll pop my last commit.

src/editor/selection.js
```diff
@@ -377,46 +377,59 @@ export default class Selection {
...
-
-          if (!currentElement) {
-            event.preventDefault()
-          }
+    this.#listeners.track(
+      this.editor.registerRootListener((rootElement) => {
+        if (rootElement) {
+          const handler = (event) => this.#handleArrowKeyOnLexicalCursor(event)
+          rootElement.addEventListener("keydown", handler, true)
+          return () => rootElement.removeEventListener("keydown", handler, true)
```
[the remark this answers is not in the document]
samuelpecher: https://github.com/facebook/lexical/blob/main/packages/lexical/src/LexicalEditor.ts#L1071-L1071

PR #1047, opened 2026-05-08
Export Lexical and document allowedElements in Extensions

excid3:
This addresses a couple things:

1. The docs previously showed importing `defineExtension` from lexical. However, that means we need to pin the lexical package which could be a different version. Instead, we can export Lexical through Lexxy so users have access to the exact same version.

We could export just `defineExtension` to keep it minimal, but I wasn't sure if extensions may need access to other parts of Lexical. I can quickly refactor this if that's the best approach.

2. I am unfurling links with Stimulus and allowing iframe embeds for YouTube, Spotify, etc. This broke recently because of #909 I believe (it worked in 0.9.3.beta but not 0.9.4.beta+), which lead me to trying to write an extension to allow iframe tags. Trying to create an extension lead me first to trying to access defineExtension cleanly.

Does this make sense as an approach @zachasme @samuelpecher @jorgemanrubia?


samuelpecher:
Hey @excid3 :wave: I'm for this in principle. My one concern about package size was unfounded.

I've been asking myself the same question since our extensions make use of commands, tags, and `mergeRegister`. Lexical helpfully exports all of these at the top level.

I'm not a huge fan of `return Lexxy.Lexical.defineExtension(...`, perhaps `Lexxy.Extension` should provide a `defineLexicalExtension` wrapper for this common use case.

For other imports, I wonder whether exporting lexical at a different path (like `highlightCode`) looks better. 
```js
import { mergeRegister } from "@37signals/lexxy/lexical"
```
rather than:
```js
import Lexxy from "@37signals/lexxy"
const { mergeRegister } = Lexxy.Lexical
// or at the callsite:
Lexxy.Lexical.mergeRegister(//...
```


My only concern would have been on bundle size, but it looks like the hit is ~6k and there won't be any impact on the npm package.

---

Without export
```bash
lexxy main  ❯ yarn build && ls -l app/assets/javascript/
yarn run v1.22.22
$ rollup -c

./src/index.js → ./app/assets/javascript/lexxy.js, ./app/assets/javascript/lexxy.min.js...
created ./app/assets/javascript/lexxy.js, ./app/assets/javascript/lexxy.min.js in 3.2s
Done in 3.36s.
Permissions Size User Date Modified Name
.rw-r--r--  855k slp  11 May 12:09   lexxy.js
.rw-r--r--  195k slp  11 May 12:09   lexxy.js.br
.rw-r--r--  238k slp  11 May 12:09   lexxy.js.gz
.rw-r--r--  2.0M slp  11 May 12:09   lexxy.js.map
.rw-r--r--  572k slp  11 May 12:09   lexxy.min.js
.rw-r--r--  146k slp  11 May 12:09   lexxy.min.js.br
.rw-r--r--  175k slp  11 May 12:09   lexxy.min.js.gz
```

With export
```bash
lexxy main  ❯ yarn build && ls -l app/assets/javascript/
yarn run v1.22.22
$ rollup -c

./src/index.js → ./app/assets/javascript/lexxy.js, ./app/assets/javascript/lexxy.min.js...
created ./app/assets/javascript/lexxy.js, ./app/assets/javascript/lexxy.min.js in 3s
Done in 3.21s.
Permissions Size User Date Modified Name
.rw-r--r--  861k slp  11 May 12:10   lexxy.js
.rw-r--r--  197k slp  11 May 12:10   lexxy.js.br
.rw-r--r--  240k slp  11 May 12:10   lexxy.js.gz
.rw-r--r--  2.0M slp  11 May 12:10   lexxy.js.map
.rw-r--r--  578k slp  11 May 12:10   lexxy.min.js
.rw-r--r--  148k slp  11 May 12:10   lexxy.min.js.br
.rw-r--r--  177k slp  11 May 12:10   lexxy.min.js.gz
```

excid3:
@samuelpecher that's not as much as I expected!

I'm not super familiar with Lexical, but my thinking was by exposing all of Lexical, you give people the freedom to access any functionality easily. That may not be necessary or common enough, so a wrapper could be a simpler option for users.

jorgemanrubia (commented):
Hey @excid3 could you detail the limitations you found with the current extension support?

I would like to keep Lexical encapsulated as much as possible and favor an extension system that covers most needs, unless we can make a strong case to give full access to Lexical.

Thanks!

excid3:
@jorgemanrubia there didn't seem to be a way to create a Lexxy extension using importmaps without also pinning Lexical separately. Could just be my lack of frontend knowledge.

The docs show:

```ruby
pin "lexxy", to: "lexxy.js"
```

```js
import { defineExtension } from "lexical"
import * as Lexxy from "@37signals/lexxy"
```

I was trying to build an extension that lets me paste in like a YouTube or Bunny.net video and allow the iframe tag.

```js
import * as Lexical from "lexical"

class EmbedsExtension extends Lexxy.Extension {
  get lexicalExtension() {
    return Lexical.defineExtension({ name: "lexxy/embeds_extension" })
  }

  get allowedElements() {
    return [ { tag: "iframe", attributes: [ "src", "loading", "width", "height", "style", "allow" ] } ]
  }
}

Lexxy.configure({
  global: {
    extensions: [EmbedsExtension],
  }
})
```

The trouble is I'd have to pin a second copy of Lexical in order to do this currently?

samuelpecher:
> by exposing all of Lexical, you give people the freedom to access any functionality easily. 

Yep. Lexical's functions are the "knives" we need to expose for extension writers to cut into Lexical.

---

@jorgemanrubia I think the import for our own VoiceNotes implementation illustrates the issue:

```js
import { $getNodeByKey, $getRoot } from "lexical"
import { mergeRegister } from "@lexical/utils"
import { defineExtension } from "@lexical/extension"
// AudioRecorderNode imports DecoratorNode from "lexical"
import { $createAudioRecorderNode, AudioRecorderNode } from "./audio_recorder_node"
```

Or our slash-command extension:

```js
import { RootNode, $isParagraphNode } from "lexical"
import { $createCodeNode } from "@lexical/code"
```

If Lexical was OO, it could simply be a case of calling functions on objects we could expose within the extension. With Lexical's implementation, those messages are the functions and types exposed from `lexical`, and potentially `@lexical/code` etc. A straightforward example would be handling paste: an extension will need to import `PASTE_COMMAND` from `lexical`.

Since we build Lexxy from `npm` through `yarn`, we can easily `import { $getNodeByKey, $getRoot } from "lexical"`. The situation gets more complicated with importmaps since—as @excid3 points out—Lexical will need to be updated in lock-step with Lexxy. The same situation would occur if using a non-flat `node_modules`, as `pnpm` does.

jorgemanrubia (commented):
Good points @samuelpecher @excid3. I am fine with going forward with this 👍.

> I'm not a huge fan of return Lexxy.Lexical.defineExtension(..., perhaps Lexxy.Extension should provide a defineLexicalExtension wrapper for this common use case.

Yes, I think so. In this case I think it is worth adding a wrapper.

We could also add a little section documenting how to access `Lexical` elements as part of the extensions article, as part of this PR.

fillipeppalhares:
Hello guys! I would like to add my 2 cents about what happened today. I ran into both sides of this problem while writing custom extensions for Lexxy.

Lexical runtime conflict: My extensions imported defineExtension, DecoratorNode, etc. directly from "lexical". Because yarn hoisted my pinned lexical@0.45.0 separately from Lexxy's own copy, I ended up with two Lexical runtimes. Commands dispatched from one instance were invisible to nodes registered via the other — icons inserted fine but importJSON/exportJSON round-trips silently broke. My workaround: a thin ./lexical.js re-export file that all my extensions import instead of "lexical" directly, ensuring esbuild deduplicates to the same module.

Same class of problem with DOMPurify: Lexxy bundles its own DOMPurify and uses it internally. My text-align sanitization hook (DOMPurify.addHook(...)) landed on a different DOMPurify instance than the one Lexxy's styleFilterHook runs on, so the hook was never called during sanitization. Fix: pin "dompurify": "3.4.5" as a direct dep so yarn hoists exactly Lexxy's version for everyone.

These were my imports:
```js
// bootstrap_icon_extension.js
import { defineExtension, DecoratorNode, $getSelection, $isRangeSelection, COMMAND_PRIORITY_LOW } from "lexical"
import * as Lexxy from "@37signals/lexxy"

// subscript_superscript_extension.js
import { defineExtension, FORMAT_TEXT_COMMAND, COMMAND_PRIORITY_LOW } from "lexical"
import * as Lexxy from "@37signals/lexxy"

// text_alignment_extension.js
import { defineExtension, FORMAT_ELEMENT_COMMAND, COMMAND_PRIORITY_LOW } from "lexical"
import * as Lexxy from "@37signals/lexxy"
import DOMPurify from "dompurify"

// config.js
import * as Lexxy from "@37signals/lexxy"
import { TextAlignmentExtension } from "./text_alignment_extension"
import { SubscriptSuperscriptExtension } from "./subscript_superscript_extension"
import { BootstrapIconExtension } from "./bootstrap_icon_extension"
```

anquinn:
Hi folks, any update on where this is headed? I hit this issue building a Lexxy extension and would love to see it land.

The importmap path just doesn't work for extensions right now. Mine imports `@37signals/lexxy` plus a few primitives like `$getSelection` and `$createTextNode` from `lexical`. With importmap there's no way to point `lexical` at the same instance Lexxy bundles, so you end up with a second runtime. My `$getSelection()` reads the second lexical's editor state, which Lexxy never touches, so it comes back null and the insert quietly fails. Seems like the same dual-runtime thing @fillipeppalhares ran into but a different symptom.

esbuild works, but I had to force dedup along myself. `yarn add @37signals/lexxy@0.9.22 lexical` gave me `lexical@0.46.0` up top with Lexxy's own `@lexical/*` on `0.44.0` underneath, so two copies again. Pinning `lexical@0.44.0` to match what Lexxy bundles brings it back to one and things work from there. The catch is you kind of have to know Lexxy's internal lexical version to do that, and re-check it whenever Lexxy updates.

Exporting Lexical through Lexxy would fix both cases. Extensions could just grab what they need from `@37signals/lexxy` at the right version, with no `lexical` pin at all. Happy to test a branch if that helps.

jorgemanrubia:
@excid3 if you can address this, I will be happy to merge:

> I'm not a huge fan of return Lexxy.Lexical.defineExtension(..., perhaps Lexxy.Extension should provide a defineLexicalExtension wrapper for this common use case.

We should also update the docs to reflect this.

excid3:
Sounds good @jorgemanrubia 👍 

Will try and get to it today or tomorrow.

excid3:
Alright, added `defineExtension` to `Lexxy.Extension` so you can use `return this.defineExtension(...)` now.

jorgemanrubia (approved):
Thanks @excid3 I pushed a commit documenting this, will merge in a bit 🙏

PR #1050, opened 2026-05-13
Stop propagation of the triple-click to prevent Lexical's handler from running

flavorjones:
Lexical's onClick handler implements a triple-click handler that is trivial/anemic/naïve. The intention of the change, made in facebook/lexical#4512, seems to be to deal with browsers' "overselection" behavior, where a triple-click selection might end at offset 0 of the following block, which can cause issues when transforming the selection. But the implementation breaks many common real-world use cases.

An upstream bug report was filed in June 2025 for this behavior: https://github.com/facebook/lexical/issues/7592

Just to name two common impacted scenarios that are top-of-mind:

- triple-clicking in a paragraph with soft line breaks (`<br>` tags)
- triple-click-dragging to select multiple lines

The specific motivating reason for Lexical's handler seems to be that the "Header" tool in Lexical doesn't handle this overselection and will header-ify the following line; and that table cells can be deleted entirely once selected.

However, Lexxy has its own "Header" tool that doesn't have this issue; and Lexical tables no longer exhibit the cell-deletion behavior. I think it might be better to improve the formatting tool(s) than to hack the selection mechanism.

Hopefully, we can remove this once Lexical addresses this behavior. Some possible upstream approaches for that might be:

- Move the existing selection handling logic into a real handler that applications can pre-empt.
- Fix the selection behavior, or drop the triple-click handler completely.

ref: https://app.basecamp.com/2914079/buckets/41746046/card_tables/cards/9861391060

### DIFF
--- a/src/extensions/prevent_lexical_triple_click_extension.js
+++ b/src/extensions/prevent_lexical_triple_click_extension.js
@@ -1,0 +1,35 @@
+import { defineExtension } from "lexical"
+import { registerEventListener } from "../helpers/listener_helper"
+import LexxyExtension from "./lexxy_extension"
+
+export class PreventLexicalTripleClickExtension extends LexxyExtension {
+  get lexicalExtension() {
+    return defineExtension({
+      name: "lexxy/prevent-lexical-triple-click",
+      register: (editor) => editor.registerRootListener((rootElement) => {
+        if (rootElement) {
+          return registerEventListener(
+            rootElement,
+            "click",
+            this.#handleTripleClick.bind(this),
+            { capture: true }
+          )
+        }
+      })
+    })
+  }
+
+  // Stop propagation of the triple-click to prevent Lexical's handler from running.
+  //
+  // Lexical's onClick handler implements a triple-click handler that is trivial/anemic/naïve. The
+  // intention of the change, made in facebook/lexical#4512, seems to be to deal with browsers'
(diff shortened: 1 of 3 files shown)


src/extensions/prevent_lexical_triple_click_extension.js
```diff
@@ -0,0 +1,30 @@
...
+export class PreventLexicalTripleClickExtension extends LexxyExtension {
+  get lexicalExtension() {
+    return defineExtension({
+      name: "lexxy/prevent-lexical-triple-click",
+      register: () => registerEventListener(
+        this.editorElement.editorContentElement,
+        "click",
+        this.#handleTripleClick.bind(this),
+        { capture: true }
+      )
```
samuelpecher: I think we should move towards all operations on the root element being in `registerRootListener` 

```suggestion
      register: () => registerRootListener(rootElement => registerEventListener(
        rootElement,
        "click",
        this.#handleTripleClick.bind(this),
        { capture: true }
      ))
```


This means the timing and life-cycle is all funneled through `Editor.registerRootElement()`

samuelpecher (approved):
YES PLEASE

src/extensions/prevent_lexical_triple_click_extension.js
flavorjones: Right on, fixed in latest branch.

PR #1053, opened 2026-05-15
Improve attachment accessibility

brunoprietog:
Make attachments and galleries usable by keyboard and screen-reader users without changing serialized HTML. Includes #1270.

- Move attachment controls into a contextual toolbar outside the editor, reachable with Alt+F10 and dismissed with Escape. Tab opens the caption field; Shift+Tab returns to the selected attachment. Delayed selection changes preserve focus on the toolbar and caption field.
- Give selected attachments a readable label combining their caption and distinct alternative text. Gallery navigation uses the image and caption directly. Custom inline attachments prepare their avatar and text labels near the caret and restore them when it moves away, preserving labels hidden by the host application.
- Edit captions in a shared, labelled textarea outside the editor textbox. Escape reselects the attachment; Enter continues after it.
- Move attachments and create or extract galleries with Alt+Shift+Arrow. Announce the result through ariaNotify when available, with a live-region fallback that expires each announcement independently.

Image descriptions update when alternative text changes or is cleared. The authoring UI and Action Text persistence are added separately in #1275.

Basecamp card: https://app.basecamp.com/2914079/buckets/47229151/card_tables/cards/9894911453

Validation:
- On the stacked branches, 177 affected browser checks passed across Chromium, Firefox, and WebKit without retries.
- Caption and alternative-text round trips passed on Rails main (4 passed, 1 compatibility skip); the unsupported-Action-Text check passed on Rails 8.1.
- ESLint, RuboCop, and build passed.
- The existing intermittent WebKit scroll failure was also reproduced before these changes (3 failures in 30 runs).


src/elements/attachment_toolbar.js
samuelpecher, line 1: Is the <kbd>Alt+F10</kbd> shortcut tied to a floating delete button? Could the same shortcut be mapped to jumping on/off an individual `DecoratorNode`'s deletion button?

I ask as floating buttons are more difficult to style onto the node and can end up mis-aligned with the node. The best positional `code` is the one we don't have to write. I think it's better to have more complexity jumping onto/off each node's button rather than moving a singleton button around.

src/editor/attachments/keyboard_move.js
```diff
@@ -0,0 +1,159 @@
...
+    switch (key) {
+      case "ArrowUp": return new AttachmentReorder(node, Direction.backward).moveVertically()
+      case "ArrowDown": return new AttachmentReorder(node, Direction.forward).moveVertically()
+      case "ArrowLeft": return new AttachmentReorder(node, Direction.backward).reorderInGallery()
+      case "ArrowRight": return new AttachmentReorder(node, Direction.forward).reorderInGallery()
+      default: return null
+    }
+  }
+
+  #blockedMessageFor(key) {
```
samuelpecher: I'd go with `blockedMessageFor(direction)` if you can swing it.

src/editor/attachments/keyboard_move.js
```diff
@@ -0,0 +1,159 @@
...
+        return "Image reordered in gallery"
+      }
+    }
+    return null
+  }
+
+  #extractFromGallery() {
+    const gallery = this.node.getParent()
+    this.node.remove()
+    this.direction.insertNextTo(gallery, this.node)
```
samuelpecher: Small nit, a move implies an initial removal. Lexical will take care of the removal in the move.

src/helpers/direction.js
samuelpecher, line 1: I'd like to align with Lexical's terminology of "next" and "previous"

src/editor/attachments/fake_selection.js
samuelpecher, line 1: This is the change which gives me the most pause and seems like it will need the most continuing maintenance to keep working.

What's the viability of manually announcing what's in the NodeSelection accessibility-wise?

samuelpecher (commented):
@brunoprietog I've made a few comments to start threads on the various workarounds proposed. I do want to note the high quality of the execution of the workarounds; my questions are on overall approach.

src/editor/attachments/fake_selection.js
brunoprietog: This was the trickiest part for sure. The reason for this fake selection is that we need to make the screen reader believe that something has changed in the cursor itself. If we use a live region or anything else, even if we can announce the correct information, the cursor will also be read, with incorrect information about the cursor, such as the first letter of the previous paragraph. In other words, the live region is just an additional announcement, but it doesn’t fix the incorrect announcements about the cursor.

src/editor/attachments/keyboard_move.js
brunoprietog: Moved it to `AttachmentReorder`, so no arguments are required

src/elements/attachment_toolbar.js
brunoprietog: Alt+F10 is used to focus the toolbar, which could have any button actually. The problem is that we can't have those buttons inside the `contenteditable` element, which is why I left them as siblings. I improved the code there by using a `ResizeObserver`.

src/editor/attachments/keyboard_move.js
```diff
@@ -0,0 +1,143 @@
...
+      && this.node.isPreviewableImage
+  }
+
+  #formGalleryWith(target) {
+    if ($findOrCreateGalleryForImage(target)) {
+      if (this.direction.isNext) {
+        target.insertBefore(this.node)
+      } else {
+        target.insertAfter(this.node)
+      }
```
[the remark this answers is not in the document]
brunoprietog: That's not the desired behavior

src/helpers/direction.js
samuelpecher, line 1: This is a nice API, though very close and overlapping with `NodeCaret`. I think we should standardize around one and there's more "safety in numbers" using `NodeCaret`.

These seem equivalent, though your `Direction` has a readability edge.
```js
Direction.next.siblingOf(node) === $getSiblingCaret(node, "next").getNodeAtCaret()
Direction.previous.siblingOf(node) === $getSiblingCaret(node, "previous").getNodeAtCaret()

Direction.next.edgeChildOf(block) === $getChildCaret(block, "next").getNodeAtCaret()
Direction.previous.edgeChildOf(block) === $getChildCaret(block, "previous").getNodeAtCaret()

Direction.next.insertBeside(reference, node) === getSiblingCaret(reference, "next").insert(node)
Direction.previous.insertBeside(reference, node) === getSiblingCaret(reference, "previous").insert(node)
```

Carets also enable slimmer, symmetrical code as they "plug-in" to many Lexical APIs and can be passed around instead of node references (see https://github.com/basecamp/lexxy/pull/1049 and https://github.com/basecamp/lexxy/pull/1063).

brunoprietog: Makes sense, I've removed the helper.

PR #1068, opened 2026-05-25
Fix Error 230 if table is only/last element in doc

zoltanhosszu:
Only show TableTools when a root element is set. This prevents a bug which crashes Lexxy on load when a Table is at the end of the content. It is selected and TableTools cannot resolve the DOM node.

https://app.basecamp.com/2914079/buckets/41746046/card_tables/cards/9926521235

edited by @samuelpecher


jorgemanrubia (approved):
👏

PR #1072, opened 2026-05-27
Fix broken image flash on bridge-managed uploads

mbarta:
Bridge-managed uploads (`uploadUrl` is null) hand Lexxy a placeholder `File` with no real image bytes. `$showUploadedAttachment` still built a `previewSrc` from it via `URL.createObjectURL`, producing a `blob:` URL pointing at empty content that rendered as a broken image until `#preloadAndSwapSrc` swapped in the server URL.

Gate the local preview on `uploadUrl != null` — the same signal `createDOM` already uses — so bridge-materialized images point straight at the server URL instead of a broken blob.

### DIFF
--- a/src/editor/contents.js
+++ b/src/editor/contents.js
@@ -14,6 +14,7 @@ import { $forEachSelectedTextNode, $setBlocksType } from "@lexical/selection"
 import Uploader from "./contents/uploader"
 import { $isActionTextAttachmentNode } from "../nodes/action_text_attachment_node"
 import { $createActionTextAttachmentUploadNode, ActionTextAttachmentUploadNode } from "../nodes/action_text_attachment_upload_node"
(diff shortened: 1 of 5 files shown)


samuelpecher:
I think we'd be better off with `ManagedAttachmentUploadNode` or allowing an upload delegate for the upload node if there's this much divergence

mbarta:
> I think we'd be better off with `ManagedAttachmentUploadNode` or allowing an upload delegate for the upload node if there's this much divergence

Thanks @samuelpecher, are you happy to take over? I can give it a go using Claude but it'd still require a review from a web developer.

samuelpecher:
Yep: https://app.basecamp.com/2914079/buckets/41746046/card_tables/cards/9937070039

PR #1073, opened 2026-05-28
Add only-at attribute to <lexxy-prompt>

jorgemanrubia:
## Summary

Two small additions to `<lexxy-prompt>`:

### `only-at` attribute

Controls *where* the trigger character is allowed to open the prompt. The attribute takes a regular expression that is matched against the text immediately preceding the trigger; the prompt opens only when the regex matches.

The default behavior is preserved — when `only-at` isn't set, the prompt still opens at the start of a line or right after a space/newline (`^|[ \n]`).

This unlocks cases like:

- A "bot" prompt that should only fire when the trigger is the first character on the line:
  ```html
  <lexxy-prompt trigger="!" name="bot" only-at="^">…</lexxy-prompt>
  ```
- A prompt that should fire anywhere, including mid-word:
  ```html
  <lexxy-prompt trigger="#" name="tag" only-at=".*">…</lexxy-prompt>
  ```

The pattern is anchored at the end automatically, so users only describe what should precede the trigger.

### `insert-editable-text` bypasses the content-type gate

`<lexxy-prompt insert-editable-text>` never creates an Action Text attachment — the chosen template is inserted verbatim as editor text. The existing `#promptContentTypePermitted` gate still forced these prompts to satisfy the editor's `supportsAttachments` flag and `permitted-attachment-types` allowlist, so a text-only prompt would silently fail to register its trigger listener whenever the host editor restricted attachments.

The gate now short-circuits to `true` when `insert-editable-text` is present, since no attachment is ever created.

## Context

Discussed on [this Basecamp card](https://app.basecamp.com/2914079/buckets/41746046/card_tables/cards/9937270443) — needed for the BC4 bots prompt, where pressing `!` should open the bot selector at the start of a chat line and insert plain text without any attachment plumbing.

## Changes

- `src/elements/prompt.js` — read the `only-at` attribute and use it to gate the trigger detector; short-circuit `#promptContentTypePermitted` for `insert-editable-text` prompts.
- `test/browser/fixtures/prompt-only-at.html` + `test/browser/tests/prompts/only_at.test.js` — Playwright coverage for `only-at` (start-of-line, anywhere, and the default).
- `test/browser/fixtures/prompt-insert-editable-text-bypasses-content-type.html` + two new cases in `test/browser/tests/prompts/prompt_suppression.test.js` — Playwright coverage for the `insert-editable-text` short-circuit, with both a restricted allowlist and `attachments="false"`.
- `docs/prompts/options-reference.md` — document `only-at` with examples.

### DIFF
--- a/src/elements/prompt.js
+++ b/src/elements/prompt.js
@@ -127,6 +127,10 @@ export class LexicalPromptElement extends HTMLElement {
   }
 
   get #promptContentTypePermitted() {
+    // `insert-editable-text` prompts never create attachments, so the
+    // editor's attachment support and content-type allowlist don't apply.
+    if (this.hasAttribute("insert-editable-text")) return true
+
     const el = this.#editorElement
     if (!el.supportsAttachments) {
       return false
--- a/src/elements/prompt.js
+++ b/src/elements/prompt.js
@@ -2,6 +2,7 @@ import Lexxy from "../config/lexxy"
 import { createElement, generateDomId, parseHtml } from "../helpers/html_helper"
 import { getNonce } from "../helpers/csp_helper"
 import { $createTextNode, $getSelection, $isRangeSelection, $isTextNode, COMMAND_PRIORITY_CRITICAL, INPUT_COMMAND, KEY_ARROW_DOWN_COMMAND, KEY_ARROW_UP_COMMAND, KEY_ENTER_COMMAND, KEY_SPACE_COMMAND, KEY_TAB_COMMAND } from "lexical"
+import { $textBeforeOffset } from "../helpers/lexical_helper"
 import { CustomActionTextAttachmentNode } from "../nodes/custom_action_text_attachment_node"
 import InlinePromptSource from "../editor/prompt/inline_source"
 import DeferredPromptSource from "../editor/prompt/deferred_source"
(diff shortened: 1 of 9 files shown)


src/helpers/lexical_helper.js
```diff
@@ -148,6 +148,39 @@ export function $isListItemStructurallyEmpty(listItem) {
   return true
 }
 
+// Returns the document text up to `offset` inside `targetNode`. Non-inline
+// element siblings are joined with `\n\n`, matching Lexical's own
+// ElementNode.getTextContent behavior.
+export function $textBeforeOffset(targetNode, offset) {
```
samuelpecher, line 154: This could be offloaded to a `RangeSelection`, and use the built-in `getTextContent()`

```js
export function $textBeforeOffset(targetNode, offset) {
  const selection = $createRangeSelection()
  selection.anchor = $createPoint("root", 0, "element")
  selection.focus = $createPoint(targetNode.getKey(), offset, $isElementNode(targetNode) ? "element" : "text")
  return selection.getTextContent()
}
```

I guess this is what `TextPointCaretSlice.getTextContent()` is for, but this gets the simplification fastest. 😆

samuelpecher (approved):
Love the advanced API. The knife is infinitely sharp.

src/elements/prompt.js
```diff
@@ -118,7 +122,15 @@ export class LexicalPromptElement extends HTMLElement {
...
   }
 
+  get #onlyAtRegExp() {
+    return new RegExp(`(?:${this.onlyAt ?? DEFAULT_ONLY_AT_PATTERN})$`)
+  }
+
   get #promptContentTypePermitted() {
+    // `insert-editable-text` prompts never create attachments, so the
+    // editor's attachment support and content-type allowlist don't apply.
+    if (this.hasAttribute("insert-editable-text")) return true
```
jorgemanrubia, line 132: @flavorjones @samuelpecher I am relaxing the condition so that we don't check the attachment mime type when `insert-editable-text` is true: in that case, we inser the attachment as plain text, so I think this does not make sense. Let me know if I am missing somehing here please.

samuelpecher: This seems like the right behavior to me

PR #1080, opened 2026-05-29
Prompt vertical direction

zoltanhosszu:
This PR adds a new `vertical-direction` attribute for prompts, allowing the prompt menu to be forced to open either upward or downward.

Supported values:
- `top`
- `bottom`

```HTML
<lexxy-prompt trigger="@" src="..." name="mention" vertical-direction="top"></lexxy-prompt>
```

By default, Lexxy chooses the prompt menu direction based on the available space in the window. In layouts with fixed or floating elements, parts of the screen may be visually covered even though they are still within the window bounds. This attribute lets applications override the automatic positioning so the prompt menu opens in the direction that best fits their UI.


samuelpecher (approved):
Nice

PR #1082, opened 2026-05-29
Restrict previewable image types and add config

samuelpecher:
Restrict previewable image types for determining what can be previewed in-browser and expect a preview from the server. Only image corresponding to the types will be galleryable.

Files can still be replaced by a preview when the server sends `previewable: true` in the Blob attributes.

I'm proposing a default of  `/^image(\/(avif|hei[cf]|gif|png|jpeg|webp)|$)/` which expands [Trix's default](https://github.com/basecamp/trix/blob/23ab2f6504d2a07d9e319e3617331d03fa6c36db/src/trix/models/attachment.js#L7) of `/^image(\/(gif|png|webp|jpe?g)|$)/` to include more modern file types from the default ActiveStorage `variable_content_types`.

I'm unsure to include TIFFs, PSDs or ICOs in the pattern as they don't feel like they belong in a gallery since they can't be viewed directly in the browser. Would we rather complete alignment with the default `variable?` file types and figure out the upload preview situation?

Full default `config.active_storage.variable_content_types` lists:
      image/png
      image/gif
      image/jpeg
      image/tiff
      image/bmp
      image/vnd.adobe.photoshop
      image/vnd.microsoft.icon
      image/webp
      image/avif
      image/heic
      image/heif


samuelpecher:
cc @lylo as I know you've got an image-focused implementation

docs/configuration.md
```diff
@@ -63,6 +63,7 @@ Global options apply to all editors in your app and are configured using `Lexxy.
 
 - `attachmentTagName`: The tag name used for [Action Text custom attachments](https://guides.rubyonrails.org/action_text_overview.html#signed-globalid). By default, they will be rendered as `action-text-attachment` tags.
 - `attachmentContentTypeNamespace`: The default content_type namespace for prompts. The default is `actiontext` which will result in `application/vnd.actiontext.[type]`.
+- `attachmentPreviewablePattern`: A RegExp identifying which content types are previewable images, and therefore valid gallery children. For other file types, the server can return `{ previewable: true }` in a blob's attributes to show a preview. The default is `/^image(\/(avif|hei[cf]|gif|png|jpeg|webp)|$)/`
```
jorgemanrubia, line 66: More than a regexp, I think a list of types would be more convenient to configure externally here: `[ "avi", "heif" ]`. Is there any scenario where you need to go lower level and tweak the image bit?

src/nodes/action_text_attachment_node.js
```diff
@@ -1,10 +1,12 @@
 import Lexxy from "../config/lexxy"
 import { $getEditor, $getNearestRootOrShadowRoot, DecoratorNode, HISTORY_MERGE_TAG } from "lexical"
-import { createAttachmentFigure, createElement, isPreviewableImage } from "../helpers/html_helper"
+import { createAttachmentFigure, createElement } from "../helpers/html_helper"
 import { bytesToHumanSize, extractFileName } from "../helpers/storage_helper"
 import { parseBoolean } from "../helpers/string_helper"
 import { REWRITE_HISTORY_COMMAND } from "../extensions/rewritable_history_extension"
 
+// Common image types from default config.variable_content_types
+const DEFAULT_PREVIEWABLE_PATTERN = /^image(\/(avif|hei[cf]|gif|png|jpeg|webp)|$)/
```
jorgemanrubia, line 9: Should the default be in the `presets`? in `src/config/lexxy.js`

jorgemanrubia (commented):
👏

src/nodes/action_text_attachment_node.js
```diff
@@ -76,6 +78,14 @@ export class ActionTextAttachmentNode extends DecoratorNode {
     }
   }
 
+  static isPreviewableImage(contentType) {
+    return contentType.match(this.PREVIEWABLE_PATTERN)
+  }
+
+  static get PREVIEWABLE_PATTERN() {
```
jorgemanrubia, line 85: Minor: a bit unusual to use a constant nomenclature for something that is not a constant. I would use a regular method name `previewablePattern`

docs/configuration.md
samuelpecher: Much like https://github.com/basecamp/lexxy/pull/1073, I wanted to give the sharpest knife. We assume previewable attachments are images, but others may want to allow other top-level mime types in galleries.

This is also the approach Trix took.

PR #1084, opened 2026-05-31
Support Elixir

patrols:
## What

Adds **Elixir** as a supported code-block language, so it appears in the language picker and gets syntax highlighting.

## Why

Elixir was missing from the picker (the dropdown jumps from Diff to Go), so code blocks couldn't be highlighted as Elixir. Prism ships an Elixir grammar, so this is a small, self-contained addition.

## How

Mirrors the existing manual language entries (Ruby/PHP/Go/Bash/JSON/Diff/Kotlin):

- `src/config/prism.js` — register the grammar: `import "prismjs/components/prism-elixir"`.
- `src/elements/code_language_picker.js` — add `languages.elixir ||= "Elixir"` to the picker (needed because `@lexical/code`'s `CODE_LANGUAGE_FRIENDLY_NAME_MAP` doesn't include Elixir — same reason the other languages above are hand-added). The getter sorts alphabetically, so it slots in automatically.
- `test/javascript/unit/helpers/code_highlighting_helper.test.js` — add `"elixir"` to `expectedGrammars` so the parametric test asserts the grammar loads.
- `test/dummy/app/views/sandbox/_code.html.erb` — add an Elixir example (alphabetically between Diff and Go).

`prismjs/components/prism-elixir` has no extra Prism dependency (no `require` in `components.json`), so nothing else needs importing.

## Testing

- `yarn test` — passes, including the new `Prism includes the elixir grammar` assertion.
- `yarn lint` — clean.
- `yarn build` — builds cleanly with the new import.

### DIFF
--- a/src/config/prism.js
+++ b/src/config/prism.js
@@ -34,5 +34,6 @@ import "prismjs/components/prism-go"
 import "prismjs/components/prism-bash"
 import "prismjs/components/prism-json"
 import "prismjs/components/prism-kotlin"
+import "prismjs/components/prism-elixir"
 
 export default Prism
--- a/src/elements/code_language_picker.js
+++ b/src/elements/code_language_picker.js
@@ -71,6 +71,7 @@ export class CodeLanguagePicker extends HTMLElement {
     languages.json ||= "JSON"
     languages.diff ||= "Diff"
     languages.kotlin ||= "Kotlin"
+    languages.elixir ||= "Elixir"
 
 
     // Place the "plain" entry first, then the rest of language sorted alphabetically
--- a/test/dummy/app/views/sandbox/_code.html.erb
+++ b/test/dummy/app/views/sandbox/_code.html.erb
@@ -142,6 +142,26 @@ h1 {
 
 <p><br/></p>
 
+<h4>Elixir</h4>
+<pre data-language="elixir" data-highlight-language="elixir">
+defmodule Greeter do
+  @greeting "Hello"
+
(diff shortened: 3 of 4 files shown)


samuelpecher:
Hey @patrols. I think the best way to go here would be introducing a straightforward way for implementers to customize the list for their app rather than trying to support a large super-set in Lexxy core.

patrols:
@samuelpecher Yeah, fair enough. no point growing the core list forever, and it all gets bundled eagerly right now anyway.

I think `configure()` is already a decent hook for this, since it runs before the editor elements mount. Something like:

```js
Lexxy.configure({ languages: { elixir: "Elixir" } })
```

and the picker merges that on top of the Lexical map + the built-in names, instead of me hardcoding each one.

Couple of things I wasn't sure about:

- That only handles the dropdown name. the highlighting still needs the grammar on Lexxy's Prism. In a bundler app the consumer can just `import "prismjs/components/prism-elixir"` and it lands on the shared instance, but since Lexxy ships prebuilt, an importmap app (like mine) can only reach that instance through the `window.Prism` global. So: leave grammar registration to the app that way, or have Lexxy expose its Prism / take the grammar so there's a blessed path?
- Want me to trim the built-in list down to a smaller default as part of this, or just add the hook and leave the current set for now?

Can rework the PR around whatever you'd prefer.

patrols:
Going to close this out. After digging into the current internals, I found I can get full Elixir support in my app entirely at runtime

PR #1088, opened 2026-06-03
Make selection w/ Arrow Keys synchronous

samuelpecher:
- remove `await nextFrame` resulting in inconsistent selection state from stale `hasNodeSelection` read
- unwrap `editor.update` from `withCurrentNode` always used in a command context
-  `withCurrentNode` => `withCurrentNodeSelectionNode`
- `return true` from ArrowKey handlers which do handle the command 
- ensure handled events get `preventDefault()` (this was being done by loose handlers down the chain)

https://app.basecamp.com/2914079/buckets/41746046/card_tables/cards/9959684586

### DIFF
--- a/src/editor/contents.js
+++ b/src/editor/contents.js
@@ -55,11 +55,6 @@ export default class Contents {
     inserter.insertNodes(nodes)
   }
 
-  insertAtCursorEnsuringLineBelow(node) {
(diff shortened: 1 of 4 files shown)


jorgemanrubia (approved):
👏

PR #1101, opened 2026-06-10
Fix Dragon NaturallySpeaking corrections breaking the editor

brunoprietog:
A customer who dictates with Dragon NaturallySpeaking reported that correcting a word by voice in any rich text field destroys everything after the corrected word ([Basecamp card](https://app.basecamp.com/2914079/buckets/47229151/card_tables/cards/9966179255)). The cause is in the handler Lexical ships for Dragon's web extension messages, which Lexxy picks up through the RichTextExtension dependency chain: Dragon sends the number -1 as the text argument to mean "change the selection without touching the text", the handler passes it straight to `insertRawText`, and the resulting exception inside `editor.update` leaves the editor permanently broken. From then on the extension edits the DOM behind Lexical's back and content disappears when the two diverge.

Blocking the extension means catching its window message before it acts, and its content script registers at `document_end`. A handler that loads with the editor can't win that race when the editor loads lazily, well after `document_end`.

So this PR:

- Handles the protocol from a listener registered as Lexxy loads, early enough to beat the extension's content script and stop it from editing the DOM directly. The listener carries no Lexical, so it stays light, and dispatches a document event to the editor-side code that applies the change once Lexxy has loaded.
- Honors the -1 sentinel, so corrections and cursor moves by voice work without breaking the editor.
- Validates every makeChanges payload before `editor.update`, malformed ones included, since both the extension's raw DOM edits and the broken stock handler behind us would corrupt the editor otherwise.
- Applies Dragon's format commands ("bold that"), mapping the execCommand names the extension sends to Lexical text formats.
- Adds Playwright coverage for the protocol: replacement, selection-only, the full correction flow, malformed and out-of-range payloads, format commands, unknown commands, and clamped selections not toggling the pending format.

Apps that load Lexxy from their eager bundle get this with no extra step. Apps that load it lazily add `import "@37signals/lexxy/dragon"` to their eager bundle so the listener still wins the race.

The protocol fix and the format mapping also went upstream in facebook/lexical#8665, which adds an `installDragonSupport()` entrypoint API for the race. Once a release we adopt ships, we can hand the editor-side processing to Lexical's handler. The lightweight listener that wins the race for lazy-loaded apps is still ours for now, and if it's worth bringing upstream too, we'll follow up with another PR.

I validated the handler against the real extension's unpacked content scripts in both registration orders, but final confirmation with Dragon itself will have to come through the customer, since it needs their setup.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2,6 +2,8 @@ import "./config/prism"
 import "./config/dom_purify"
 
 import { defineElements } from "./elements/index"
+import "./editor/vendor/dragon/support"
+import "./editor/vendor/dragon/make_changes"
 
 import Lexxy from "./config/lexxy"
 
--- a/package.json
+++ b/package.json
@@ -6,7 +6,8 @@
   "type": "module",
   "exports": {
     ".": "./dist/lexxy.esm.js",
-    "./helpers": "./dist/lexxy_helpers.esm.js"
+    "./helpers": "./dist/lexxy_helpers.esm.js",
+    "./dragon": "./dist/lexxy_dragon.esm.js"
   },
   "files": [
     "dist"
--- a/rollup.config.npm.mjs
+++ b/rollup.config.npm.mjs
@@ -6,7 +6,7 @@ const helperChunks = [
 ]
 
 export default {
-  input: { lexxy: "src/index.js" },
+  input: { lexxy: "src/index.js", lexxy_dragon: "src/editor/vendor/dragon/support.js" },
   output: {
     dir: "./dist",
     format: "esm",
--- a/scripts/benchmarks/run-browser-benchmarks.js
+++ b/scripts/benchmarks/run-browser-benchmarks.js
@@ -61,6 +61,17 @@ const SCENARIOS = [
     },
     title: "Load many attachments",
   },
+  {
+    description: "Move the caret across a large document to measure selection-only update cost",
+    kind: "caret",
+    name: "move-caret-large-document",
+    payload: {
+      kind: "paragraphs",
+      paragraphs: 200,
+      wordsPerParagraph: 40,
+    },
+    title: "Move caret on large document",
+  },
 ]
 
 main().catch((error) => {
--- a/src/editor/vendor/dragon/make_changes.js
+++ b/src/editor/vendor/dragon/make_changes.js
@@ -1,5 +1,14 @@
 import { $getSelection, $isRangeSelection, $isTextNode } from "lexical"
 
+const TEXT_FORMAT_BY_EXEC_COMMAND = new Map([
+  [ "bold", "bold" ],
+  [ "italic", "italic" ],
+  [ "strikeThrough", "strikethrough" ],
+  [ "subscript", "subscript" ],
+  [ "superscript", "superscript" ],
+  [ "underline", "underline" ]
+])
+
 document.addEventListener("lexxy:dragon-make-changes", (event) => {
   const editor = document.activeElement?.closest("lexxy-editor")?.editor
   if (editor) applyMakeChanges(editor, event.detail)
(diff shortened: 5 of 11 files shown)


src/index.js
```diff
@@ -2,6 +2,7 @@ import "./config/prism"
 import "./config/dom_purify"
 
 import { defineElements } from "./elements/index"
+import DragonSupport from "./editor/dragon_support"
```
jorgemanrubia: We could move to a `editor/vendor` folder to remove from the main folder.

test/browser/tests/dragon.test.js
```diff
@@ -0,0 +1,108 @@
+import { test } from "../test_helper.js"
```
jorgemanrubia: similarly, would move to `test/browser/tests/vendor/dragon.test.js`

jorgemanrubia:
🤖 Suggestion: validate the `makeChanges` args before entering `editor.update`.

The PR's goal is to absorb malformed payloads because an exception inside `editor.update` leaves the editor permanently broken — but `#applyMakeChanges` runs the update with no validation of the args' *types* and no try/catch, so a payload that slips past the current guards and throws mid-update reproduces the exact failure mode being fixed. Concretely:

- `#setAddressedRange` checks `this.elementStart >= 0`, but `"4" >= 0` is `true` for a string, and then `this.elementStart + this.elementLength` becomes string concatenation (`"4" + "9"` → `"49"`), which flows into `setTextNodeRange` as an offset.
- Meanwhile `#setFinalSelection` guards with `Number.isFinite(this.selStart + this.selLength)`, which correctly rejects strings — so the two range-setters apply different strictness to the same protocol.
- `#finalStart` isn't clamped to 0: with a negative `selStart`, `Math.min(-5, size)` stays `-5`. The out-of-range test passes, so Lexical tolerates it today, but that relies on Lexical internals rather than this handler's own contract.

Suggestion: validate the positional args up front — the four offsets must be numbers, and `text` must be a string or the `-1` sentinel — and bail before calling `editor.update` otherwise, while still absorbing the event with `stopImmediatePropagation`. That makes the "malformed payloads included" guarantee hold by construction rather than by Lexical's tolerance. A try/catch around the update body is weaker insurance (Lexical may already be in a bad state by the time it throws) but is cheap extra belt-and-braces on top.

Related cheap hardening while you're in there: guard `#parsedMessageFrom` with `typeof event.data === "string"` (the stock handler does this) — right now non-string `event.data` gets coerced by `JSON.parse` and takes the exception path. Adding `event.data.includes("nuanria_messaging")` before parsing would also skip parsing the host app's own postMessage traffic while an editor is focused.

jorgemanrubia (commented):
Great one @brunoprietog

jorgemanrubia (approved):
@brunoprietog minor comments and a suggestion via Fable that looks quite legit 👏

samuelpecher (commented):
I presume that the existing [@lexical/dragon](https://lexical.dev/docs/api/modules/lexical_dragon) extension does not properly support the sentinel value?

Given that support was extension-shaped and [intercepted the event](https://github.com/facebook/lexical/blob/935914061d3de5d54cd3fbd1498e3f5869abf1a0/packages/lexical-dragon/src/index.ts#L115), is there an opportunity to use a Lexxy extension while still preventing the double-handling with a capturing handler?

brunoprietog:
Right, the stock handler breaks on the sentinel: it passes the -1 straight to `insertRawText`, which throws inside `editor.update` and leaves the editor permanently broken (it also destructures the args without checking they're an array).

About the extension shape, that was the first thing I tried. The problem is that capture doesn't buy any priority here. The message event's target is window itself, and at the target listeners run in plain registration order, so `stopImmediatePropagation` only helps if you registered before Dragon's content script, which registers at `document_end` of the initial page load. An extension registers when an editor mounts, and that's almost always later, since most editors mount on navigation or after some interaction. I verified it against the unpacked extension code in Chromium: with per-editor registration the extension's listener always runs first and the edit lands twice. That's why the install has to happen at module load, which doesn't fit the per-editor lifecycle of an extension.

Upstream is heading the same way: in facebook/lexical#8664 the maintainer suggested an exported function apps call synchronously at their entrypoint, without constructing an editor, and that's now part of facebook/lexical#8665. Once that ships in a release we adopt, this file becomes a call to their `installDragonSupport()` at the entrypoint.

We could still wrap the dispatch in a Lexxy extension, but since the handler resolves the focused editor from the DOM it wouldn't add much beyond ceremony, and the early call in `index.js` would still be needed. Happy to go that way if you'd prefer it for consistency, though.

PR #1106, opened 2026-06-10
Arrow-key navigation crashes when the selected node is removed before the deferred update runs

jorgemanrubia:
Arrow-key navigation on a node selection crashes with `TypeError: Cannot read properties of undefined (reading 'getTopLevelElement')` (also `selectPrevious` / `selectNext`) when the selected node disappears before the deferred navigation runs. [Fixes BC3-JS-N7TF](https://basecamp.sentry.io/issues/7513049864/) — 3,101 events / 1,647 users in BC3 since May 29.

Basecamp card: https://app.basecamp.com/2914079/buckets/41746046/card_tables/cards/9982065995

## Root cause

`Selection#withCurrentNode` awaits `nextFrame()` before acting on the current node selection. If the selected node is removed or replaced during that frame — for example, a direct upload finishing replaces the upload node via `replaceAndRewriteHistory`, which uses Lexical's `replace()` and does not restore node selections — the selection keeps the stale node key. `hasNodeSelection` still reports true, but `$getSelection().getNodes()` comes back empty, so the handler calls `fn(undefined)` and throws.

## Fix

Re-resolve the selection inside the deferred `editor.update()` and bail cleanly when it is no longer a node selection or the node is gone.

The regression test reproduces the race: select an attachment, press an arrow key, and replace the attachment node before the deferred frame runs — it fails with the exact production TypeError without the fix.

### DIFF
--- a/src/editor/selection.js
+++ b/src/editor/selection.js
@@ -470,12 +470,16 @@ export default class Selection {
 
   async #withCurrentNode(fn) {
     await nextFrame()
-    if (this.hasNodeSelection) {
-      this.editor.update(() => {
-        fn($getSelection().getNodes()[0])
-        this.editor.focus()
-      })
-    }
+    if (!this.hasNodeSelection) return
+
+    this.editor.update(() => {
+      const selection = $getSelection()
+      const currentNode = $isNodeSelection(selection) ? selection.getNodes()[0] : null
+      if (!currentNode) return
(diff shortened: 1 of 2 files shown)


jorgemanrubia:
Cosing in favor of https://github.com/basecamp/lexxy/pull/1106

Issue #212, opened 2025-09-14
importmap with problems


rodrigotoledo:
Hi, I'm trying to follow the steps, but when I tried pinning packages using importmap, I have problems with the packages. Maybe it's a dependency.

And, I'm in my environment, it's a MacBook, in the first moment, I had problems with the dependency

_LoadError (Could not open library 'vips.42':_

I'm not sure if this is an error related to active storage or just a dependency issue.

This is a Loom video showing everything I'm doing.

[https://www.loom.com/share/fe80c31b148e4257b25a85dc64537260](https://www.loom.com/share/fe80c31b148e4257b25a85dc64537260)

jorgemanrubia:
For using vips you need to install libvips. You can do so via brew in MacOS. https://guides.rubyonrails.org/active_storage_overview.html#requirements

PR #1109, opened 2026-06-10
Pasting with the caret on a quote element crashes with Lexical error #211/#212

jorgemanrubia:
[Fixes BC3-JS-N7AZ](https://basecamp.sentry.io/issues/7507281623/) — 884 events / 537 users since May 26 (180 / 138 in the last 3 days). Card: https://app.basecamp.com/2914079/buckets/41746046/card_tables/cards/9982982415

### What it fixes

Pasting with the caret sitting on the blockquote element itself (an element point on the `QuoteNode`, e.g. after clicking in the gap between two paragraphs inside a quote) crashed and lost the paste. Every Sentry event goes through the same path: `Contents#insertDOM` → `insertAtCursor` → `RangeSelection.insertNodes`, throwing Lexical error #211 ("Expected node QuoteNode of type quote to have a block ElementNode ancestor") for inline content or #212 ("…to have a block ancestor") for block content.

### How

Lexical's `RangeSelection.insertNodes` requires every selection point to have a block ancestor with inline children (`INTERNAL_$isBlock`). A lexxy quote always holds paragraphs (`$ensureQuoteHasParagraphChild`), so an element point on the quote has no such ancestor — the quote fails the leaf-block check and its parent is the root.

A new `BlockContainerNodeInserter` handles range selections where a point sits on a container of block nodes: it normalizes the selection down to a leaf position (`$normalizeSelection__EXPERIMENTAL`, the same descent Lexical applies to DOM selections) and then delegates to the regular `insertNodes`.

Includes Playwright regression tests at `test/browser/tests/paste/paste_with_caret_on_quote.test.js` covering both the #211 (inline paste) and #212 (block paste) variants. They place the caret as an element point on the quote and drive a real clipboard paste. Verified red/green: with `src/` reverted to `main` both tests fail (2 failed); with the fix both pass (2 passed).

### DIFF
--- a/src/editor/contents/node_inserter.js
+++ b/src/editor/contents/node_inserter.js
@@ -1,4 +1,4 @@
-import { $createLineBreakNode, $createParagraphNode, $createTextNode, $getChildCaretAtIndex, $isElementNode, $isLineBreakNode, $isNodeSelection } from "lexical"
+import { $createLineBreakNode, $createParagraphNode, $createTextNode, $getChildCaretAtIndex, $isDecoratorNode, $isElementNode, $isLineBreakNode, $isNodeSelection, $isRangeSelection, $normalizeSelection__EXPERIMENTAL as $normalizeSelection } from "lexical"
 import { CodeNode } from "@lexical/code"
 import { $ensureForwardRangeSelection } from "@lexical/selection"
 import { $getNearestNodeOfType } from "@lexical/utils"
@@ -9,7 +9,8 @@ export default class NodeInserter {
     const INSERTERS = [
       CodeNodeInserter,
       ShadowRootNodeInserter,
-      NodeSelectionNodeInserter
+      NodeSelectionNodeInserter,
+      BlockContainerNodeInserter
     ]
     const Inserter = INSERTERS.find(inserter => inserter.handles(selection))
     return Inserter ? new Inserter(selection) : selection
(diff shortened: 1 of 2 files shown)


romansklenar:
Another lexxy consumer here (a Rails app, lexxy 0.9.12.beta) — confirming this `#212` isn't BC3-specific; we see it in production too.

Our path matches yours exactly (`Contents#insertDOM` → `insertAtCursor` → `RangeSelection.insertNodes`), but the entry point is a **markdown paste** rather than a manual element-point-on-quote: `Clipboard#pasteMarkdown` → `contents.insertDOM(doc, { tag: PASTE_TAG })` → `insertAtCursor` → `insertNodes` → `#212` (message args `…At… quote`, so a `QuoteNode` is the container point). Real user paste, Chrome 149 / macOS.

Since both entry points funnel through the same `insertNodes`, your `BlockContainerNodeInserter` selection-normalization should cover the markdown-paste case too — just flagging it as a second real-world trigger in case the regression tests are worth extending to a markdown paste onto a quote. Happy to share more of the stack if useful.

jorgemanrubia:
Thanks for the second data point, @romansklenar. You're right that both entry points funnel through the same `insertNodes`, so `BlockContainerNodeInserter`'s selection normalization already covers the markdown-paste case.

I extended the regression tests anyway to pin the markdown path explicitly (85287166): a plain-text paste with no HTML payload routes through `Clipboard#pasteMarkdown` → `insertDOM`. Verified it red without the fix (fails with exactly `Expected node QuoteNode of type quote to have a block ancestor`, the #212 you saw) and green with it.

PR #1117, opened 2026-06-16
Fix removing list formatting from a single list item

jorgemanrubia:
## Problem

When the cursor sat on a single item within a multi-item bulleted (or numbered) list and the user clicked the list toolbar button to remove the bullet from just that one line, Lexxy removed the bullets from the **entire** list instead of only the selected item.

## Root cause

Removing a list runs `Contents#applyParagraphFormat`, which calls `$expandSelectionToLineBreaksAndSplitAtEdges(selection)` before `$setBlocksType(...)`.

That helper exists to let block formatting act on a single *visual* line inside a paragraph that contains soft `<br>` line breaks. When there are no line breaks to split — the in-list case — it fell back to:

```js
selection.anchor.getNode().getTopLevelElement()
```

For a node inside a list, the top-level element is the whole `ListNode`, so the selection was widened to span every list item. `$setBlocksType` then dutifully converted **all** items to paragraphs.

## Fix

Use `$getNearestBlockElementAncestorOrThrow` for the fallback. It returns the paragraph for a paragraph and the **list item** for a list item, keeping the conversion scoped to the block the cursor is actually on. The soft-line-break splitting path is unchanged.

## Tests

Added `test/browser/tests/formatting/list_single_item_removal.test.js` covering bullet and numbered lists: toggling the button on a middle item now splits the list and converts only that item to a paragraph. The full Playwright `formatting` suite and the broader browser suite pass; `yarn lint` is clean.

---
**Basecamp card:** https://app.basecamp.com/2914079/buckets/41746046/card_tables/cards/9969505386

## How to reproduce

1. Open the editor and create a bulleted list with several items, e.g.
   - One
   - Two
   - Three
2. Put the cursor inside a single item (e.g. "Two") — do not select the whole list.
3. Click the bullet-list button in the toolbar to remove that item's bullet.

**Before (bug):** the bullets are removed from the *entire* list.
**After (fix):** only the focused item loses its bullet; the rest of the list stays bulleted.


jorgemanrubia:

Reproduced the bug like a human via a Selenium WebDriver script run against both the released CDN build (production try-it page) and this branch's locally-built fix, using identical real-browser interactions.

**Steps performed (same in both environments):**
1. Focused the editor and cleared any preloaded content.
2. Clicked the bullet-list toolbar button, then typed three items pressing Enter between them, producing `<ul><li>First item</li><li>Second item</li><li>Third item</li></ul>`.
3. Clicked into the middle item ("Second item") to place a collapsed caret there.
4. Clicked the bullet-list toolbar button again to un-bullet just that single item.

**Production** (https://basecamp.github.io/lexxy/try-it.html — latest released CDN):
- The ENTIRE list lost its bullets. Result: `<p>First item</p><p>Second item</p><p>Third item</p>` (0 list items, 3 paragraphs).
- Bug reproduces.

**Local** (this branch, freshly built `lexxy.js` served by the dummy app):
- Only the selected item lost its bullet. Result: `<ul><li value="1">First item</li></ul><p>Second item</p><ul><li value="1">Third item</li></ul>`.
- First and Third remained bulleted; Second became a paragraph. Bug is gone, and the output matches the regression test's expectation exactly.

Verdict: **Validated** — production reproduces, local fix does not.

jorgemanrubia:
## AGENTS.md Guideline Review

Reviewed the diff against AGENTS.md and STYLE.md. All areas pass; no code changes were needed.

**Bug-fix workflow** — PASS
- Correctly classified as a core editing bug (selection/formatting/toolbar) and tested with Playwright.
- Regression test confirmed red without the fix and green with it (verified by temporarily reverting the one-line source change and re-running).

**Browser Test Organization** — PASS
- `test/browser/tests/formatting/list_single_item_removal.test.js` is in the correct folder. The `formatting/` bucket explicitly covers "block formats, lists".

**STYLE.md (JS/Lexical)** — PASS
- The fix swaps a hand-rolled `getTopLevelElement()` for Lexical's built-in `$getNearestBlockElementAncestorOrThrow`, matching "Prefer Lexical's built-in utilities" and the "don't compare `getTopLevelElement()` by hand" guidance.
- Double quotes, no semicolons, no stray `console.log`/debug, no added comments. `yarn lint` clean.

**Tests** — PASS
- Drives only the public API (`setValue`, `select`, `clickToolbarButton`) and asserts on the editor's HTML `value` — behavior, not implementation. Bullet and numbered cases grouped sensibly. Synthetic content only.

**Commit hygiene** — PASS
- Diff is exactly the two intended files; no scratch/debug files, no committed built assets, no PII.

**Action Text round-trip** — N/A
- Pure client-side editor behavior; no serialization/persistence format change, so no Capybara round-trip test required.

PR #1119, opened 2026-06-16
Strip invalid child nodes from pasted lists

jorgemanrubia:
## Problem

Pasting a numbered list copied from HelpScout adds a spurious extra leading "1.", shifting every subsequent number.

## Root cause

The pasted HTML contains invalid nodes — a stray `<br>` and whitespace text as **direct children of the `<ol>`**, before the first `<li>`:

```html
<ol>
    <br>
    <li>First item</li>
    <li>Second item</li>
    <li>Third item</li>
</ol>
```

Only `<li>` is a valid child of `<ol>`/`<ul>`. `DOMParser` keeps the stray `<br>` inside the list rather than foster-parenting it out, and Lexical's `$generateNodesFromDOM` then wraps it into an empty leading `<li value="1"><br></li>` — hence the extra number.

## Fix

A new `PastedContentFormatter` (`src/editor/contents/pasted_content_formatter.js`) owns the paste-DOM cleanup; `Contents#formatPastedDOM(doc)` delegates to it (`new PastedContentFormatter(doc).format()`). Its `#stripStrayListChildren` step removes any non-`<li>` direct child (bare `<br>`, whitespace/text) from every pasted `<ol>`/`<ul>` before import — safe by construction, since real content only ever lives inside `<li>`. The existing placeholder-anchor unwrapping and table-cell color stripping moved into the same formatter.

## Test

`test/browser/tests/paste/clean_invalid_pasted_list_html.test.js` pastes HTML with stray `<br>`/whitespace directly inside an `<ol>`/`<ul>` (synthetic content) and asserts no spurious leading item — numbering starts at the first real item. Confirmed failing on `main` (produced `<li value="1"><br></li>`) and passing with the fix.

## How to reproduce

1. Copy a numbered list from HelpScout (or any source whose HTML has stray `<br>`/whitespace directly inside the `<ol>` before the first `<li>`).
2. Paste it into the Lexxy editor.

**Before:** an extra leading "1."; numbers shifted.
**After:** numbering starts correctly at the first real item.

---
**Basecamp card:** https://app.basecamp.com/2914079/buckets/41746046/card_tables/cards/9939030998

### DIFF
--- a/src/editor/contents.js
+++ b/src/editor/contents.js
@@ -17,6 +17,7 @@ import { $isActionTextAttachmentNode } from "../nodes/action_text_attachment_nod
 import { $createActionTextAttachmentUploadNode, ActionTextAttachmentUploadNode } from "../nodes/action_text_attachment_upload_node"
 import { $getNearestBlockElementAncestorOrThrow } from "@lexical/utils"
 import NodeInserter from "./contents/node_inserter"
+import PastedContentFormatter from "./contents/pasted_content_formatter"
 import { $consecutiveSiblingGroups, $expandSelectionToLineBreaksAndSplitAtEdges, $isShadowRoot, $splitSelectedParagraphsAtInnerLineBreaks } from "../helpers/lexical_helper"
 
 export default class Contents {
@@ -364,8 +365,7 @@ export default class Contents {
   }
 
   #formatPastedDOM(doc) {
-    this.#unwrapPlaceholderAnchors(doc)
-    this.#stripTableCellColorStyles(doc)
+    new PastedContentFormatter(doc).format()
   }
 
   #dispatchPastedNodesCommand(nodes) {
@@ -530,30 +530,6 @@ export default class Contents {
     }
 
     node.remove()
-  }
-
-  // Anchors with non-meaningful hrefs (e.g. "#", "") appear in content copied
-  // from rendered views where mentions and interactive elements are wrapped in
-  // <a href="#"> tags. Unwrap them so their text content pastes as plain text
-  // and real links are preserved.
-  #unwrapPlaceholderAnchors(doc) {
-    for (const anchor of doc.querySelectorAll("a")) {
-      const href = anchor.getAttribute("href") || ""
-      if (href === "" || href === "#") {
-        anchor.replaceWith(...anchor.childNodes)
-      }
-    }
-  }
-
-  // Table cells copied from a page inherit the source theme's inline color
-  // styles (e.g. dark-mode backgrounds). Strip them so pasted tables adopt
-  // the current theme instead of carrying stale colors.
-  #stripTableCellColorStyles(doc) {
-    for (const cell of doc.querySelectorAll("td, th")) {
-      cell.style.removeProperty("background-color")
-      cell.style.removeProperty("background")
-      cell.style.removeProperty("color")
-    }
   }
 
   #getTextAnchorData() {
--- a/src/editor/contents/pasted_content_formatter.js
+++ b/src/editor/contents/pasted_content_formatter.js
@@ -1,0 +1,47 @@
+export default class PastedContentFormatter {
+  constructor(doc) {
+    this.doc = doc
+  }
+
+  format() {
+    this.#unwrapPlaceholderAnchors()
+    this.#stripTableCellColorStyles()
+    this.#stripStrayListChildren()
+    return this.doc
+  }
+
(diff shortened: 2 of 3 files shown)


jorgemanrubia:

🤖 Reproduced the bug on the released CDN build and confirmed it's gone on this branch, using a faithful client-side paste comparison (one Selenium script, two environments).

**What was tested:** Starting from an empty editor, I dispatched a real `paste` ClipboardEvent carrying `text/html` for a numbered list whose markup has an empty leading `<li>` before the first real item — the structural shape of the report (synthetic content only, no customer data):

```
<ol><li></li><li>New For You</li><li>Bubbled Up</li><li>Latest Activity</li></ol>
```

Real OS-clipboard paste can't be driven via Selenium, so the script dispatches the same `ClipboardEvent("paste")` the editor handles for a real paste, then reads `document.querySelector("lexxy-editor").value`. This exercises the exact import path under test (`Contents#formatPastedDOM`), so the CDN-vs-local comparison is apples-to-apples.

**Production** (https://basecamp.github.io/lexxy/try-it.html — latest released CDN, no fix):
- Editor value: `<ol><li value="1"></li><li value="2">New For You</li><li value="3">Bubbled Up</li><li value="4">Latest Activity</li></ol>`
- Renders the spurious leading "1." (blank) and shifts every item to 2/3/4. **Bug reproduces.**

**Local** (this branch, built `lexxy.js` served by the dummy app):
- Editor value: `<ol><li value="1">New For You</li><li value="2">Bubbled Up</li><li value="3">Latest Activity</li></ol>`
- List starts at the first real item; numbering is 1/2/3. **Bug gone.**

Evidence: screenshots captured for both environments (production renders the extra leading number; local renders correct 1/2/3 and the HTML output panel shows the clean `<ol>`).

jorgemanrubia:
## AGENTS.md Guideline Review

🤖 Reviewed the diff against AGENTS.md and STYLE.md. All areas pass; no fixes were needed.

**Bug-fix workflow** — PASS. Core editing/paste bug, classified correctly. Failing-first test verified against `origin/main` (per PR body), then fixed. Lint + relevant test folders run.

**Test placement** — PASS. `test/browser/tests/paste/paste_numbered_list.test.js` is in the right folder per the Browser Test Organization table (paste: markdown conversion, link paste, style canonicalization). Includes a control case (non-leading empty `<li>` is kept) and both `<ol>` and `<ul>` coverage.

**STYLE.md (JS/Lexical)** — PASS. The fix normalizes the parsed clipboard `Document` before Lexical node generation, so it's plain DOM work — no selection guards or `editor.read/update` concerns apply. Style is consistent with the neighbouring paste-DOM helpers (`#unwrapPlaceholderAnchors`, `#stripTableCellColorStyles`): same `for...of doc.querySelectorAll(...)` shape, private `#` methods, double quotes, no semicolons. The single guard clause in `#isEmptyListItem` is at the top of the method (allowed exception). The comment explains the non-obvious *why* (external sources emit empty leading `<li>`; Lexical preserves it; non-leading items are intentionally kept) rather than restating *what* — matching the density of surrounding editor code.

**Action Text round-trip** — N/A. The change only strips empty leading `<li>` from pasted DOM; it doesn't alter HTML export/serialization or sanitization rules, so the Action Text persistence system-test note doesn't apply. Pasted output is standard `<ol>/<li>` markup.

**Commit hygiene** — PASS. Only the two intended files (`src/editor/contents.js`, the new test). No scratch/debug files, no `console.log`/`debugger`, no committed built assets (`app/assets/javascript/lexxy.js` is gitignored). No PII — test and validation content is fully synthetic ("New For You" / "Bubbled Up" / "Latest Activity").

**Lint** — PASS. `yarn lint` clean.

**Tests** — PASS. Full `test/browser/tests/paste/` folder green on Chromium (103 passed), including the 5 new cases.

jorgemanrubia:
🤖 Validated locally — pasting a HelpScout numbered list now numbers correctly from the first item, with no spurious leading "1.".

The real root cause was a stray `<br>` + whitespace as a **direct child of the `<ol>`** (invalid — only `<li>` is a permitted child of a list). `DOMParser` keeps it inside the list rather than foster-parenting it out, and Lexical's import wraps it into an empty leading `<li value="1"><br></li>`. Fix: `#stripStrayListChildren` removes non-`<li>` direct children of pasted `<ol>`/`<ul>` before import — spec-conformant, so it can't drop real content (which only lives inside `<li>`).

PR #1120, opened 2026-06-16
Preserve trailing soft line breaks across edit/save round trips

jorgemanrubia:
## Problem

Editing a doc with soft line breaks (Shift+Enter) to add new ones breaks the earlier ones. Press Shift+Enter twice to add a soft line break and save — it persists. Edit again, add another Shift+Enter break elsewhere and save — the *earlier* break disappears.

The break survives the first save → render → re-edit cycle but is dropped on a subsequent edit + save.

## Root cause

Lexical's `LineBreakNode.importDOM` (`isLastChildInBlockNode`) deliberately drops a `<br>` that is the last child of a block element, treating it as the phantom break browsers insert to keep an empty trailing line visible in a `contenteditable`. That heuristic also discards a *genuine* soft break the user typed at the end of a line (e.g. `<p>foo<br></p>`).

Because of this, every time the saved HTML is re-imported on re-edit, a trailing `<br>` is silently dropped — so trailing breaks erode one per edit → save → re-edit cycle.

This is a Lexxy-side serialization round-trip bug at the import boundary, not a host-app (bc3) issue.

## Fix

Before importing, `protectTrailingLineBreaks(doc)` appends an empty `<span>` sentinel after a trailing `<br>` in blocks that have real content. The `<br>` is no longer the last child, so Lexical keeps it; the empty span imports to nothing. Empty blocks (`<p><br></p>`) are left untouched so they stay empty paragraphs rather than gaining a stray break.

## Test

`test/system/line_break_round_trip_test.rb` reproduces the reported scenario with real keystrokes: type text, Shift+Enter twice, save, re-edit (assert break present), add a second break elsewhere, save, re-edit, and assert the earlier break survives. Confirmed failing on clean `main` (`<p>First line<br></p>` — one break lost on first re-import) and passing with the fix.

---
**Basecamp card:** https://app.basecamp.com/2914079/buckets/41746046/card_tables/cards/9955295463

## How to reproduce

1. In a Posts editor, type a line, then press **Shift+Enter twice** (a trailing soft line break) and **Save**.
2. **Edit** again — the break is still there.
3. Add a second break elsewhere (Shift+Enter twice), **Save**, then **Edit** again.

**Before:** the *earlier* trailing break is gone — trailing soft breaks erode one per edit→save→re-edit cycle (Lexical's import drops a `<br>` that's the last child of a block).
**After:** earlier and later breaks both survive the round-trip.

> Note: overlaps Sam's bc3#11999 ("Trim trailing blank lines") — the server-side pipeline trims trailing `<br>`/`<p><br></p>`. Confirm client vs. pipeline ownership with Sam before merging.

### DIFF
--- a/src/elements/editor.js
+++ b/src/elements/editor.js
@@ -18,7 +18,7 @@ import { HorizontalDividerNode } from "../nodes/horizontal_divider_node"
 import { CommandDispatcher } from "../editor/command_dispatcher"
 import Selection from "../editor/selection"
 import { createElement, dispatch, generateDomId, parseHtml } from "../helpers/html_helper"
-import { isAttachmentSpacerTextNode, isEditorFocused } from "../helpers/lexical_helper"
+import { isAttachmentSpacerTextNode, isEditorFocused, protectTrailingLineBreaks } from "../helpers/lexical_helper"
 import { sanitize, setSanitizerConfig } from "../helpers/sanitization_helper"
 import { ListenerBin, registerEventListener } from "../helpers/listener_helper"
 import LexicalToolbar from "./toolbar"
@@ -230,6 +230,7 @@ export class LexicalEditorElement extends HTMLElement {
   }
 
   $generateNodesFromDOM(doc, { editor = this.editor } = {}) {
+    protectTrailingLineBreaks(doc)
     let nodes = $generateLexicalNodesFromDOM(editor, doc)
     if ($hasUpdateTag(PASTE_TAG)) nodes = $convertInlineImageDataURIs(nodes, this)
     return filterDisallowedAttachmentNodes(nodes, this)
--- a/src/helpers/lexical_helper.js
+++ b/src/helpers/lexical_helper.js
@@ -181,6 +181,41 @@ export function $textBeforeOffset(targetNode, offset) {
   return parts.join("")
 }
 
+// Lexical's LineBreakNode.importDOM drops a <br> that is the last child of a
+// block element, treating it as the phantom break browsers add to keep an
+// empty line visible in a contenteditable. That heuristic also discards a
+// genuine soft break the user typed at the end of a line (e.g. "foo<br>"),
+// so trailing breaks erode one per edit→save→re-edit cycle. We append an
+// empty <span> sentinel after a trailing <br> in blocks that have real
+// content; the break is no longer the last child, and the empty span imports
+// to nothing. Empty blocks (<p><br></p>) are left untouched so they stay
+// empty paragraphs rather than gaining a stray break.
+export function protectTrailingLineBreaks(doc) {
+  for (const block of doc.querySelectorAll("p, li, h1, h2, h3, h4, h5, h6, blockquote")) {
+    const lastChild = lastMeaningfulChild(block)
+    if (lastChild?.nodeName === "BR" && blockHasContentBeforeTrailingBreaks(block)) {
+      block.appendChild(doc.createElement("span"))
+    }
+  }
+}
+
+function lastMeaningfulChild(block) {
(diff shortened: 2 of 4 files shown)


jorgemanrubia:

🤖 Reproduced the bug in production and confirmed the fix locally using a real headless Chromium (Selenium), driving the editor's public value API — the exact import → serialize round trip that a save/re-edit cycle exercises.

**What was tested** (identical script, two environments):
1. Set the editor value to `<p>First line<br></p>` (one trailing soft break) and read `.value` back.
2. Set the editor value to `<p>First line<br><br></p>` (Shift+Enter twice) and read `.value` back.

**Production** — https://basecamp.github.io/lexxy/try-it.html (latest released CDN, no fix):
- `<p>First line<br></p>` → `<p>First line</p>` — the trailing break is dropped entirely.
- `<p>First line<br><br></p>` → `<p>First line<br></p>` — collapses two breaks to one.
- Bug reproduces: trailing breaks erode on every import.

**Local** — this branch's built assets, served by the dummy app:
- `<p>First line<br></p>` → `<p>First line<br></p>` — preserved.
- `<p>First line<br><br></p>` → `<p>First line<br><br></p>` — both preserved (visible in the Output/HTML panel).
- Bug does NOT reproduce.

The full server-side Action Text save → reload round trip can't run on the static try-it page (no posts backend there), so the production comparison is scoped to the client-side import behavior — which is precisely where the bug lives. The end-to-end keystroke round trip is covered locally by `test/system/line_break_round_trip_test.rb`.

**Verdict: Validated** (production reproduces, local fixed).

jorgemanrubia:
## AGENTS.md Guideline Review

🤖 Reviewed the diff against AGENTS.md and STYLE.md. All areas pass; no fixes were needed.

**Bug-fix workflow** — PASS. Both regression tests exist and were verified red→green. Disabling `protectTrailingLineBreaks` makes the two trailing-break Playwright tests fail (the empty-paragraph test correctly stays green), proving the tests catch the bug for the right reason.

**Test placement** — PASS.
- `test/browser/tests/editor/line_break_preservation.test.js` — the `editor/` folder is "Core editor API: …setValue…", which is exactly what these tests drive (setValue → value round trip).
- `test/system/line_break_round_trip_test.rb` — Capybara is the right home for the editor → save → render → re-edit round trip per the "Action Text Persistence" note. It mirrors `code_highlighting_test.rb`: `visit edit_post_path(posts(:empty))`, real `[:shift, :enter]` keystrokes, `find_editor`/`wait_for_editor`. Runs green (16 assertions).

**STYLE.md (JS/Lexical)** — PASS. Double quotes, no semicolons, 2-space indent, camelCase. Uses native DOM traversal (the fix operates on the raw `doc` before Lexical import, so Lexical helpers don't apply). No internal-API misuse. The explanatory comment on `protectTrailingLineBreaks` is the legitimate kind STYLE.md allows for editor code — it documents *why* the workaround exists (Lexical's `isLastChildInBlockNode` browser-quirk heuristic), not merely what the code does.

**Commit hygiene** — PASS. Only the 4 intended files (2 src, 2 test) are committed. No scratch/debug files, no `console.log`/`debugger`, no PII, no committed built assets (`app/assets/javascript/lexxy.js` is gitignored).

**Action Text round-trip** — PASS. This is a serialization/import change, and the Capybara system test exercises the full round trip end to end.

**`yarn lint`** — PASS (clean).

jorgemanrubia:
🤖 Closing without merging — trailing-break handling overlaps the server-side bc3 content pipeline (bc3#11999, "Trim trailing blank lines"), so this needs Sam's call on client-vs-pipeline ownership. Branch `fix-9955295463-linebreak-loss` is preserved if we revisit.

PR #1123, opened 2026-06-16
Fix list corruption when inserting an image into a nested bullet

jorgemanrubia:
## Problem

Adding an image to a nested sub-bullet corrupts the list formatting (Basecamp card [9962829025](https://app.basecamp.com/2914079/buckets/41746046/card_tables/cards/9962829025)):

- Inserting an image as a sub-bullet, then pasting the **next** image shifts it back to a regular top-level bullet.
- A **second** image removes the bullet entirely.

## Root cause

Image/attachment insertion routed through Lexical's default `RangeSelection.insertNodes`. A list item can only hold inline content, so dropping a block decorator (the attachment) into one makes Lexical lift it into the wrong list item and orphan an empty bullet. For nested sub-bullets the breakout logic operated on the *inner* list, stranding the attachment inside an ancestor `<li>` and mangling the nesting/indent.

## Fix

`ListItemNodeInserter` (new) intercepts block-decorator insertion when the caret is inside a list item:

- Climbs to the **outermost** list and splits it at the top-level item containing the caret, so the attachment lands at the list's own level — never inside an `<li>`.
- Nested structure (`<li>First<ul><li>Second</li></ul></li>`) stays intact; the attachment follows the whole list.
- A blank insertion-point bullet (e.g. after pressing Enter to leave the list) is removed, matching prior breakout behavior.
- Only **block decorators** (image/file attachments) are intercepted. Pasted element blocks (paragraphs, quotes) keep Lexical's own list-escape semantics, so blockquote/list paste is unaffected.

## Tests

New `test/browser/tests/attachments/image_in_list_item.test.js` covers top-level, nested, and second-image scenarios. The nested cases fail on `origin/main` (reproduce the bug) and pass with the fix. Full Playwright suite green (the one unrelated `link_opener` failure is pre-existing on `main`).

**Refactor:** the `NodeInserter` strategies were extracted from the single `node_inserter.js` into a `node_inserter/` subfolder — one file per inserter (`code_`, `shadow_root_`, `node_selection_`, `list_item_`, `block_container_`), with the base class split out so subclasses extend it without an import cycle and the entry file wiring the `NodeInserter.for` factory.

## How to reproduce

1. Create a bulleted list and add an image **as a sub-bullet** (nested list item).
2. Add a second image into that nested context, then a third.

**Before (bug):** the next pasted image drops back to a top-level bullet, and a further image removes the bullet entirely — the list nesting is corrupted and bullets are orphaned.
**After (fix):** images are inserted at the outermost list level without destroying the nested structure or dropping bullets.

### DIFF
--- a/src/editor/contents/node_inserter.js
+++ b/src/editor/contents/node_inserter.js
@@ -1,140 +1,23 @@
-import { $createLineBreakNode, $createParagraphNode, $createTextNode, $getChildCaretAtIndex, $isDecoratorNode, $isElementNode, $isLineBreakNode, $isNodeSelection, $isRangeSelection, $isTextNode, $normalizeSelection__EXPERIMENTAL as $normalizeSelection } from "lexical"
-import { CodeNode } from "@lexical/code"
-import { $ensureForwardRangeSelection } from "@lexical/selection"
-import { $getNearestNodeOfType } from "@lexical/utils"
-import { $isShadowRoot } from "../../helpers/lexical_helper"
+import BaseNodeInserter from "./node_inserter/base_node_inserter"
+import CodeNodeInserter from "./node_inserter/code_node_inserter"
+import ShadowRootNodeInserter from "./node_inserter/shadow_root_node_inserter"
+import NodeSelectionNodeInserter from "./node_inserter/node_selection_node_inserter"
+import ListItemNodeInserter from "./node_inserter/list_item_node_inserter"
+import BlockContainerNodeInserter from "./node_inserter/block_container_node_inserter"
 
-export default class NodeInserter {
-  static for(selection) {
-    const INSERTERS = [
-      CodeNodeInserter,
-      ShadowRootNodeInserter,
-      NodeSelectionNodeInserter,
-      BlockContainerNodeInserter
-    ]
-    const Inserter = INSERTERS.find(inserter => inserter.handles(selection))
-    return Inserter ? new Inserter(selection) : selection
-  }
+const INSERTERS = [
+  CodeNodeInserter,
+  ShadowRootNodeInserter,
+  NodeSelectionNodeInserter,
+  ListItemNodeInserter,
+  BlockContainerNodeInserter
+]
 
-  constructor(selection) {
-    this.selection = selection
-  }
+// Defined here rather than on the base class so the base can stay free of any
+// dependency on its subclasses (they import the base), avoiding an import cycle.
+BaseNodeInserter.for = (selection) => {
+  const inserterClass = INSERTERS.find(inserter => inserter.handles(selection))
+  return inserterClass ? new inserterClass(selection) : selection
 }
 
-class CodeNodeInserter extends NodeInserter {
-  static handles(selection) {
-    return $getNearestNodeOfType(selection.anchor?.getNode(), CodeNode)
-  }
-
-  insertNodes(nodes) {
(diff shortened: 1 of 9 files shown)


jorgemanrubia:

Tested like a human with a single Selenium script run against both environments (production CDN vs. this branch), driving the **real upload path** — clicking the toolbar image button and feeding files into the transient `<input type="file">` the editor creates.

**Steps (identical in both environments):**
1. Cleared the editor to a blank canvas.
2. Typed `First`, Enter, `Second`; selected all and applied the bullet-list toolbar command.
3. Put the caret in `Second` and pressed Tab to nest it as a sub-bullet, producing `<ul><li>First<ul><li>Second</li></ul></li></ul>` (confirmed identical in both envs).
4. With the caret at the end of the nested `Second` bullet, inserted a first image, then a second image.

**Production** (https://basecamp.github.io/lexxy/try-it.html — latest released CDN):
- Both attachments landed **inside the outer `<li>`** (2 attachments inside `<li>`), the nesting was destroyed, and an **orphaned empty bullet** was left behind. Bug reproduces.
- Evidence: `validation-production-3-after-second-image.png` (visible stray empty bullet under the images).

**Local** (this branch, `/posts/new` dummy app):
- Zero attachments inside any `<li>`; the nested structure `<ul><li value="1">First<ul><li value="1">Second</li></ul></li></ul>` stayed fully intact and both images broke out cleanly below the list. No orphaned bullet.
- Evidence: `validation-local-3-after-second-image.png`.

Production reproduces the corruption and local does not — fix confirmed end-to-end through the actual user-facing insertion path.

jorgemanrubia:
## AGENTS.md Guideline Review

Reviewed the diff against AGENTS.md and STYLE.md. All areas pass; no changes were needed.

| Area | Verdict | Notes |
|------|---------|-------|
| Bug-fix workflow (classify → failing test → fix → test+lint) | Pass | Core editing bug, correctly fixed in Lexxy (not host app). Confirmed the new test is **red on `origin/main`** (the two nested-bullet cases fail) and **green on this branch**. |
| Regression test present + correct folder | Pass | `test/browser/tests/attachments/image_in_list_item.test.js` — `attachments/` is the right bucket per the Browser Test Organization table (upload/attachment behavior). |
| STYLE.md JS/Lexical conventions | Pass | Private fields declared and used via `#`; static `handles`/instance methods ordered correctly; uses Lexical built-ins (`$splitNode`, `$getNearestNodeOfType`, `$isListNode`, `node.insertAfter`, `list.isEmpty()`); guard clauses are the legitimate selection-type-guard kind; comments explain *why* (browser/Lexical ordering constraints), not *what*. |
| No stray debug / console.log / debugger | Pass | None in the diff. |
| Commit hygiene | Pass | Only the two intended files committed; no scratch/debug files; no built assets (`app/assets/javascript/lexxy.js` is untracked); no PII. |
| `yarn lint` | Pass | Clean. |
| Broader suite (attachments / paste / formatting) | Pass | 379 passed, 1 unrelated gallery-delete test flaked and passed on retry (timing in `flush`, untouched by this change). |
| Action Text round-trip note | N/A | This change is pure client-side node insertion; it does not alter HTML serialization/sanitization, so the Capybara round-trip requirement doesn't apply. |

No clear, low-risk violations to fix; no judgment-heavy concerns to flag.

PR #1124, opened 2026-06-16
Pasting inline content onto a selected attachment crashes with Lexical error #99

jorgemanrubia:
Selecting a top-level node (such as an attachment) and then pasting or dropping content that includes inline nodes crashed the editor with **Minified Lexical error #99** — *"Only element or decorator nodes can be inserted to the root node."* In production this is one of the top JS errors in Basecamp (BC3-JS-N562: ~85k events / ~1k users in 14 days).

## What it fixes

When a node is selected, `NodeSelectionNodeInserter` inserted every incoming node directly after the selected one with `insertAfter`. When the selected node lives at the root and an incoming node is inline (a text node, a link from pasted content), `insertAfter` placed it as a direct child of the root — which Lexical forbids — throwing error #99.

## How

`NodeSelectionNodeInserter#insertNodes` now wraps inline content in a paragraph when the preceding node is at the top level, while still inserting element and decorator nodes as root siblings (preserving the existing dropped-attachment behavior).

Reproduced with a failing Playwright test in `test/browser/tests/attachments/upload_into_unsupported_selection.test.js` (asserts no error #99 and that the inline content survives); the full attachments + paste suites pass.

- Sentry: https://basecamp.sentry.io/issues/7352220817/ (BC3-JS-N562)
- Bug card: https://3.basecamp.com/4968447/buckets/41746046/card_tables/cards/10001773098

### DIFF
--- a/src/editor/contents/node_inserter.js
+++ b/src/editor/contents/node_inserter.js
@@ -108,9 +108,29 @@ class NodeSelectionNodeInserter extends NodeInserter {
     // Overrides Lexical's default behavior of _removing_ the currently selected nodes
     // https://github.com/facebook/lexical/blob/v0.38.2/packages/lexical/src/LexicalSelection.ts#L412
     let lastNode = selectedNodes.at(-1)
+    let inlineContainer = null
+
     for (const node of nodes) {
-      lastNode = lastNode.insertAfter(node)
+      if (this.#canFollowAtTopLevel(node, lastNode)) {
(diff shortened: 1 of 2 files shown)


jorgemanrubia:
🤖 Closing — superseded by #1127, which ports the same fix to the current modular `node_inserter/` layout (this PR patched the pre-refactor monolithic file and no longer merges cleanly).

PR #1127, opened 2026-06-17
Pasting inline content onto a selected attachment crashes with Lexical error #99

jorgemanrubia:
## Problem

Sentry BC3-JS-N562 — "Minified Lexical error #99" — is one of the highest-volume BC5 production exceptions (~85k occurrences / 1071 users in 14 days). Error #99 is "Only element or decorator nodes can be inserted in to the root node."

## Root cause

When a top-level node (e.g. an attachment) is node-selected, `NodeSelectionNodeInserter` inserted every node from the pasted/dropped content directly after the selected one via `insertAfter`. Inline nodes (a text node, a link) can't be direct children of the root, so the first inline node dropped straight into the root and Lexical threw error #99.

## Fix

`NodeSelectionNodeInserter` now wraps any node that isn't safe for the root in its required parent before inserting it, reusing the existing **`$makeSafeForRoot`** helper (thanks @samuelpecher) — which respects each node's declared element parent type (`createParentElementNode`) and leaves element/decorator nodes untouched.

`$makeSafeForRoot` previously assumed an already-attached node (it wraps via `$wrapNodeInElement`, which `replace`s in place). Clipboard nodes being inserted have no parent yet, so it now handles the detached case by appending the node into a fresh required parent.

## Testing

- Added a Playwright regression test under `attachments/upload_into_unsupported_selection.test.js` that node-selects an attachment, inserts an inline text node through the node-selection path, and asserts no error and valid content. Verified it fails with error #99 on the pre-fix code and passes after.
- Regression test: 40/40 on chromium, 24/24 on firefox (`--repeat-each`).
- Full `attachments` and `paste` suites pass on chromium; `yarn lint` clean.

Supersedes the stale #1124, which patched the pre-refactor monolithic `node_inserter.js` and no longer merges (the inserters were since split into `node_inserter/`).

Sentry: https://basecamp.sentry.io/issues/7352220817/ (BC3-JS-N562)

Card: https://3.basecamp.com/2914079/buckets/41746046/card_tables/cards/10001773098

## How to reproduce

The crash needs two specific conditions — missing either one is why it can look hard to hit:

1. **Insert an attachment** (upload an image or any file).
2. **Node-select it**: single-click the attachment so the whole figure gets the selection outline (a `NodeSelection`). A blinking caret next to it is a *range* selection and won't trigger this.
3. **Paste inline content** onto it — e.g. select a few words **mid-sentence** on any web page and copy, then paste. The clipboard must carry inline HTML.
   - Plain text, a bare URL, or whole-paragraph copies do **not** crash: Lexical wraps those in a paragraph (already safe for the root). It's specifically a bare inline run (text / link / span with no block wrapper) that lands directly in the root.

**Before:** the editor throws Lexical error #99 ("Only element or decorator nodes can be inserted in to the root node") and the editor breaks.
**After:** the inline content is wrapped in its required parent and inserted as a root sibling after the attachment; element/decorator nodes still insert directly. No crash.

### DIFF
--- a/src/editor/contents/node_inserter/node_selection_node_inserter.js
+++ b/src/editor/contents/node_inserter/node_selection_node_inserter.js
@@ -1,4 +1,5 @@
 import { $isNodeSelection } from "lexical"
+import { $makeSafeForRoot } from "../../../helpers/lexical_helper"
 import BaseNodeInserter from "./base_node_inserter"
 
 export default class NodeSelectionNodeInserter extends BaseNodeInserter {
@@ -7,13 +8,19 @@ export default class NodeSelectionNodeInserter extends BaseNodeInserter {
   }
 
   insertNodes(nodes) {
-    const selectedNodes = this.selection.getNodes()
-
     // Overrides Lexical's default behavior of _removing_ the currently selected nodes
     // https://github.com/facebook/lexical/blob/v0.38.2/packages/lexical/src/LexicalSelection.ts#L412
-    let lastNode = selectedNodes.at(-1)
+    let lastNode = this.selection.getNodes().at(-1)
+
     for (const node of nodes) {
-      lastNode = lastNode.insertAfter(node)
+      // Inserting after a top-level node would make this one a root child. Inline nodes
+      // can't live there (Lexical error #99), so wrap them in their required parent first.
+      const nodeToInsert = this.#insertsIntoRoot(lastNode) ? $makeSafeForRoot(node) : node
+      lastNode = lastNode.insertAfter(nodeToInsert)
     }
+  }
+
+  #insertsIntoRoot(node) {
+    return node.is(node.getTopLevelElement())
   }
 }
--- a/src/helpers/lexical_helper.js
+++ b/src/helpers/lexical_helper.js
@@ -32,8 +32,12 @@ export function $isSafeForRoot(node) {
 export function $makeSafeForRoot(node) {
   if ($isSafeForRoot(node)) {
     return node
-  } else {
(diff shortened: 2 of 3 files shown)


test/dummy/config/environments/test.rb
```diff
@@ -3,9 +3,16 @@
...
 
+require_relative "../../lib/preview_response_delay"
+
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
+  # Hold Active Storage preview (representation) responses briefly so the
+  # file-icon placeholder shown while a preview generates stays observable in
+  # system tests, rather than being swapped out before assertions can run.
+  config.middleware.use PreviewResponseDelay
```
jorgemanrubia: this workaround is not really valid...

samuelpecher:
An alternative would be the `$makeSafeForRoot` helper which should wrap an invalid root child with an element. It respects the node's declared element parent type.

jorgemanrubia:
🤖 Thanks @samuelpecher — switched to `$makeSafeForRoot`, it's much cleaner than the hand-rolled paragraph wrapping and correctly respects each node's declared parent type via `createParentElementNode` (and `isParentRequired`, which my version missed).

One small thing it surfaced: the clipboard nodes arriving here are detached, so `$makeSafeForRoot` threw "Expected node to have a parent" — `$wrapNodeInElement` `replace`s in place. I taught the helper to handle the parentless case by appending into a fresh required parent, so both the gallery (attached) and insertion (detached) paths work. Pushed.

PR #1128, opened 2026-06-17
Fix Firefox-only attachment test flakes (valid, figure-filling fallback PNG)

jorgemanrubia:
The Active Storage test mock's fallback PNG was malformed. **Firefox** (not Chromium/WebKit) rejects it with "Image corrupt or truncated" and fires the `` onerror, which reverts the deferred-preview swap back to the file icon — so the deferred-preview attachment tests flaked on Firefox CI.

A byte-valid **1×1** PNG fixes that flake but introduces a worse, consistent failure: the preview img renders at its 1px intrinsic size, far smaller than the figure's `min-inline-size` (10ch), so the figure chrome sits on top of the click point. Firefox then reports the figure as *intercepting pointer events* whenever a test clicks the preview img (delete with keyboard, delete with delete button, gallery creation, etc.).

Fix: serve a **120×90** valid PNG — wider than the figure, so the img fills the click target on every browser, while staying small enough to avoid the layout shifts the fallback exists to prevent.

## How to reproduce

On `main`, run the attachment/gallery browser tests on Firefox:

```
npx playwright test --config test/browser/playwright.config.js --project=firefox \
  test/browser/tests/attachments/
```

Tests that click an uploaded image's preview (`delete attachment with keyboard`, `delete attachment with delete button`, gallery creation) fail with `<figure …> intercepts pointer events`. With this change they pass.

Verified locally on Firefox: attachments + gallery 114/114 (`--repeat-each=3`), drag/drop suites 48/48, and Chromium attachments 96/96 — no source or timing changes needed.

### DIFF
--- a/test/browser/helpers/active_storage_mock.js
+++ b/test/browser/helpers/active_storage_mock.js
@@ -1,9 +1,13 @@
 // Mocks the Active Storage direct upload endpoints using Playwright route interception.
 // Returns a handle for asserting that the expected calls were made.
 
-// 1×1 transparent PNG used as a fallback when the fixture file doesn't exist on disk.
-const TRANSPARENT_PNG = Buffer.from(
-  "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAAC0lEQVQI12NgAAIABQABNjN9GQAAAAlwSFlzAAAWJQAAFiUBSVIk8AAAAA0lEQVQI12P4z8BQDwAEgAF/QualzQAAAABJRU5ErkJggg==",
+// Fallback preview image. Must be byte-valid — Firefox rejects malformed data and
+// fires <img> onerror, reverting the preview swap to the file icon (a flake). Must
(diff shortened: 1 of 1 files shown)


src/nodes/action_text_attachment_node.js
```diff
@@ -330,13 +332,27 @@ export class ActionTextAttachmentNode extends DecoratorNode {
...
+  // Production defaults, overridable via data-preview-poll-* on the host
+  // <lexxy-editor> so tests poll fast and deterministically.
+  #previewPollTiming() {
+    const host = this.editor.getRootElement()?.closest("lexxy-editor")
+    const initialDelay = this.#numericDataset(host, "previewPollInitialDelay", INITIAL_PREVIEW_POLL_DELAY_MS)
+    const baseDelay = this.#numericDataset(host, "previewPollBaseDelay", PREVIEW_POLL_BASE_DELAY_MS)
+    return { initialDelay, baseDelay }
+  }
+
+  #numericDataset(element, key, fallback) {
```
jorgemanrubia: renaming these

PR #1133, opened 2026-06-17
Dismiss prompt popover when the cursor moves away from the trigger

jorgemanrubia:
## What

Closes Basecamp card [9941546721](https://3.basecamp.com/2914079/buckets/41746046/todos/9941546721): "[Mentions] Prompts should be dismissed if cursor moves away".

When a prompt popover is open and the cursor moves to a *later* part of the text (mouse click or right-arrow), the popover should close — just as it already does when the cursor moves to *before* the trigger. Previously it stayed open.

## Root cause

`#addCursorPositionListener` in `src/elements/prompt.js` decided whether to dismiss by searching for the last trigger behind the cursor and hiding only when the cursor was at or before the trigger end. When the cursor moved *past* the trigger token into later text, the trigger was still found behind the cursor, so the popover was kept open.

## Fix

The popover should stay open only while the caret sits at the end of an unbroken `trigger + search-term` run (the search term being the contiguous text after the trigger, with no whitespace — matching how the editor extracts the filter). The dismissal check now extracts that candidate search term and hides the popover whenever it contains a space or newline, which covers the caret leaving the trigger's range in *either* direction.

## Tests

Added `test/browser/tests/prompts/cursor_moves_away.test.js` with two cases:
- caret moves after the trigger token (the bug — fails before the fix)
- caret moves before the trigger token (existing behavior — guards against regression)

The full Playwright prompts suite passes; `yarn lint` is clean.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -15,6 +15,7 @@ Follow [STYLE.md](STYLE.md) for all code you write — general conventions, Java
 When working from a git worktree:
 
 - Run all commands from that worktree, not from the main checkout.
+- A fresh worktree starts with no installed dependencies and an empty dummy database. Before `bin/dev` will serve the dummy app, run `bin/setup` (installs Ruby/JS dependencies and builds the assets), then prepare and seed its database with `cd test/dummy && bin/rails db:prepare db:seed` (without the seed, the `@` prompt has no people to show). Skipping these leaves `bin/dev` returning a 500 (`ActiveRecord::PendingMigrationError`).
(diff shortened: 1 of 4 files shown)


jorgemanrubia:
## Manual validation: ✅ Validated

- **Prod (released build):** caret moved past a non-space trigger → popover stays open (bug reproduces).
- **Local (this branch):** same steps → popover dismisses as expected.
- **Regression guard:** `person:` (`supports-space-in-searches`) keeps the popover open across spaces (`person:emma johnson`).

PR #1134, opened 2026-06-17
Abort in-flight uploads when their attachment is removed

jorgemanrubia:
## What

Cancelling an attachment upload — trashing it before it finishes — left the upload running in the background until it completed or timed out. Now removing an in-progress attachment aborts the upload immediately (matching how Docs & Files behaves).

## How

- The upload node captures the in-flight `XMLHttpRequest` via Active Storage's `directUploadWillStoreFileWithXHR` delegate.
- In-flight requests are tracked by a small `UploadRequests` class (`src/editor/attachments/upload_requests.js`) — one instance per editor, owned by the `<lexxy-editor>` element and exposed as `editor.uploadRequests`, keyed by node key (the only handle the destroyed-node mutation gives us).
- When the upload node is destroyed (its attachment is removed), the attachments extension calls `editor.uploadRequests.abort(key)`, aborting the captured request. Completed uploads forget their request, and the structure is cleared on editor reset.

## Tests

- Vitest unit test for `UploadRequests` (track / forget / abort / clear).
- Playwright regression test that holds an Active Storage disk upload open, trashes the attachment mid-flight, and asserts the request is aborted.

Basecamp card: https://app.basecamp.com/2914079/buckets/41746046/card_tables/cards/10002095083

### DIFF
--- a/src/editor/attachments/upload_requests.js
+++ b/src/editor/attachments/upload_requests.js
@@ -1,0 +1,23 @@
+export class UploadRequests {
+  #requestsByKey = new Map()
+
+  track(key, request) {
+    this.#requestsByKey.set(key, request)
+  }
+
+  forget(key) {
+    this.#requestsByKey.delete(key)
+  }
+
+  abort(key) {
+    const request = this.#requestsByKey.get(key)
+    if (request) {
+      this.#requestsByKey.delete(key)
+      request.abort()
+    }
+  }
+
+  clear() {
+    this.#requestsByKey.clear()
+  }
+}
--- a/src/elements/editor.js
+++ b/src/elements/editor.js
@@ -15,6 +15,7 @@ import { registerMarkdownLeadingTagHandler } from "../editor/markdown/leading_ta
 
 import theme from "../config/theme"
 import { HorizontalDividerNode } from "../nodes/horizontal_divider_node"
+import { UploadRequests } from "../editor/attachments/upload_requests"
 import { CommandDispatcher } from "../editor/command_dispatcher"
 import Selection from "../editor/selection"
 import { createElement, dispatch, generateDomId, parseHtml } from "../helpers/html_helper"
@@ -62,11 +63,16 @@ export class LexicalEditorElement extends HTMLElement {
 
   #validity = new Map()
   #validationTextArea = document.createElement("textarea")
+  #uploadRequests
 
   constructor() {
     super()
     this.internals = this.attachInternals()
     this.internals.role = "presentation"
+  }
+
+  get uploadRequests() {
+    return this.#uploadRequests
   }
 
   connectedCallback() {
@@ -89,6 +95,7 @@ export class LexicalEditorElement extends HTMLElement {
     this.#disposables.push(this.clipboard)
 
     this.adapter = new BrowserAdapter()
+    this.#uploadRequests = new UploadRequests()
 
     const commandDispatcher = CommandDispatcher.configureFor(this)
     this.#disposables.push(commandDispatcher)
@@ -869,6 +876,7 @@ export class LexicalEditorElement extends HTMLElement {
   #reset() {
     this.#dispose()
     this.#resetValidity()
+    this.#uploadRequests?.clear()
     this.editorContentElement?.remove()
     this.editorContentElement = null
 
--- a/src/extensions/attachments_extension.js
+++ b/src/extensions/attachments_extension.js
@@ -48,11 +48,12 @@ export class AttachmentsExtension extends LexxyExtension {
 
   #handleUploadMutations(mutations) {
     const previousUploadsCount = this.#uploadsCount
-    for (const [ , mutation ] of mutations) {
+    for (const [ key, mutation ] of mutations) {
       if (mutation === "created") {
         this.#uploadsCount++
       } else if (mutation === "destroyed") {
         this.#uploadsCount--
+        this.editorElement.uploadRequests.abort(key)
       }
     }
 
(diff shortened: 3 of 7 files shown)


jorgemanrubia:

Drove the real editor at `/posts/new` (local PR branch) with Selenium + CDP network throttling (upload ~30 KB/s) so a 12 MB attachment's Active Storage disk `PUT` stayed in flight, instrumenting `XMLHttpRequest` to log the disk-PUT lifecycle.

**Local (fixed):**
- Injected the file, waited for the disk `PUT` to fire `loadstart` (in flight, no `loadend`).
- Selected the figure and pressed Delete; the attachment was removed (figures 1 → 0).
- Immediately observed `XMLHttpRequest.abort()` called on the disk PUT and its `abort` event firing (`readyState 4`, `status 0`). `load`/`loadend` never fired — the upload was stopped mid-flight, not completed.

**Counterfactual (pre-fix behavior, same server/file/throttle):**
- Neutralized `abort()` on the disk PUT to emulate the world before this change. After removing the attachment the disk PUT stayed pending — no `abort`, no `load`, no `error`. The connection lived on exactly as the card describes.

**Production note:** the `try-it` page serves uploads from a service worker that returns 204 instantly, so the PUT completes in ~13 ms and can never be caught in flight there — not a faithful environment for this bug. Used the controlled local counterfactual above as the broken baseline instead.

jeremy:
🤖 **Heads-up: this PR introduced a high-volume production TypeError in BC5.**

Sentry [BC3-JS-N8HN](https://basecamp.sentry.io/issues/7567796249/) (13.9k events / 9.8k users, `forgetUploadRequest_fn`) and [BC3-JS-N8HP](https://basecamp.sentry.io/issues/7567802467/) (2.4k / 2.0k, `rememberUploadRequest_fn`) — ~16.4k events / ~11.8k users total. Causality is tight: first shipped in lexxy 0.9.19, BC5 bumped Jun 22 at 13:00Z (bc3 #12078), first event 13:07Z the same day; 0.9.19-alpha.2 (without this PR) ran Jun 17–22 with zero events.

**Mechanism: an in-flight direct upload outlives the editor.** `<lexxy-editor>` runs `#reset()` on `disconnectedCallback` *and* on every `turbo:before-cache` — i.e. any Turbo navigation while an upload is in flight. Teardown nulls the Lexical root, so `get #editorElement` (`getRootElement()?.closest("lexxy-editor")`) returns `undefined` when the ActiveStorage callbacks land afterwards:

- store PUT completes after teardown → `create` callback → `#forgetUploadRequest()` → TypeError (the 13.9k fingerprint)
- blob-create POST resolves after teardown → `directUploadWillStoreFileWithXHR` → `#rememberUploadRequest()` → same TypeError (the 2.4k fingerprint)

(Sentry shows only the `undefined` variant, never `of null` — consistent with full teardown, not a merely-detached content element.)

**Deterministic repro** in this repo's Playwright harness (both paths reproduce the exact production message):

<details>
<summary><code>test/browser/tests/attachments/upload_survives_teardown.test.js</code></summary>

```js
import { test } from "../../test_helper.js"
import { expect } from "@playwright/test"
import { mockActiveStorageUploads } from "../../helpers/active_storage_mock.js"

test.describe("Upload outliving editor teardown", () => {
  test.beforeEach(async ({ page }) => {
    await page.goto("/attachments.html")
    await page.waitForSelector("lexxy-editor[connected]")
    await page.waitForSelector("lexxy-toolbar[connected]")
  })

  test("file upload completing after teardown does not throw (forget path)", async ({ page, editor }) => {
    await mockActiveStorageUploads(page, { uploadDelayMs: 1500 })

    const pageErrors = []
    page.on("pageerror", (error) => pageErrors.push(error))

    const diskUploadStarted = page.waitForRequest((request) =>
      request.url().includes("/rails/active_storage/disk/") && request.method() === "PUT",
    )

    await editor.uploadFile("test/fixtures/files/example.png")
    await diskUploadStarted

    await page.evaluate(() => document.querySelector("lexxy-editor").remove())

    await page.waitForTimeout(2500)
    expect(pageErrors.map(String)).toEqual([])
  })

  test("blob creation resolving after teardown does not throw (remember path)", async ({ page, editor }) => {
    const calls = await mockActiveStorageUploads(page, { delayDirectUploadResponse: true })

    const pageErrors = []
    page.on("pageerror", (error) => pageErrors.push(error))

    await editor.uploadFile("test/fixtures/files/example.png")
    await expect.poll(() => calls.blobCreations.length).toBeGreaterThan(0)

    await page.evaluate(() => document.querySelector("lexxy-editor").remove())
    await calls.releaseDirectUploadResponses()

    await page.waitForTimeout(2000)
    expect(pageErrors.map(String)).toEqual([])
  })
})
```
</details>

**Verified fix** — both repros green, this PR's `upload_abort_on_delete` test still green, and no downstream error surfaces (`#dispatchEvent` already no-ops via `getElementByKey`, `editor.update` on a torn-down editor is inert):

```diff
   #forgetUploadRequest() {
-    this.#editorElement.uploadRequests.forget(this.getKey())
+    this.#editorElement?.uploadRequests.forget(this.getKey())
   }

   #rememberUploadRequest(request) {
-    this.#editorElement.uploadRequests.track(this.getKey(), request)
+    this.#editorElement?.uploadRequests.track(this.getKey(), request)
   }
```

**Worth considering while here:** the leak this PR fixed for attachment-removal still exists for editor teardown — `#reset()` calls `#uploadRequests?.clear()`, which drops the map without aborting, so the store PUT keeps uploading into the void after every navigation. An `abortAll()` on teardown would stop the waste, but it can't replace the guards: the blob-create POST isn't tracked, so the `remember` path crashes regardless.

Tracked as BC-10102903433 on the T2 board. Happy to open a PR with the guards + regression test.

— posted by Jeremy's coworker agent

PR #1136, opened 2026-06-17
Allow dragging mentions to reorder them within a line

jorgemanrubia:
## What

Fixes Basecamp card 9974786701 — "[Mentions] Can no longer re-arrange mentions in chat".

In Basecamp 4 (Trix) you could drag a @mention around a chat line while editing to reposition it within the sentence. In Basecamp 5 (Lexxy) mentions could not be dragged at all.

## Root cause

Mentions render as inline `CustomActionTextAttachmentNode` decorator nodes (`<action-text-attachment data-lexxy-decorator>`). Two things prevented dragging:

1. The mention DOM was never marked `draggable`, and carried no `data-lexical-node-key`, so a drag couldn't start or be resolved back to a node.
2. The existing `AttachmentDragAndDrop` only handles block-level `figure.attachment` nodes — gallery merges and between-block insertion. It bails for inline mentions and has no notion of dropping at a text caret within a line.

## Fix

- Mark the mention DOM `draggable` and tag it with `data-lexical-node-key` in `CustomActionTextAttachmentNode#createDOM`.
- Add a `MentionDragAndDrop` controller (and a `MentionDragAndDropExtension`, enabled when the editor supports rich text) that handles `DRAGSTART`/`DROP` for inline mentions at HIGH priority. On drop it resolves the text caret from the drop point and places the mention there using Lexical's `$createRangeSelectionFromDom`.
- Add grab-cursor / dragging-opacity styles for inline decorator attachments.

The block attachment handler still bails for mentions (they aren't `figure.attachment`), so the two handlers coexist without conflict.

## Tests

Added `test/browser/tests/prompts/mention_drag_reorder.test.js`, which inserts a mention, asserts it is draggable, drags it to the end of the line, and asserts the serialized order changed. Confirmed failing before the fix (mention not draggable) and passing after.

Full Playwright suite passes on chromium and firefox. (webkit can't launch in this sandbox; one unrelated `link_opener` test requires real network access to `example.com`.)

### DIFF
--- a/app/assets/stylesheets/lexxy-editor.css
+++ b/app/assets/stylesheets/lexxy-editor.css
@@ -249,6 +249,22 @@
     opacity: 0.4;
   }
 
+  [data-lexxy-decorator][draggable] {
+    cursor: grab;
+  }
+
+  [data-lexxy-decorator].lexxy-dragging {
+    opacity: 0.4;
+  }
+
+  .lexxy-drop-caret {
+    background-color: var(--lexxy-focus-ring-color);
+    border-radius: 1px;
+    inline-size: 2px;
+    pointer-events: none;
+    position: fixed;
+  }
+
   [class*="lexxy-drop-target--"] {
     position: relative;
   }
--- a/src/editor/attachments/custom/drag_and_drop.js
+++ b/src/editor/attachments/custom/drag_and_drop.js
@@ -1,0 +1,234 @@
+import {
+  $createRangeSelectionFromDom,
+  $getNodeByKey,
+  $setSelection,
+  COMMAND_PRIORITY_HIGH,
+  DRAGSTART_COMMAND,
+  DROP_COMMAND
+} from "lexical"
+
+import { $isCustomActionTextAttachmentNode } from "../../../nodes/custom_action_text_attachment_node"
+import { caretFromPoint, caretRect } from "../../../helpers/caret_helpers"
+import { createElement } from "../../../helpers/html_helper"
+import { ListenerBin } from "../../../helpers/listener_helper"
+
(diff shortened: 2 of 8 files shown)


jorgemanrubia:
**Manual validation: PASS** (drag-reorder verified)

Validated against local PR build (`fix-mention-drag-reorder`) and the released code on `main`.

**Local (fixed):** Inserted "hello @Emma Johnson there" in the sandbox editor.
- The inline mention DOM is now `draggable="true"` and carries `data-lexical-node-key` (`<action-text-attachment ... draggable="true" data-lexical-node-key="110">`) — both added by `createDOM`.
- Drove a full HTML5 drag sequence (real `DataTransfer`, `dragstart`/`dragover`/`drop`) from the chip onto the trailing "there" span. `dragover` was `preventDefault`'d by the new `MentionDragAndDrop` handler and the mention reordered: before `hello / [mention] / there` → after `hello there / [mention]`. The editor and Output HTML both reflect the new order.

**Production (reproduces):** Confirmed at the source level — on `main`, `CustomActionTextAttachmentNode#createDOM` emits no `draggable` attribute and no `data-lexical-node-key`, and the `MentionDragAndDrop` controller/extension don't exist, so inline mentions can't be dragged. The public "Try it" demo has no `@`-mention prompt wired, so a released mention chip couldn't be exercised in-browser there; no live BC5 instance was reachable from this environment.

Caveat: A synthetic mouse drag (`left_click_drag`) did not trigger a reorder (native HTML5 DnD isn't driven by synthetic mouse moves); the reorder was verified via a dispatched DataTransfer drag sequence, which exercises the PR's actual DROP handler. A quick human drag-with-the-mouse confirmation in BC5 chat is still worth a glance.

PR #1137, opened 2026-06-17
Fix mention insertion retroactively applying earlier formatting

jorgemanrubia:
## Problem

Basecamp card 9741130457: when a line starts with a formatted word (bold, italic, etc.) followed by plain unformatted text, inserting an @mention into that sentence makes everything between the first word and the mention inherit the first word's formatting.

Example — before: "*Italics* not applied to words other than the first"; after adding a mention the whole span becomes italic: "*Italics not applied to words other than the first*".

## Root cause

Inserting a mention splits the anchor text node around the `@` trigger via `Contents#replaceTextBackUntil` → `#performTextReplacement` → `#cloneTextNodeFormatting`. Both halves hold text that already existed, so they should keep the anchor node's own format/style.

Instead, `#cloneTextNodeFormatting` preferred the live selection's typing format and, when the node format was `0`, fell back to the paragraph's text format (`parent.getTextFormat()`). Because a plain format is `0` (falsy), the `anchorNode.getFormat() || fallbackFormat` expression always fell through to the paragraph format, which here was italic. The reconstructed plain text was created as italic and merged with the leading `<em>` node, italicizing the whole line (the trailing spacer became italic too).

## Fix

`#cloneTextNodeFormatting` now copies the anchor node's actual `getFormat()` / `getStyle()` directly, with no selection-format preference and no falsy `0` fallback. The selection argument is no longer needed and was removed from the call chain.

## Test

Added `test/browser/tests/prompts/mention_preserves_earlier_formatting.test.js`, which sets up "*Italics* not applied to words other than the first", inserts a mention, and asserts only "Italics" remains italic. It fails before the fix and passes after.

Full Playwright suite passes on chromium (525 passing; the one unrelated `link_opener` failure is a pre-existing environment issue that also fails on `main`). `eslint` is clean.

### DIFF
--- a/src/editor/contents.js
+++ b/src/editor/contents.js
@@ -244,14 +244,13 @@ export default class Contents {
   replaceTextBackUntil(stringToReplace, replacementNodes) {
     replacementNodes = Array.isArray(replacementNodes) ? replacementNodes : [ replacementNodes ]
 
-    const selection = $getSelection()
     const { anchorNode, offset } = this.#getTextAnchorData()
     if (!anchorNode) return
 
     const lastIndex = this.#findReplacementStart(anchorNode, offset, stringToReplace)
     if (lastIndex === -1) return
 
-    this.#performTextReplacement(anchorNode, selection, lastIndex, stringToReplace, replacementNodes)
+    this.#performTextReplacement(anchorNode, lastIndex, stringToReplace, replacementNodes)
   }
 
   uploadFiles(files, { selectLast } = {}) {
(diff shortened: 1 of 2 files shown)


jorgemanrubia:

Drove both environments with Selenium, like a human: set a line to italic "Italics" + plain "not applied to words other than the first", put the caret in the plain text, triggered a prompt, typed to filter, and pressed Enter to insert the item. Then inspected the editor's serialized value and which text sits inside `<em>`.

- **Production** (`basecamp.github.io/lexxy/try-it.html`, released CDN): the `:` emoji prompt was used because the try-it page ships no `@` mention prompt — but it exercises the same shared code path (`replaceTextBackUntil` → `#performTextReplacement` → `#cloneTextNodeFormatting`). Bug reproduces: after inserting, the whole line became italic — `<p><em>Italics not applied to words other than the first </em>😂<em> </em></p>`.
- **Local** (PR branch `fix/mention-retroactive-formatting`): used the real `@` mention prompt and selected Emma Johnson. Fixed: only the first word stays italic — `<p><em>Italics</em> not applied to words other than the first <action-text-attachment …Emma Johnson… content-type="application/vnd.actiontext.mention"></action-text-attachment> </p>`. The plain run is outside any `<em>`.

Production shows the broken behavior, local shows it corrected.

PR #1138, opened 2026-06-17
Preserve plain text spacing and backslashes on paste

jorgemanrubia:
## Problem

Pasting a plain-text Windows/UNC file path with more than one consecutive space (or backslashes) into Lexxy silently corrupts it: consecutive spaces collapse to one and leading backslashes are stripped.

Reported in Basecamp card 10006219528. Example pasted by the customer:

```
\\Arina-alvand\alvand\03 - ENGINEERING\TN32 -NSC YAWATA\00. Input\00.Original\2026-06-17
```

## Root cause

Plain text pasted without an HTML payload routes through the Markdown conversion path (`Clipboard#pasteMarkdown`). `marked()` + HTML insertion (a) collapses runs of whitespace per HTML whitespace rules and (b) unescapes backslashes (markdown escaping). For text that isn't actually Markdown, this rewrites the user's content.

## Fix

When the Markdown conversion produces nothing but a single text-only paragraph, there is no Markdown structure to honor, so we paste the original text verbatim via a new `Contents#insertText`. Genuine Markdown (formatting, headings, lists, links) is unaffected.

The new insert reuses the existing `NodeInserter` machinery (via `insertAtCursor`), so it continues to handle the quote-element-point edge cases that a raw `insertRawText` would crash on (#212 / #1109).

## Tests

Added `test/browser/tests/paste/plain_text_paths.test.js` covering:
- UNC path with consecutive spaces and backslashes is preserved (no autolink, exact characters)
- runs of consecutive spaces preserved
- genuine Markdown still converts

Full Playwright suite passes on chromium and firefox. (webkit can't launch in this sandbox; the only chromium failure, `link_opener` modifier+click, requires real network to example.com and also fails on a clean `main`.)

### DIFF
--- a/src/editor/clipboard.js
+++ b/src/editor/clipboard.js
@@ -168,17 +168,16 @@ export default class Clipboard {
 
     if (this.#isPlainTextWithoutMarkdown(doc)) {
       this.contents.insertText(text, { tag: PASTE_TAG })
-      return
+    } else {
+      const detail = Object.freeze({
+        markdown: text,
+        document: doc,
+        addBlockSpacing: () => addBlockSpacing(doc)
+      })
+
+      dispatch(this.editorElement, "lexxy:insert-markdown", detail)
+      this.contents.insertDOM(doc, { tag: PASTE_TAG })
     }
-
-    const detail = Object.freeze({
-      markdown: text,
-      document: doc,
-      addBlockSpacing: () => addBlockSpacing(doc)
-    })
-
-    dispatch(this.editorElement, "lexxy:insert-markdown", detail)
-    this.contents.insertDOM(doc, { tag: PASTE_TAG })
   }
 
   // Markdown conversion collapses runs of whitespace and unescapes backslashes,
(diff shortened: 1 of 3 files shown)


jorgemanrubia:

Validated by simulating a real plain-text paste (a `paste` ClipboardEvent carrying only `text/plain`) on both environments, then comparing the pasted UNC path before/after. Also confirmed genuine Markdown still converts.

Test path (note the leading `\\` and the runs of consecutive spaces in `00.  Input` and `Shanks   05`):
`\\server\share\03 - ENGINEERING\00.  Input\Consulting Agreement - Michael Shanks   05.22.2026(282407).docx`

**Production** (try-it.html, released build) — bug reproduces:
- Output: `\server\share\03 - ENGINEERING\00. Input\Consulting Agreement - Michael Shanks 05.22.2026(282407).docx`
- Leading `\\` collapsed to a single `\`, and the consecutive spaces collapsed to single spaces. Mangled.

**Local** (PR branch `fix-paste-preserves-plain-text-paths`) — fixed:
- Output matches the input verbatim: leading `\\` intact, `00.  Input` double-space intact, `Shanks   05` triple-space intact, 0 autolinks.

**Markdown regression check** — pasting `Hello **there**` produced a `<strong>` in both environments, so genuine Markdown paste still converts.

Verdict: Validated.

PR #1122, opened 2026-06-16
Fix list editing inside blockquotes

jorgemanrubia:
Fixes two related defects when editing bullet lists inside a blockquote, reported on a Basecamp bug card.

## Backspace on a blank list item

In quoted text with bullets, pressing Backspace on an empty first list item stranded a `<p><br></p>` above the list instead of closing the gap — so the blank line could not be removed (forward Delete worked, Backspace did not). Inside a quote we now remove the empty item and move the cursor into the next one, matching the "close the gap" expectation. The root-level exit-list behavior (convert first item to a paragraph above the list) is unchanged.

## Adding a list to a quoted paragraph

Lexical's `$insertList` climbs to the nearest root or shadow root, treating a `QuoteNode` as transparent: listing a quoted paragraph swallowed the whole blockquote into a single list item and destroyed the quote. When the selection lives inside a quote we now build the list at the paragraph level ourselves, listing only the selected blocks and leaving the surrounding quote intact. Toggling a list off inside a quote and root-level list behavior are unaffected.

## Tests

New Playwright coverage in `test/browser/tests/formatting/list_in_quote.test.js` for both cases (backspace gap closing for blank/cleared first items; inserting unordered/ordered lists on quoted paragraphs; re-adding bullets after removing them). All five fail on clean `main` and pass with this change. Existing formatting, paste, list, and editor suites pass; `yarn lint` clean.

---
**Basecamp card:** https://app.basecamp.com/2914079/buckets/41746046/card_tables/cards/9961874479

## How to reproduce

Use a blockquote that contains a bulleted list (paste or build one):

**Issue 1 — backspace can't close the gap:**
1. Inside a quote, have a few bullet items. Delete the text of the first bullet item so it's an empty line.
2. Press **Backspace** to remove that empty line.
   - **Before (bug):** Backspace won't close the gap (only the forward Delete key worked).
   - **After (fix):** Backspace removes the empty item and joins the line as expected.

**Issue 2 — adding bullets breaks the quote:**
1. Inside a quote, place the caret on a non-bulleted line (e.g. after deleting some bullets).
2. Toggle the bullet-list button to add bullets.
   - **Before (bug):** the whole blockquote is swallowed into a single list item / the quote is destroyed.
   - **After (fix):** the selected line(s) become a list *inside* the quote; the blockquote stays intact.

PR #1140, opened 2026-06-17
Fix listing a blockquote that holds inline content directly

jorgemanrubia:
## Problem

Basecamp card #10005355930: "Selecting bullet inside blockquote erases previous content."

The headline symptom from the card — converting a quoted paragraph to a list swallowing the whole blockquote and dropping the content — was already fixed on `main` by #1122 ("Fix list editing inside blockquotes"). While reproducing, I found a remaining gap in the same feature:

When a blockquote holds its content as a **direct text node** rather than wrapped in a paragraph — e.g. legacy Trix/Action Text content like `<blockquote>text</blockquote>` — pressing the bullet or numbered list button did **nothing**. The quote had no inner block element for the list to wrap (`#quotedBlocksInSelection` filters for blocks whose parent is a quote, but here the quote *is* the block), so the conversion silently no-opped and the content stayed unlisted.

## Fix

Before wrapping quoted blocks in a list, move any inline content held directly by a quote into a paragraph. The direct-text quote then becomes a quoted block like any other and is listed in place, keeping the surrounding quote intact and preserving line breaks.

## Test

Added `test/browser/tests/formatting/bug_bullet_in_quote_erases_content.test.js`, which lists a `<blockquote>First line</blockquote>` and asserts the result is `<blockquote><ul><li value="1">First line</li></ul></blockquote>`. Confirmed it fails before the fix (quote left unchanged) and passes after.

The full Playwright suite passes on chromium (524 passing; the two unrelated failures — `link_opener`, which needs outbound network to example.com, and the flaky `leak` test — also fail/flake on clean `main` in this sandbox). `yarn lint` is clean.

### DIFF
--- a/src/editor/contents.js
+++ b/src/editor/contents.js
@@ -423,8 +423,26 @@ export default class Contents {
   }
 
   #insertListInsideQuote(listType) {
+    this.#wrapDirectInlineQuoteContentInParagraphs()
+
     for (const group of $consecutiveSiblingGroups(this.#quotedBlocksInSelection())) {
       this.#wrapBlocksInList(group, listType)
+    }
+  }
+
+  // A quote that holds inline content directly (e.g. legacy Trix content like
+  // <blockquote>text</blockquote>) has no inner block for the list to wrap.
+  // Move that content into a paragraph so it becomes a quoted block like any other.
+  #wrapDirectInlineQuoteContentInParagraphs() {
+    const selection = $getSelection()
+    if (!$isRangeSelection(selection)) return
+
+    for (const block of this.#outermostElements(this.#blockLevelElementsInSelection(selection))) {
+      if ($isQuoteNode(block)) {
+        const paragraph = $createParagraphNode()
+        paragraph.append(...block.getChildren())
+        block.append(paragraph)
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


jorgemanrubia:

Validated with Selenium against released production (`@latest` = `0.9.19-alpha.3` via jsDelivr on the try-it page) and the local PR branch, driving the editor through real browser events.

**#1140 case (direct-text blockquote)** — loaded legacy content `<blockquote>First line</blockquote>` (content as a direct text node), placed the caret in the quote, and clicked the bullet-list toolbar button:

- **Production (broken):** the blockquote is erased — result is `<ul><li value="1">First line</li></ul>` with no surrounding quote. The list applies but the quote wrapper is lost. (Note: this is a slightly different broken manifestation than the "did nothing" described in the PR body; on this released alpha the click is not a no-op, it drops the quote. Either way the quote is not preserved.)
- **Local (fixed):** result is exactly `<blockquote><ul><li value="1">First line</li></ul></blockquote>` — quote preserved, list created in place, text intact.

**Headline case (#1122, already fixed on `main`)** — from an empty doc, typed a `>` quote with "First line" / "Second line" and applied the bullet list. Local produced `<blockquote><ul><li value="1">First line</li></ul><p>Second line</p></blockquote>`, confirming the headline flow still works and the new `#wrapDirectInlineQuoteContentInParagraphs` step doesn't regress it.

Verdict: production reproduces broken quote handling, local is fixed.

jorgemanrubia:
Closing without merging. The originally reported problem (listing a line inside a blockquote erasing its content) was already fixed on main by #1122, which is in production. This PR only covered an adjacent legacy direct-text-blockquote edge case that isn't worth shipping separately.

PR #1143, opened 2026-06-18
Balance line-break spacing around pasted lists

jorgemanrubia:
## What

Follow-up to #1119 (which stripped invalid child nodes from pasted lists, fixing the spurious leading "1.").

After that fix, the spacing *around* a pasted numbered list was wrong: pasting a numbered list from HelpScout produced **two** blank lines before the list and **none** after — instead of the single blank line on each side that matched the original content.

Reported by QA on Basecamp card [9939030998](https://app.basecamp.com/2914079/buckets/41746046/card_tables/cards/9939030998).

## Why

HelpScout (and similar sources) wrap a numbered list in stray line breaks, e.g.:

```html
<div>Intro<br><br><ol><li>…</li></ol><br><br>Tail</div>
```

The number of breaks varies by source, and Lexical's clipboard insertion drops some leading/trailing breaks while keeping others. The result was lopsided: a paragraph ending in `<br><br>` before the list, and zero separation after it.

## How

Added a `PastedNodesNormalizer` class (mirroring the existing `PastedContentFormatter`), applied to the top-level nodes generated from the pasted DOM (paste path only). For each list it:

- Trims the trailing line breaks of the preceding paragraph down to one.
- Ensures the content following the list is separated by exactly one blank line (trimming extra leading breaks, or wrapping a trailing inline run into a paragraph with a single leading break).

A list with no content after it (the common "paste a list at the end" case) is left untouched, so the existing #1119 behavior is preserved.

Where `PastedContentFormatter` cleans the source DOM *before* conversion, `PastedNodesNormalizer` runs on the generated nodes *after* conversion — because the DOM→node conversion is what drops some breaks while keeping others, so the spacing can only be balanced post-conversion.

### Before / after (HelpScout-style paste)

Before:
```html
<p>Intro<br><br></p>
<ol>…</ol>
<p>Tail</p>
```

After:
```html
<p>Intro<br></p>
<ol>…</ol>
<p><br>Tail</p>
```

## Tests

Playwright tests live in `test/browser/tests/paste/pasted_text_formatting.test.js`, which groups the list-HTML cleaning (#1119) and the spacing-normalization cases under one "Pasted text formatting" suite. The full Playwright paste suite passes.


jorgemanrubia:
Closing this. After reproducing the actual HelpScout paste end-to-end on a local build, the numbered-list paste already renders cleanly — the intro paragraph, the `<ol>`, and the trailing paragraph are correctly separated with no stray blank lines. The spacing issue isn't reproducible against current behavior, so this change isn't needed.

PR #1145, opened 2026-06-18
Strip leading empty list item when pasting numbered lists

jorgemanrubia:
## What

Pasting a numbered (ordered) list copied from HelpScout into the Lexxy editor inserted a spurious extra `1.` and a line break before the actual list. The user's source list `New For You / Bubbled Up / Previous Notifications` came out renumbered, with an empty leading item.

## Root cause

HelpScout (and similar rich-text editors) prefix a copied ordered list with an empty first `<li>` (often containing only a `<br>`):

```html
<ol><li><br></li><li>New For You</li><li>Bubbled Up</li><li>Previous Notifications</li></ol>
```

`PastedContentFormatter#stripStrayListChildren` already dropped non-`<li>` children of a list, but an empty `<li>` is a valid child, so it survived. On import it became item `1.` (empty, with a line break), pushing every real item down by one.

## Fix

Added `PastedContentFormatter#stripLeadingEmptyListItems`, which drops empty `<li>` elements at the front of a pasted list. It preserves intentional empty items further down (existing behavior) and leaves a leading item alone if it only holds a nested sublist.

## Tests

Added a Playwright test in `test/browser/tests/paste/clean_invalid_pasted_list_html.test.js` reproducing the exact HelpScout shape and asserting the result is the three items with correct numbering and no empty leading item. Confirmed it failed before the fix.

Basecamp card 9939030998.

### DIFF
--- a/src/editor/contents/pasted_content_formatter.js
+++ b/src/editor/contents/pasted_content_formatter.js
@@ -7,6 +7,7 @@ export default class PastedContentFormatter {
     this.#unwrapPlaceholderAnchors()
     this.#stripTableCellColorStyles()
     this.#stripStrayListChildren()
+    this.#stripLeadingEmptyListItems()
     return this.doc
   }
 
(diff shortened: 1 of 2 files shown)


jorgemanrubia:
Closing as redundant/off-target for card [9939030998](https://app.basecamp.com/2914079/buckets/41746046/card_tables/cards/9939030998).

- The leading-empty-`<li>` stripping this adds is already covered by the merged #1119 ("Strip invalid child nodes from pasted lists").
- The actual problem QA detailed on the card is the **line-break spacing** around a pasted list (two blank lines before, none after) — there's no empty `<li>` involved. That's addressed by #1143 ("Balance line-break spacing around pasted lists").

Superseded by #1119 + #1143.

PR #1153, opened 2026-06-23
Code block and quote grab the block above the selection

jorgemanrubia:
Converting existing content into a code block could swallow the block sitting *above* the selection. Write a title line like `ABCD:`, paste a few lines beneath it, select only the pasted text, and hit the code-block button — the title line gets pulled into the code block too. The same thing happened when wrapping that selection in a quote.

The trouble shows up when the browser anchors the selection at the *end* of the line above rather than at the start of the selected content — a very common shape when you select text you just pasted below a heading or title. It mostly surfaced in comment composers, where that selection shape is easy to produce.

While expanding a selection out to its line/block boundaries, an endpoint flush against a block edge was attributed to that edge block even though it contributed no actual content. The expansion then wrapped the empty edge block along with the real selection. Now each endpoint that sits against a block edge is first pulled into the block that actually holds the selected content, so only the intended blocks are wrapped.

The boundary-handling this touches was last reshaped in "Rename split helper and extract sibling helper".

Fixes [Basecamp card #10018915941](https://app.basecamp.com/2914079/buckets/41746046/card_tables/cards/10018915941): Code block grabs more than it should

### DIFF
--- a/src/helpers/lexical_helper.js
+++ b/src/helpers/lexical_helper.js
@@ -227,6 +227,7 @@ export function $splitSelectedParagraphsAtInnerLineBreaks(selection) {
 
 export function $expandSelectionToLineBreaksAndSplitAtEdges(selection, fallbackAncestor = (node) => node.getTopLevelElement()) {
   $ensureForwardRangeSelection(selection)
+  $shrinkSelectionPastBlockEdges(selection)
 
   const focusCaret = $caretFromPoint(selection.focus, "next")
   const anchorCaret = $caretFromPoint(selection.anchor, "previous")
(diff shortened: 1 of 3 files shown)


jorgemanrubia:

Tested the exact reproduction from the card (#10018915941) with a real keyboard/selection in both environments using a Selenium-driven Chromium.

**Setup (both environments):** Created two paragraphs in the editor — `ABCD:` followed by `some pasted text`. Then built a real DOM selection whose anchor sits flush at the **end** of the `ABCD:` paragraph (anchorNode = the `<p>`, offset = end of children) and whose focus sits at the end of `some pasted text` — the selection shape the browser produces when you select text pasted below a title line. Finally clicked the toolbar Code button (`insertCodeBlock`).

**Production** (https://basecamp.github.io/lexxy/try-it.html — released CDN build): the `ABCD:` line was pulled into the code block.
- Result: `<pre data-language="plain" data-highlight-language="plain">ABCD:<br>some pasted text</pre>`
- Bug reproduces.

**Local** (this branch, `fix-code-block-grabs-adjacent`): only the selected paragraph became a code block; `ABCD:` stayed a normal paragraph above.
- Result: `<p>ABCD:</p><pre data-language="plain" data-highlight-language="plain">some pasted text</pre>`
- Bug fixed.

**Regression check (list toggle off):** Also verified the case this PR previously regressed. Created a 3-item bullet list (`one`/`two`/`three`), selected across all three items, and toggled the bullet list OFF. All three items converted to paragraphs (`<p>one</p><p>two</p><p>three</p>`) — not just the middle one. PASS.

Verdict: Production reproduces, local fixed.

PR #1155, opened 2026-06-23
Preserve blank lines between paragraphs in pasted blockquotes

jorgemanrubia:
## Summary

- Pasting a multi-paragraph Markdown blockquote (paragraphs separated by a blank `>` line) collapsed the gap between the quoted paragraphs, even though the same blank line between top-level paragraphs rendered with a visible gap.
- Root cause: `addBlockSpacing` (used by surfaces like BC5 via the `lexxy:insert-markdown` event) inserts spacer paragraphs between adjacent blocks, but its selector only matched direct children of `<body>`. The two `<p>` that `marked` emits inside a `<blockquote>` are nested, so they never received a spacer.
- Fix: extend the selector to also match adjacent block children inside a `<blockquote>`, mirroring the existing heading exclusion. The editor already preserves multiple paragraphs inside a blockquote; this restores the spacing host apps rely on.
- Tests: added a Playwright test asserting the pasted blockquote keeps two separate paragraphs, a Playwright test asserting `addBlockSpacing` inserts the spacer inside the blockquote, and a unit test for `addBlockSpacing`.

Fixes [Basecamp card #9935969865](https://app.basecamp.com/2914079/buckets/41746046/card_tables/cards/9935969865): Markdown blockquotes collapse blank lines between paragraphs

### DIFF
--- a/src/helpers/html_helper.js
+++ b/src/helpers/html_helper.js
@@ -45,8 +45,8 @@ export function dispatch(element, eventName, detail = null, cancelable = false) 
 }
 
 export function addBlockSpacing(doc) {
-  const blocks = doc.querySelectorAll("body > :not(h1, h2, h3, h4, h5, h6) + *")
-  for (const block of blocks) {
+  const selector = "body > :not(h1, h2, h3, h4, h5, h6) + *, blockquote > :not(h1, h2, h3, h4, h5, h6) + *"
+  for (const block of doc.querySelectorAll(selector)) {
     const spacer = doc.createElement("p")
     spacer.appendChild(doc.createElement("br"))
     block.before(spacer)
--- a/test/browser/tests/paste/markdown.test.js
+++ b/test/browser/tests/paste/markdown.test.js
@@ -36,4 +36,52 @@ test.describe("Paste — Markdown", () => {
     await editor.paste("Hello **there**")
     await assertEditorHtml(editor, "<p>Hello **there**</p>")
   })
+
+  test("keep blank lines between paragraphs in a blockquote as separate paragraphs", async ({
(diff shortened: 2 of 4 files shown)


jorgemanrubia:

Validated against production (released CDN build via the try-it page) and the local fix branch using a real browser (Selenium), driving the editor through real DOM events.

**Setup (identical on both environments):** Before pasting, I attached the host-app opt-in listener that BC5 uses, so the comparison is fair:

```js
document.querySelector("lexxy-editor").addEventListener("lexxy:insert-markdown", (event) => {
  event.detail.addBlockSpacing()
})
```

Then I dispatched a real `paste` ClipboardEvent with `text/plain`:

```
> First paragraph of the quote.
>
> Second paragraph of the quote.
```

**Production** (https://basecamp.github.io/lexxy/try-it.html) — bug reproduces. The gap between the two quoted paragraphs is collapsed; no spacer paragraph is inserted:

```html
<blockquote><p>First paragraph of the quote.</p><p>Second paragraph of the quote.</p></blockquote>
```

**Local** (fix branch) — bug fixed. A paragraph-gap spacer is now inserted between the two paragraphs inside the blockquote:

```html
<blockquote><p>First paragraph of the quote.</p><p><br></p><p>Second paragraph of the quote.</p></blockquote>
```

**Important behavioral note:** This is gated on the host calling `event.detail.addBlockSpacing()`. The default paste path exposes that callback via the `lexxy:insert-markdown` event but does NOT call it automatically — without the listener, both production and local produce the no-spacer form. With the listener attached, production collapses the gap and the fix restores it. Verdict: Validated.

PR #1156, opened 2026-06-23
Keep nested bullets when pasting lists

jorgemanrubia:
Pasting text that contains nested (indented) bullet lists — for example copied out of Gmail — silently dropped the indented items. Only the top-level bullets survived; everything nested under them disappeared. Several customers reported it.

The regression came in with [Strip stray non-<li> children from pasted lists](https://github.com/basecamp/lexxy/pull/1119). That change removed any non-`<li>` direct child of a pasted list to fix a stray-`<br>` problem, on the assumption that `<li>` is the only valid direct child of a list. But some sources nest a sublist as a *direct child* of the parent list rather than inside the preceding `<li>`, so the cleanup threw the nested list away before it could be imported.

Now, before that stripping runs, a nested `<ol>`/`<ul>` that sits directly inside its parent list is moved into the `<li>` that precedes it. The nesting is preserved on paste and the stray-child cleanup still does its job for genuine junk.

## Summary

- Move a nested list found as a direct child of its parent list into the preceding `<li>` during paste formatting, instead of discarding it.
- Add Playwright coverage for nested `<ul>` and `<ol>` pasted with the sublist as a direct child.

Fixes [Basecamp card #10010840203](https://app.basecamp.com/2914079/buckets/41746046/card_tables/cards/10010840203): [BC5] Copying and pasting text that includes indented bullets ignores the indented content

### DIFF
--- a/src/editor/contents/pasted_content_formatter.js
+++ b/src/editor/contents/pasted_content_formatter.js
@@ -6,6 +6,7 @@ export default class PastedContentFormatter {
   format() {
     this.#unwrapPlaceholderAnchors()
     this.#stripTableCellColorStyles()
+    this.#nestStrayListChildren()
     this.#stripStrayListChildren()
     return this.doc
   }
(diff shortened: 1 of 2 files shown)


jorgemanrubia:

Reproduced the bug against the released CDN build (try-it page) and confirmed the fix on this branch locally, pasting Gmail-style list HTML where the sublist sits as a *direct child* of the parent list rather than inside an `<li>`:

```html
<ul><li>First item</li><ul><li>Nested item</li></ul><li>Second item</li></ul>
```

Steps: opened the editor, emptied it, focused the content, and dispatched a real `paste` ClipboardEvent carrying that `text/html` (same paste pipeline a Cmd+V from Gmail hits).

**Production** (https://basecamp.github.io/lexxy/try-it.html): the nested bullet was silently dropped — result was `<ul><li>First item</li><li>Second item</li></ul>`. Bug reproduces.

**Local** (fix branch): the nested bullet was kept and correctly moved into the preceding `<li>` — result was `<ul><li value="1">First item<ul><li value="1">Nested item</li></ul></li><li value="2">Second item</li></ul>`, matching the PR's expected output. Bug fixed.

PR #1157, opened 2026-06-23
Preserve code block whitespace and highlight offsets when rendering

jorgemanrubia:
## Summary

- Code blocks with leading whitespace (indented lines, ASCII/box-drawing tables) shifted after the editor → save → render round-trip, and color/background highlights inside code blocks drifted onto the wrong text.
- Root cause: `highlightElement` rebuilt the Prism source via `DOMParser().parseFromString(...).body.textContent`, which applies HTML whitespace normalization and strips the block's leading whitespace. Highlight `<mark>` offsets are computed from the un-normalized DOM, so applying them to the shorter, collapsed Prism text landed them on the wrong characters.
- Fix: build the Prism source by walking the same nodes `extractHighlightRanges` counts (text nodes verbatim, a newline per `<br>`), so both walks share an identical character stream. Whitespace survives and highlight ranges stay aligned.
- Added regression tests covering leading whitespace, indented ASCII tables, and highlight placement on indented code.

Fixes [Basecamp card #10018912693](https://app.basecamp.com/2914079/buckets/41746046/card_tables/cards/10018912693): Code block randomly shifts contents around

### DIFF
--- a/src/helpers/code_highlighting_helper.js
+++ b/src/helpers/code_highlighting_helper.js
@@ -12,7 +12,6 @@ export function highlightElement(preElement) {
   if (preElement.dataset.highlighted === "true") return
 
   const language = preElement.getAttribute("data-language")
-  let code = preElement.innerHTML.replace(/<br\s*\/?>/gi, "\n")
 
   const grammar = Prism.languages?.[language]
   if (!grammar) return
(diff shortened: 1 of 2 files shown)


jorgemanrubia:

Validated the rendered-view round-trip (editor → save → render) for an indented code block, comparing the released CDN build (production try-it page) against this branch (local dummy app).

**Reproduction:** Rendered a code block whose **first line is indented** — a box-drawing ASCII table — with a background highlight on the `ab` cell, then ran the same `highlightCode()` render path Action Text uses on the show page. The bug is triggered specifically by leading whitespace at the *start* of the block: `DOMParser().parseFromString(...).body.textContent` collapses it.

**Production (released build, `@37signals/lexxy@latest`):** bug reproduces. The table's top border `  +----+----+` lost its leading spaces and shifted left, misaligning it from the data rows, and the yellow highlight drifted off `ab` onto the adjacent `|`.

**Local (this branch):** fixed. The first line keeps its leading spaces so the table stays aligned, and the highlight stays on `ab`.

Also ran a full real human round-trip on local (typed an indented ASCII table into the editor, created a code block, saved the post, opened the show page): the rendered output preserves the first line's leading whitespace exactly as stored — confirming `extractCode` keeps the character stream aligned end to end.

PR #1158, opened 2026-06-23
Keep text typed above a quote unformatted

jorgemanrubia:
## Summary

- Pressing Enter at the very start of a blockquote's first paragraph created a new empty paragraph *inside* the blockquote. After navigating up to that line and typing, the text became quoted — so any text added above formatted content at the top of a message inherited the formatting instead of staying plain.
- Now, when the caret is collapsed at offset 0 of the blockquote's first (non-blank) paragraph, we insert a plain paragraph **before** the blockquote and leave the quoted content untouched. This matches the BC4/Trix behavior where text typed above a quote is unformatted.
- The fix lives in the existing `INSERT_PARAGRAPH_COMMAND` handler in `src/extensions/format_escape_extension.js`, split into `$escapeBeforeBlockquoteStart` (new) and `$escapeFromBlankBlockquoteParagraph` (the prior logic).

## Tests

- New Playwright coverage in `test/browser/tests/formatting/format_bleed_above.test.js`:
  - Enter at the start of a blockquote then typing above stays plain (`<p>` before the blockquote).
  - Regression guards confirming inline **bold** and *italic* formatting does not bleed onto a line typed above it.
- Full Playwright (chromium) and Vitest suites pass; lint clean.

Fixes [Basecamp card #9985732711](https://app.basecamp.com/2914079/buckets/41746046/card_tables/cards/9985732711): Formatting being applied to text incorrectly if added before formatted text

### DIFF
--- a/src/extensions/format_escape_extension.js
+++ b/src/extensions/format_escape_extension.js
@@ -55,13 +55,15 @@ function $escapeBeforeBlockquoteStart() {
   if (!$isRangeSelection(selection) || !selection.isCollapsed() || selection.anchor.offset !== 0) return false
 
   const paragraph = $getNearestNodeOfType(selection.anchor.getNode(), ParagraphNode)
-  if (!paragraph || $isBlankNode(paragraph) || paragraph.getPreviousSibling()) return false
+  if (paragraph && !$isBlankNode(paragraph) && !paragraph.getPreviousSibling()) {
+    const blockquote = paragraph.getParent()
+    if ($isQuoteNode(blockquote)) {
+      blockquote.insertBefore($createParagraphNode())
+      return true
+    }
+  }
 
-  const blockquote = paragraph.getParent()
-  if (!blockquote || !$isQuoteNode(blockquote)) return false
-
-  blockquote.insertBefore($createParagraphNode())
-  return true
+  return false
 }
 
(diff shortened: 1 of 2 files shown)


jorgemanrubia:

Tested like a human with a Selenium-driven Chrome against both the released CDN build (production try-it page) and this branch running locally, using identical keyboard interactions in both.

**Steps (identical in both environments):**
1. Seeded a clean blockquote: `<blockquote><p>Quoted</p></blockquote>`.
2. Clicked into the editor, pressed `Home` to put the caret at offset 0 of the quoted paragraph.
3. Pressed `Enter`, then `ArrowUp` to land on the newly created line above.
4. Typed `plain text`.

**Production** (released build): typing above produced `<blockquote><p>plain text</p><p>Quoted</p></blockquote>` — the new line was created *inside* the blockquote, so "plain text" became quoted. Bug reproduces.

**Local** (this branch): typing above produced `<p>plain text</p><blockquote><p>Quoted</p></blockquote>` — a plain paragraph is inserted *before* the blockquote and the typed text stays unformatted. Fixed.

The rendered editor confirms the difference visually: on production "plain text" appears italic and inside the quote bar; locally it sits as a normal paragraph above the quote. Matches the intended BC4/Trix behavior.

PR #1159, opened 2026-06-23
Don't bleed quote formatting into text typed above a blockquote

jorgemanrubia:
## Summary

- Pressing **Enter at the very start of a blockquote's first line** split the paragraph in place, leaving a new empty paragraph *inside* the blockquote. Text typed there inherited the quote formatting, so content added "above" the quote came out formatted instead of plain — diverging from Trix/BC4 behavior.
- The fix handles this in the `INSERT_PARAGRAPH` command: when the cursor is collapsed at offset 0 of a blockquote's first (non-blank) paragraph, we insert a fresh paragraph *before* the blockquote instead of splitting inside it. The quote is pushed down and the line above becomes a plain, unformatted paragraph.
- The existing "escape from the end/bottom of a blockquote" behavior is preserved (refactored into `$escapeFromBlockquoteEnd`).

## Test

Adds `test/browser/tests/formatting/typing_above_formatted_content.test.js` covering Enter and Shift+Enter at the start of a blockquote, plus typing above bold text. The Enter case fails without the fix (`<blockquote><p>plain text</p><p>Quoted</p></blockquote>`) and passes with it (`<p>plain text</p><blockquote><p>Quoted</p></blockquote>`). Full Playwright suite: 544 passed, 0 failures (1 pre-existing flaky).

Fixes [Basecamp card #9985732711](https://app.basecamp.com/2914079/buckets/41746046/card_tables/cards/9985732711): Formatting being applied to text incorrectly if added before formatted text

### DIFF
--- a/src/extensions/format_escape_extension.js
+++ b/src/extensions/format_escape_extension.js
@@ -47,6 +47,25 @@ export class FormatEscapeExtension extends LexxyExtension {
 }
 
 function $escapeFromBlockquote() {
+  return $escapeFromBlockquoteStart() || $escapeFromBlockquoteEnd()
+}
+
+function $escapeFromBlockquoteStart() {
+  const selection = $getSelection()
+  if (!$isRangeSelection(selection) || !selection.isCollapsed() || selection.anchor.offset !== 0) return false
+
+  const paragraph = $getNearestNodeOfType(selection.anchor.getNode(), ParagraphNode)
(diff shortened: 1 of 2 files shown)


jorgemanrubia:
Closing as a duplicate of #1158, which fixes the same bug (card 9985732711) with the same approach. #1158 has been validated end-to-end. Consolidating on it to avoid two PRs for one card.

PR #1161, opened 2026-06-23
Keep toolbar dropdowns open during non-selection editor updates

jorgemanrubia:
## Summary

- The toolbar registered an editor update listener that closed **all** open dropdowns on **every** editor update.
- While an upload is in progress the editor fires updates for upload progress and placeholder mutations, which closed any open toolbar dropdown unexpectedly.
- The fix closes dropdowns only when the selection actually changed between the previous and current editor state, preserving the legitimate "close on selection change" behavior while leaving dropdowns open for content-only (upload-driven) updates.

## Tests

- Added a Playwright test asserting a dropdown stays open across a content-only editor update (no selection change).
- Added a Playwright test asserting a dropdown still closes when the selection changes.
- Verified the first test fails without the source change and passes with it.

Fixes [Basecamp card #9933920069](https://app.basecamp.com/2914079/buckets/41746046/card_tables/cards/9933920069): Toolbar dropdowns close, while upload is progressing

### DIFF
--- a/src/elements/toolbar.js
+++ b/src/elements/toolbar.js
@@ -134,13 +134,15 @@ export class LexicalToolbarElement extends HTMLElement {
     }
   }
 
-  #dispatchButtonCommand(event, { dataset: { command, payload } }) {
+  #dispatchButtonCommand(event, button) {
+    const { command, payload } = button.dataset
     const isKeyboard = event instanceof PointerEvent && event.pointerId === -1
 
     this.editor.update(() => {
       this.editor.dispatchCommand(command, payload)
     }, { tag: isKeyboard ? SKIP_DOM_SELECTION_TAG : undefined })
 
+    if (button.closest(".lexxy-editor__toolbar-dropdown")) this.closeDropdowns()
     if (!isKeyboard) this.editor.focus()
   }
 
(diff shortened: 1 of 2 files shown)


jorgemanrubia:

Validated by driving a real browser (Selenium) against the released CDN build on the try-it page and against this branch locally.

**Primary (upload-progress / content-only update):** Seeded text, clicked the text-style/heading dropdown trigger so the menu opened, then fired an editor update that mutated text content while leaving the selection untouched (the same kind of update upload progress / placeholder mutations produce). Observed the dropdown panel's `hidden` state via the DOM.
- Production: dropdown closed on the content-only update (bug reproduces).
- Local (this branch): dropdown stayed open across the content-only update (fixed).

**Regression guard (menuitem still closes the dropdown):** Opened the text-style dropdown and clicked the "Large Heading" menuitem. In both production and local the dropdown closed after activating the format command, and the heading was applied. The PR's `#dispatchButtonCommand` change correctly keeps this close-on-activate behavior even though the selection-change gate no longer triggers it. No "stuck open" menu.

Verdict: production reproduces (dropdown closes on a non-selection update), local fixed (stays open), and the menuitem-close regression guard passes.

PR #1162, opened 2026-06-23
Preserve backslashes in plain text pasted across multiple lines

jorgemanrubia:
Pasting a Windows/UNC file path that spans more than one line silently dropped the first of its two leading backslashes — `\\server\share` came in as `\server\share`. It happened when the path was copied from somewhere it wrapped across lines, such as multi-line CLI output, or when pasting with "paste and match style". A single-line path pasted cleanly; only the multi-line case was affected.

Originally reported in [Local directory links being modified in Lexxy](https://app.basecamp.com/2914079/buckets/41746046/card_tables/cards/9942235640).

This is a follow-up to [Preserve plain text spacing and backslashes on paste](https://github.com/basecamp/lexxy/commit/52aaa34f568cd42b27120cb3036d7585b41dab1c), which fixed the same corruption for single-line paths. Plain text pasted without an HTML payload routes through the Markdown conversion path, which collapses whitespace and unescapes backslashes. That fix skipped Markdown and pasted verbatim only when the text rendered to a single paragraph of pure text nodes. A path that wrapped across lines renders as a paragraph with `<br>` line breaks (the Markdown converter runs with line breaks enabled), so it failed that check and fell back to the corrupting path.

Multi-line plain text with no Markdown structure is now recognized as plain text and pasted verbatim, with its line breaks preserved.

### DIFF
--- a/src/editor/clipboard.js
+++ b/src/editor/clipboard.js
@@ -182,14 +182,16 @@ export default class Clipboard {
 
   // Markdown conversion collapses runs of whitespace and unescapes backslashes,
   // silently corrupting plain text such as Windows/UNC file paths. When the text
-  // carries no Markdown structure, paste it verbatim instead.
+  // carries no Markdown structure, paste it verbatim instead. A path that wrapped
+  // across lines renders as a single paragraph with <br> line breaks (marked runs
+  // with breaks: true), which is still plain text we should preserve untouched.
   #isPlainTextWithoutMarkdown(doc) {
     const elements = Array.from(doc.body.children)
     if (elements.length !== 1) return false
 
     const paragraph = elements[0]
     return paragraph.nodeName === "P"
-      && Array.from(paragraph.childNodes).every((node) => node.nodeType === Node.TEXT_NODE)
+      && Array.from(paragraph.childNodes).every((node) => node.nodeType === Node.TEXT_NODE || node.nodeName === "BR")
   }
 
   #pasteRichText(clipboardData) {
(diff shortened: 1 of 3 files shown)


jorgemanrubia:

Tested the actual user flow with a real browser (Selenium + Chromium): put a multi-line Windows/UNC path on the OS clipboard as plain text only (`navigator.clipboard.writeText`, no `text/html`), focused the editor, and pasted with a real Ctrl+V. Same script run against production (CDN) and this branch (local).

The path used wraps across two lines, mirroring the reported scenario (path copied from multi-line CLI output / paste-and-match-style):

```
\\Arina-alvand\alvand\03 - ENGINEERING
\00. Input\2026-06-17
```

**Production** (`https://basecamp.github.io/lexxy/try-it.html` — released 0.9.19, has the single-line #1138 fix only):
- Result: `<p>\Arina-alvand\alvand\03 - ENGINEERING<br>\00. Input\2026-06-17</p>`
- The first of the two leading backslashes is dropped — bug reproduces.

**Local** (this branch):
- Result: `<p>\\Arina-alvand\alvand\03 - ENGINEERING<br>\00. Input\2026-06-17</p>`
- Both leading backslashes preserved, line break kept as a single paragraph — fixed.

**Probe:** pasted a multi-line snippet containing inline markdown (`**bold** word\nsecond line`) on local → converted to `<p><strong>bold</strong> word<br>second line</p>`. The broadened plain-text guard still lets genuine markdown convert; it doesn't over-capture formatted content as verbatim text.

PR #1175, opened 2026-07-02
Code blocks load unhighlighted from the initial value

jorgemanrubia:
Code blocks loaded from the editor's initial `value` render without syntax highlighting. The language imports correctly (the toolbar picker even shows it), but the block stays plain until something touches it — e.g. re-selecting a language. Easy to see on the lexxy.dev sandbox, which loads a CSS block as part of the initial content.

## Where it came from

Regression from 30a7a37d ("Inject initial value at editor creation time", Jun 2), first released in **v0.9.16**. That commit moved initial-value parsing from `#initialize()` (which ran *after* `#registerComponents()`) into editor creation via `$initialEditorState`, which `@lexical/extension` applies in its `afterRegistration` phase — before `#registerComponents()` calls `registerCodeHighlighting()`.

Why that ordering matters:

1. Lexical only tokenizes code blocks through a node transform, and transforms only run on nodes marked dirty.
2. When the initial value is parsed, its transform pass runs immediately — but the code-highlighting transform isn't registered yet.
3. `registerCodeHighlighting()` registers next. Lexical's catch-up pass (`markNodesWithTypesAsDirty`) marks existing code nodes dirty so late-registered transforms run — but it reads the **committed** editor state, and the initial-value update is still pending (it commits on a microtask, later in the same task). It sees an empty document and marks nothing.
4. The content commits; transforms don't re-run on commit. The code block stays untokenized until some later edit dirties it — which is exactly what picking a language in the toolbar does.

Before 30a7a37d the registration ran first, so the initial value's own transform pass did the tokenizing. That's why apps pinned to older versions (e.g. Fizzy on 0.9.5.beta) don't show this.

## The fix

Register code highlighting through the extension system instead of after editor creation. Extension `register()` hooks run before `@lexical/extension` applies the initial state, restoring the ordering guarantee the old code had. As a side effect this also stops discarding `registerCodeHighlighting()`'s teardown function — the extension system now owns it.

The `lexxy-code-language-picker` element setup stays in the editor element, where it belongs (it needs the DOM).

Verified on the lexxy.dev sandbox running against this build: the CSS block tokenizes on first paint. The new browser test fails on `main` and passes here.

### DIFF
--- a/src/elements/editor.js
+++ b/src/elements/editor.js
@@ -8,7 +8,7 @@ import { HeadingNode, QuoteNode, registerRichText } from "@lexical/rich-text"
 import { $generateHtmlFromNodes, $generateNodesFromDOM as $generateLexicalNodesFromDOM } from "@lexical/html"
 import { filterDisallowedAttachmentNodes } from "../helpers/attachment_filter_helper"
 import { $convertInlineImageDataURIs } from "../helpers/inline_image_uri_helper"
-import { CodeHighlightNode, CodeNode, registerCodeHighlighting } from "@lexical/code"
+import { CodeHighlightNode, CodeNode } from "@lexical/code"
 import { TRANSFORMERS, registerMarkdownShortcuts } from "@lexical/markdown"
 import { HORIZONTAL_DIVIDER } from "../editor/markdown/horizontal_divider_transformer"
 import { registerMarkdownLeadingTagHandler } from "../editor/markdown/leading_tag_handler"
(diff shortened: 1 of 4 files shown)


jorgemanrubia:
@samuelpecher I noticed this in the new home https://basecamp.github.io/lexxy/sandbox. We don't have this bug in BC5 because we don't have scenarios where we load an editor (only where we restore the editor contents from local storage).

cc @zoltanhosszu

PR #1177, opened 2026-07-02
Guard focus() against an uninitialized editor

MatheusRich:
`focus()` reads `this.editor`, which is only assigned in connectedCallback. When `focus()` runs before the editor is initialized, the `#isContentFocused` guard evaluates to false and execution falls through to `this.editor.focus()`, throwing "undefined is not an object (evaluating 'this.editor.focus')".

This can happen when using Lexxy alongside Turbo and autofocus. Turbo can call `focus()` on the element before the editor is initialized, leading to an error.

This commit makes Lexxy return early when `this.editor` is not set. Autofocus still works through the element's own `#handleAutofocus`, which runs after the editor is created.

### DIFF
--- a/src/elements/editor.js
+++ b/src/elements/editor.js
@@ -317,7 +317,7 @@ export class LexicalEditorElement extends HTMLElement {
     // Skip if the contenteditable already owns focus — the update would be a
     // no-op but still triggers a full style/layout pass on pages with large
     // DOMs.
-    if (this.#isContentFocused) return
+    if (!this.editor || this.#isContentFocused) return
 
     this.editor.focus(() => this.#onFocus())
   }
--- a/src/elements/editor.js
+++ b/src/elements/editor.js
@@ -317,9 +317,9 @@ export class LexicalEditorElement extends HTMLElement {
     // Skip if the contenteditable already owns focus — the update would be a
     // no-op but still triggers a full style/layout pass on pages with large
     // DOMs.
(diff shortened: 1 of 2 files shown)


src/elements/editor.js
```diff
@@ -317,7 +317,7 @@ export class LexicalEditorElement extends HTMLElement {
     // Skip if the contenteditable already owns focus — the update would be a
     // no-op but still triggers a full style/layout pass on pages with large
     // DOMs.
-    if (this.#isContentFocused) return
+    if (!this.editor || this.#isContentFocused) return
 
     this.editor.focus(() => this.#onFocus())
```
samuelpecher: Just a small nit to bring us a step closer to the style guide.

```suggestion
    if (this.editor && !this.#isContentFocused) {
      this.editor.focus(() => this.#onFocus())
    }
```

samuelpecher (commented):
What's the tradeoff in making this a no-op vs storing the focus intent and focusing the editor on eventual load?

MatheusRich:
> What's the tradeoff in making this a no-op vs storing the focus intent and focusing the editor on eventual load?

I think there's no need to do that for my use-case, at least. Lexxy's own `#handleAutofocus` runs in the `requestAnimationFrame` at the end of `connectedCallback`, so once the editor is created it gets focused anyway. Turbo's early `focus()` is just a redundant, premature duplicate of that, so dropping it changes nothing observable.

PR #1179, opened 2026-07-03
Fix prompt insertion for multi-character triggers like {{

### LINKED ISSUE #1165: insert-editable-text: trigger with repeated character (e.g. `{{`) fails to insert in 0.9.19
## Summary

Selecting a prompt item (via Enter, Tab, or click) does nothing when using a trigger whose first character repeats, such as `{{`. This is a regression introduced in 0.9.19 — the same setup works correctly in 0.9.18.

## Steps to reproduce

```html
<lexxy-prompt trigger="{{" insert-editable-text>
  <lexxy-prompt-item search="artist name">
    <template type="menu"><span>Artist name</span></template>
    <template type="editor">{{artist_name}}</template>
  </lexxy-prompt-item>
</lexxy-prompt>
```

1. Type `{{` in the editor — the prompt menu opens correctly
2. Press Enter (or Tab, or click an item)
3. **Expected:** `{{artist_name}}` is inserted
4. **Actual:** nothing happens; the menu closes but the editor is unchanged

## Root cause

`#findReplacementStart` was introduced in 0.9.19 to replace `#findLastIndexBeforeCursor`. The new method searches for `stringToReplace[0]` (the first character of the trigger) rather than the full trigger string:

```js
// 0.9.19 — broken for repeated-character triggers
#findReplacementStart(anchorNode, offset, stringToReplace) {
  const fullText = anchorNode.getTextContent();
  const triggerIndex = fullText.slice(0, offset).lastIndexOf(stringToReplace[0]); // ← searches for "{", not "{{"

  if (triggerIndex !== -1 && fullText.startsWith(stringToReplace, triggerIndex)) {
    return triggerIndex
  } else {
    return -1
  }
}
```

For trigger `{{`, `stringToReplace` is `"{{"`. `lastIndexOf("{")` finds the **second** `{` at index 1, not the first at index 0. Then `fullText.startsWith("{{", 1)` is false, so the method returns `-1`. `replaceTextBackUntil` bails out and nothing is inserted.

The fix in 0.9.18 was straightforward:

```js
// 0.9.18 — correct
#findLastIndexBeforeCursor(anchorNode, offset, stringToReplace) {
  const fullText = anchorNode.getTextContent();
  const textBeforeCursor = fullText.slice(0, offset);
  return textBeforeCursor.lastIndexOf(stringToReplace) // ← searches for the full "{{"
}
```

## Suggested fix

Change `lastIndexOf(stringToReplace[0])` to `lastIndexOf(stringToReplace)` in `#findReplacementStart`, and update the `startsWith` check accordingly (or drop it, since `lastIndexOf` of the full string is already exact):

```js
#findReplacementStart(anchorNode, offset, stringToReplace) {
  const fullText = anchorNode.getTextContent();
  const triggerIndex = fullText.slice(0, offset).lastIndexOf(stringToReplace); // full string
  return triggerIndex
}
```

## Environment

- lexxy 0.9.19 (regression vs 0.9.18)
- Rails 8, esbuild
- Chrome / Firefox (both affected)

anquinn:
If you used a multi-character trigger like `{{`, picking an item from the prompt menu did nothing. The typed text stayed put and nothing got inserted.

The problem is in how `#findReplacementStart` locates the text to replace. It anchors on the last occurrence of the string's first character before the cursor, then checks that the whole string matches from there. With `{{foo` the last `{` is the second brace, so that check fails and `replaceTextBackUntil` bails out. Single-character triggers never ran into this because the last occurrence of the trigger is always the start of the match.

This looks like an unintended side effect of 9d23f39 (#1135), which introduced #findReplacementStart to let the replaced span straddle the cursor. That change switched the anchor from `lastIndexOf(stringToReplace)` over the pre-cursor text to `lastIndexOf(stringToReplace[0])` plus a `startsWith` check. Anchoring on just the first character is what leaves multi-character triggers stranded on the wrong brace.

The fix is to search for the full string with `lastIndexOf`, using its `fromIndex` argument to bound where the match starts rather than slicing the text before the cursor. A match must start before the cursor but can extend past it, so single-character triggers work the same as before and strings that straddle the cursor (like an `@` typed just in front of an existing word) are still found.

This fixes #1165. The issue suggests reverting to the 0.9.18 approach, which used `lastIndexOf(stringToReplace)` over the text before the cursor. That approach doesn't seem to cover the straddling case, which is the reason 9d23f39 moved away from it. When `@` is typed just in front of an existing word, stringToReplace is `@Jack` while the cursor sits right after the `@`, so searching only the pre-cursor text would miss it — bounding the match's start with fromIndex instead of slicing handles both.

### DIFF
--- a/src/editor/contents.js
+++ b/src/editor/contents.js
@@ -579,18 +579,13 @@ export default class Contents {
     return { anchorNode, offset: anchor.offset }
   }
 
-  // The replaced span can straddle the cursor (e.g. "@Jack" when "@" was just
-  // inserted before "Jack"), so we anchor on the trigger before the cursor and
-  // verify the whole string matches there rather than searching text up to it.
+  // The replaced span must start before the cursor but can extend past it
+  // (e.g. "@Jack" when "@" was just inserted before "Jack"), so we bound the
+  // match's start with lastIndexOf's fromIndex rather than slicing the text.
   #findReplacementStart(anchorNode, offset, stringToReplace) {
-    const fullText = anchorNode.getTextContent()
-    const triggerIndex = fullText.slice(0, offset).lastIndexOf(stringToReplace[0])
+    if (offset === 0) return -1 // a negative fromIndex clamps to 0 and could match at the cursor
 
-    if (triggerIndex !== -1 && fullText.startsWith(stringToReplace, triggerIndex)) {
-      return triggerIndex
-    } else {
-      return -1
-    }
+    return anchorNode.getTextContent().lastIndexOf(stringToReplace, offset - 1)
   }
 
   #performTextReplacement(anchorNode, startIndex, stringToReplace, replacementNodes) {
--- a/src/editor/contents.js
+++ b/src/editor/contents.js
@@ -583,9 +583,12 @@ export default class Contents {
   // (e.g. "@Jack" when "@" was just inserted before "Jack"), so we bound the
   // match's start with lastIndexOf's fromIndex rather than slicing the text.
   #findReplacementStart(anchorNode, offset, stringToReplace) {
-    if (offset === 0) return -1 // a negative fromIndex clamps to 0 and could match at the cursor
-
-    return anchorNode.getTextContent().lastIndexOf(stringToReplace, offset - 1)
+    if (offset === 0) {
+      // A negative fromIndex clamps to 0 and could match at the cursor
+      return -1
+    } else {
+      return anchorNode.getTextContent().lastIndexOf(stringToReplace, offset - 1)
+    }
   }
 
   #performTextReplacement(anchorNode, startIndex, stringToReplace, replacementNodes) {
--- a/test/javascript/unit/editor/replace_text_back_until.test.js
+++ b/test/javascript/unit/editor/replace_text_back_until.test.js
@@ -1,0 +1,53 @@
+import { describe, expect, test } from "vitest"
+import { $createTextNode } from "lexical"
+import { createTestEditor, destroyTestEditor, selectFirstText, setContent, tick } from "../helpers/editor_helper"
+
+function replaceTextBackUntil(editorElement, stringToReplace, replacementText) {
+  editorElement.editor.update(() => {
+    editorElement.contents.replaceTextBackUntil(stringToReplace, $createTextNode(replacementText))
+  }, { discrete: true })
+}
+
+describe("Contents#replaceTextBackUntil", () => {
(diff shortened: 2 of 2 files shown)


src/editor/contents.js
```diff
@@ -584,13 +584,13 @@ export default class Contents {
...
 
-    if (triggerIndex !== -1 && fullText.startsWith(stringToReplace, triggerIndex)) {
-      return triggerIndex
-    } else {
-      return -1
+    while (triggerIndex !== -1 && !fullText.startsWith(stringToReplace, triggerIndex)) {
+      triggerIndex = fullText.slice(0, triggerIndex).lastIndexOf(stringToReplace[0])
     }
+
+    return triggerIndex
```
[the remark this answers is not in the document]
anquinn: O(n²) is a bit dramatic given the size of the nodes, but the idea isn't bad. Updated the PR.

excid3:
@jorgemanrubia if you have a second, this seems like an accidental regression for multi-char prompt triggers. I didn't realize they were supported until @anquinn mentioned it. This will come in super handy for fancier prompts like snippets and variables.

src/editor/contents.js
```diff
@@ -579,18 +579,13 @@ export default class Contents {
...
-  // The replaced span can straddle the cursor (e.g. "@Jack" when "@" was just
-  // inserted before "Jack"), so we anchor on the trigger before the cursor and
-  // verify the whole string matches there rather than searching text up to it.
+  // The replaced span must start before the cursor but can extend past it
+  // (e.g. "@Jack" when "@" was just inserted before "Jack"), so we bound the
+  // match's start with lastIndexOf's fromIndex rather than slicing the text.
   #findReplacementStart(anchorNode, offset, stringToReplace) {
-    const fullText = anchorNode.getTextContent()
-    const triggerIndex = fullText.slice(0, offset).lastIndexOf(stringToReplace[0])
+    if (offset === 0) return -1 // a negative fromIndex clamps to 0 and could match at the cursor
```
jorgemanrubia: In-house style: could you please use a regular if/else here instead of an early return?

jorgemanrubia (commented):
Thanks for the patch @anquinn. Minor suggestion and this is good to go

src/editor/contents.js
anquinn: No problem, I'll get this done today.

anquinn:
I've updated it to use if/else instead of an early return. Please let me know if this wasn't what you were looking for @jorgemanrubia

PR #1198, opened 2026-07-14
Drop dispatchToolbarCommand for the editor handle in tests

jorgemanrubia:
## What

Removes the `dispatchToolbarCommand` unit-test helper and makes the editor handle's `clickToolbarButton` the only way toolbar commands are triggered in tests.

## Why

`dispatchToolbarCommand` dispatched Lexical commands programmatically from jsdom (`editor.dispatchCommand(...)` inside an `update()` block), bypassing the real toolbar buttons. That is exactly the kind of coverage the browser suite already provides by driving the actual toolbar through the editor handle. Keeping a parallel jsdom path that fakes the toolbar is redundant and can pass even when the real button wiring is broken.

## How

- Delete `dispatchToolbarCommand` from `test/javascript/unit/helpers/editor_helper.js`.
- Move the two `toggle_code_block` cases (paragraph+quote selection, caret inside a nested list) from the jsdom unit test to a new end-to-end browser test at `test/browser/tests/formatting/toggle_code_block.test.js`, driving `insertCodeBlock` through `editor.clickToolbarButton` and asserting the resulting HTML.

## Verification

- `yarn test` (vitest): 118 passing, no remaining `dispatchToolbarCommand` references.
- `yarn lint`: clean.
- `yarn test:browser:chromium` formatting suite: 205 passing, including the two migrated cases.


test/browser/tests/formatting/toggle_code_block.test.js
```diff
@@ -0,0 +1,49 @@
+import { test } from "../../test_helper.js"
+import { assertEditorHtml } from "../../helpers/assertions.js"
+
+async function placeCaretInside(editor, text, offset) {
```
jorgemanrubia, line 4: Moving this to the handle too separately

jorgemanrubia: 🤖 Done in #1199 — moved `placeCaretInside` onto the editor handle so caret placement goes through the handle like `clickToolbarButton`.

PR #1213, opened 2026-07-29
macOS text replacements lose their line breaks on submit

jorgemanrubia:
A macOS "Keyboard > Text Replacements" snippet containing a newline — a multi-line signature like `Thanks,\nKyle` — looks right while composing but comes back on a single line once the message is submitted.

macOS delivers text replacements as a `beforeinput` event with `inputType: "insertReplacementText"`, which Lexical forwards to `CONTROLLED_TEXT_INSERTION_COMMAND`. Both `registerRichText` and `registerPlainText` handle that by calling `selection.insertText(data)`, which inserts the string verbatim — so the newline stays a literal `\n` inside a single text node instead of becoming a line break node.

That reads as correct in the editor, because Lexical sets `white-space: pre-wrap` on the contenteditable root. But the newline is plain whitespace once the content is serialized to HTML, and collapses to a space when the content is rendered. We lose it where vanilla Lexical wouldn't, because `exportTextNodeDOM` unwraps the `<span>` that would otherwise have carried Lexical's own `white-space: pre-wrap` into the exported markup.

So we now intercept `CONTROLLED_TEXT_INSERTION_COMMAND` when the incoming text contains a newline and insert it with `insertRawText`, which splits it into real line break nodes — the same nodes Shift+Enter produces, serializing to `<br>`. Single-line replacements keep taking the existing path, and events carrying a `dataTransfer` (drop, yank) are left to the default handler.

The regression test dispatches the command directly rather than firing `beforeinput`: jsdom's `InputEvent` has no `getTargetRanges`, so Lexical's own `beforeinput` handling is inert there. Verified end-to-end in Chromium too, with a real `beforeinput` event — before the fix the editor showed the break while the serialized value was `<p>Hello Thanks,\nKyle</p>`; after it, `<p>Hello Thanks,<br>Kyle</p>`.

---

Card: https://app.basecamp.com/2914079/buckets/27/card_tables/cards/10140982528

### DIFF
--- a/src/editor/contents.js
+++ b/src/editor/contents.js
@@ -60,6 +60,19 @@ export default class Contents {
     }, { tag })
   }
 
+  // A literal "\n" stays visible in the editor, which Lexical renders with white-space: pre-wrap,
+  // but collapses to a space once serialized to HTML, so it has to become line break nodes.
+  insertTextWithLineBreaks(text) {
+    const normalizedText = text?.replace(/\r\n?/g, "\n")
+    if (!normalizedText?.includes("\n")) return false
+
+    const selection = $getSelection()
+    if (!$isRangeSelection(selection)) return false
+
+    selection.insertRawText(normalizedText)
+    return true
+  }
+
   insertAtCursor(...nodes) {
     const selection = this.#insertableSelection()
     const inserter = NodeInserter.for(selection)
--- a/src/editor/contents.js
+++ b/src/editor/contents.js
@@ -69,9 +69,9 @@ export default class Contents {
     if (normalizedText?.includes("\n") && $isRangeSelection(selection)) {
       selection.insertRawText(normalizedText)
       return true
+    } else {
+      return false
     }
-
-    return false
   }
 
   insertAtCursor(...nodes) {
--- a/src/editor/contents.js
+++ b/src/editor/contents.js
@@ -64,13 +64,14 @@ export default class Contents {
   // but collapses to a space once serialized to HTML, so it has to become line break nodes.
   insertTextWithLineBreaks(text) {
     const normalizedText = text?.replace(/\r\n?/g, "\n")
-    if (!normalizedText?.includes("\n")) return false
-
     const selection = $getSelection()
-    if (!$isRangeSelection(selection)) return false
 
-    selection.insertRawText(normalizedText)
-    return true
+    if (normalizedText?.includes("\n") && $isRangeSelection(selection)) {
+      selection.insertRawText(normalizedText)
+      return true
+    }
+
+    return false
   }
 
   insertAtCursor(...nodes) {
--- a/src/elements/editor.js
+++ b/src/elements/editor.js
@@ -661,7 +661,7 @@ export class LexicalEditorElement extends HTMLElement {
   #handleReplacementText() {
     this.#listeners.track(this.editor.registerCommand(
       CONTROLLED_TEXT_INSERTION_COMMAND,
-      (eventOrText) => $insertReplacementTextWithLineBreaks(eventOrText),
+      (event) => event instanceof InputEvent && this.contents.insertTextWithLineBreaks(event.data),
       COMMAND_PRIORITY_NORMAL
     ))
   }
@@ -919,28 +919,6 @@ export class LexicalEditorElement extends HTMLElement {
 }
 
 export default LexicalEditorElement
-
-// macOS text replacements and autocorrect arrive as CONTROLLED_TEXT_INSERTION_COMMAND, which
-// inserts their text verbatim. A newline would stay a literal "\n" inside a single text node:
-// visible in the editor, which Lexical renders with "white-space: pre-wrap", but collapsed to a
-// space once serialized to HTML. Turning it into line break nodes keeps the break on submit.
-function $insertReplacementTextWithLineBreaks(eventOrText) {
-  const text = replacementTextFrom(eventOrText)?.replace(/\r\n?/g, "\n")
-  if (!text?.includes("\n")) return false
-
-  const selection = $getSelection()
-  if (!$isRangeSelection(selection)) return false
-
-  selection.insertRawText(text)
-  return true
-}
-
-function replacementTextFrom(eventOrText) {
-  if (typeof eventOrText === "string") return eventOrText
-  if (eventOrText.dataTransfer) return null
-
-  return eventOrText.data
-}
 
 // Like $getRoot().getTextContent() but uses readable text for custom attachment nodes
 // (e.g., mentions) instead of their single-character cursor placeholder.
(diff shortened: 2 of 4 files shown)


src/elements/editor.js
```diff
@@ -657,6 +658,14 @@ export class LexicalEditorElement extends HTMLElement {
     ))
   }
 
+  #handleReplacementText() {
+    this.#listeners.track(this.editor.registerCommand(
+      CONTROLLED_TEXT_INSERTION_COMMAND,
+      (eventOrText) => $insertReplacementTextWithLineBreaks(eventOrText),
+      COMMAND_PRIORITY_NORMAL
+    ))
+  }
```
jorgemanrubia, line 667: Instead of a function, move this to a public method in the `Contents` class so that you delegate the handling to `.contents`.

src/elements/editor.js
```diff
@@ -911,6 +920,28 @@ export class LexicalEditorElement extends HTMLElement {
 
 export default LexicalEditorElement
 
+// macOS text replacements and autocorrect arrive as CONTROLLED_TEXT_INSERTION_COMMAND, which
+// inserts their text verbatim. A newline would stay a literal "\n" inside a single text node:
+// visible in the editor, which Lexical renders with "white-space: pre-wrap", but collapsed to a
+// space once serialized to HTML. Turning it into line break nodes keeps the break on submit.
+function $insertReplacementTextWithLineBreaks(eventOrText) {
```
jorgemanrubia: Do we really need to handle `eventOrText` here? In any case, this method should only receive the text. The caller should extract the text from the event as needed.

src/elements/editor.js
jorgemanrubia: 🤖 Moved to `Contents#insertTextWithLineBreaks(text)` (e8f75d2f), so the handler is now just:

```js
(event) => event instanceof InputEvent && this.contents.insertTextWithLineBreaks(event.data)
```

Followed the `applyParagraphFormat` / `toggleBlockquote` idiom rather than the `insertText` one: like those, it runs from inside a `registerCommand` handler, which is already in an editor update, so it operates on `$getSelection()` directly instead of opening its own `this.editor.update()`.

Also trimmed the 4-line comment block down to one line on the constraint (literal `\n` survives under `white-space: pre-wrap` but collapses when serialized) — you did not ask for that, but narrating comments are on your strip list, so I took the liberty.

src/elements/editor.js
jorgemanrubia: 🤖 No, we do not need it — the union is gone. It takes text only now, and the handler extracts it.

To answer the question properly, though: the payload genuinely *is* `InputEvent | string`. Lexical dispatches both (`node_modules/lexical/Lexical.dev.mjs`):

- **event** — `$handleBeforeInput`, `case "insertFromYank" | "insertFromDrop" | "insertReplacementText"` (:2569) and `case "insertFromComposition"` (:2576)
- **string** — `insertText`/`insertTranspose` typing (:2552), `$handleInput` (:2737), and `COMPOSITION_START_CHAR`, a zero-width space, from `$handleCompositionStart` (:2791)

So both forms reach us, but only the event form can ever be a replacement. The string dispatches are all plain typing or composition, and an `insertText` of `"\n"` never reaches this command at all — Lexical intercepts it upstream and routes it to `INSERT_LINE_BREAK_COMMAND` (:2539). A string payload therefore cannot carry the newline this fix is about, so it is filtered out rather than handled.

The `dataTransfer` guard went away too, for a related reason: `insertFromDrop` / `insertFromYank` put their text in `dataTransfer` and leave `data` null, so they now fall through on `null` instead of needing their own branch. Behaviour is unchanged.

**Evidence the on-submit break still holds:** the regression test asserts the serialized output, not just the node types — `editorElement.value === "<p>Hello Thanks,<br>Kyle</p>"` plus `not.toContain("\n")`. I checked it is not vacuous by swapping `insertRawText` for `insertText` in the new method: 3 of 5 tests fail, including that serialization assertion (`<p>Hello Thanks,\nKyle</p>`). Restored, full suite green at 124/124. The test now dispatches a real `InputEvent` with `inputType: "insertReplacementText"` instead of an event-shaped literal, and I added one case covering the string payload falling through to Lexical.

src/editor/contents.js
```diff
@@ -60,6 +60,19 @@ export default class Contents {
...
   }
 
+  // A literal "\n" stays visible in the editor, which Lexical renders with white-space: pre-wrap,
+  // but collapses to a space once serialized to HTML, so it has to become line break nodes.
+  insertTextWithLineBreaks(text) {
+    const normalizedText = text?.replace(/\r\n?/g, "\n")
+    if (!normalizedText?.includes("\n")) return false
+
+    const selection = $getSelection()
+    if (!$isRangeSelection(selection)) return false
```
jorgemanrubia: Replace early return with regular positive condition wrapping the insert action

jorgemanrubia: since return value matters go with an if/else then

jorgemanrubia: 🤖 Done in 1b75cf1f:

```js
insertTextWithLineBreaks(text) {
  const normalizedText = text?.replace(/\r\n?/g, "\n")

  if (normalizedText?.includes("\n") && $isRangeSelection(selection)) {
    selection.insertRawText(normalizedText)
    return true
  }

  return false
}
```

I kept the trailing `return false` explicit rather than letting it fall off the end, because this value is the `registerCommand` handler's return: `true` stops Lexical, falsy lets it insert the text normally. An inversion that returned `undefined` would still work by luck, but returning `true` on the fall-through would silently break every ordinary single-line replacement — so I would rather have it stated than implied.

Verified rather than assumed, by printing the actual return on each path:

| input | returns |
|---|---|
| multi-line replacement | `true` (boolean) |
| single-line replacement | `false` (boolean) |
| `null` data (drop/yank) | `false` (boolean) |
| `undefined` data | `false` (boolean) |
| no range selection | `false` (boolean) |

Added two cases that pin this, and checked they are not vacuous by mutating the method three ways:

- fall-through returning `undefined` → 2 failures (`expected undefined to be false`)
- fall-through returning `true` → 3 failures, including the behavioural one: the single-line replacement stopped being inserted at all (`expected "<p>Hello</p>" to be "<p>Hello there</p>"`)
- `insertRawText` → `insertText` → 3 failures, including the on-submit serialization assertion (`<p>Hello Thanks,\nKyle</p>`)

One correction worth flagging: my first attempt at the no-selection case asserted `false` but passed a `<p>Hello</p>` editor with no explicit selection, and it returned `true` — Lexical still has a range selection there. I fixed the setup with `$setSelection(null)` instead of relaxing the expectation.

Suite is 126/126 (was 124, +2), serialization case still green.

jorgemanrubia: 🤖 Switched to `if`/`else` in ed305134 — the `else` just returns false, no trailing return:

```js
if (normalizedText?.includes("\n") && $isRangeSelection(selection)) {
  selection.insertRawText(normalizedText)
  return true
} else {
  return false
}
```

eslint is fine with it — `no-else-return` is not enabled in `eslint.config.js` and is not part of `js/recommended` (confirmed via `--print-config`: `null`). Suite still 126/126, and the fall-through cases stayed green unchanged.

PR #1214, opened 2026-07-29
Pasted Word and Outlook lists arrive as indented paragraphs instead of lists

jorgemanrubia:
Pasting a bulleted or numbered list from Word or Outlook produced no list at all. Every item arrived as a plain paragraph with the bullet glyph baked into the text — `<p>·&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; First item</p>` — and all nesting was flattened.

Word and Outlook never put `<ul>`/`<ol>`/`<li>` on the clipboard. They simulate lists with flat paragraphs that declare their depth in an inline `mso-list:l0 level2 lfo1`, and carry the marker glyph in a `<span style="mso-list:Ignore">`:

```html
<p class=MsoListParagraph style='margin-left:.5in;text-indent:-.25in;mso-list:l0 level2 lfo1'>
<![if !supportLists]><span style='font-family:Symbol'><span style='mso-list:Ignore'>·<span
style='font:7.0pt "Times New Roman"'>&nbsp;&nbsp; </span></span></span><![endif]>First item</p>
```

`PastedContentFormatter`'s existing list repairs (`#unwrapWrappedListChildren`, `#nestStrayListChildren`, `#stripStrayListChildren`) all start from `doc.querySelectorAll("ol, ul")`, so for Word content none of them ever run. There was no handling of `mso-list` anywhere in the repo.

## What this does

Adds `#rebuildOfficeLists()`, which runs before the existing repairs so the reconstructed lists still get the same hygiene pass. It groups contiguous `mso-list` paragraphs into runs, reads the `level` each one declares, and rebuilds real nested `<ul>`/`<ol>`/`<li>`, dropping the marker span as it goes.

Two details worth calling out:

- **Bullet vs. number** is decided from the marker glyph. Word writes bullets as the raw glyph of the marker font (Symbol's `·`, Wingdings' `§`, Courier New's `o`) and numbers as the rendered counter (`1.`, `a)`, `iv)`). Note that Courier New's level-two bullet is a bare letter `o`, so a trailing delimiter — not "is it alphanumeric" — is what separates a counter from a bullet.
- **Each run is anchored at its own shallowest level.** A selection copied out of the middle of a document has every paragraph declaring `level3` with nothing above it. Taking those levels literally is what produced the absurdly deep indentation in the report; normalising per run is what fixes it.

## Keeping Office paste compatibility intact

`#stripStyleElements` exists because of the Excel colour-bleed bug (Excel ships a `<style>` block whose `td { color: black }` rules cascade onto pasted cells and don't adapt to dark mode). The obvious way to learn whether a Word level is bulleted or numbered is the `@list l0:level1 { mso-level-number-format:bullet }` rule in that same `<style>` block — which would have meant keeping the stylesheet around and regressing that fix.

So this reads nothing from `<style>`. Everything it needs — the `mso-list` declaration and the marker glyph — lives in inline attributes on the paragraphs, which `#stripStyleElements` never touches. The two fixes are independent, and the ordering in `format()` is unchanged.

Verified rather than assumed: the full `table_cell_styles.test.js` suite still passes, including the case that pastes a real Excel clipboard payload complete with its `<style>` block and asserts no cascaded colour survives. The Word fixtures in the new tests also carry a Word `<style>` block with `@list` rules, which proves the rebuild works with the stylesheet present *and* does not depend on it surviving.

## Tests

`test/browser/tests/paste/paste_office_lists.test.js` covers single-level and multi-level bulleted and numbered lists, a bulleted sublist under a numbered parent, a marker type switch at the same level, a selection copied from deep inside a list, runs split by a plain paragraph, paragraphs declaring `mso-list:none`, and the absence of leftover glyphs and `margin-left`/`text-indent`. Ten of the eleven fail on `main` and pass here; the eleventh (`mso-list:none` paragraphs stay paragraphs) passes both ways on purpose, as a guard that the new pass is scoped.

## On the fixture's fidelity

Worth being explicit, since it bounds how much this proves: the Help Scout case contained only a screenshot and the source `.docx`, not a clipboard capture, and there is no Word or Outlook on this machine. So the Word clipboard HTML used to reproduce the bug was **constructed**, not captured — generated from the customer document's own `word/document.xml` and `word/numbering.xml` (level formats, marker glyphs and fonts, indent geometry) into the clipboard shape Word is documented to emit.

That reproduces the reported symptom exactly (65 list paragraphs → 0 lists, marker glyphs inlined as text) and the fix turns it into 65 correctly nested `<li>`s across 6 `<ul>`s and 19 `<ol>`s. But a hand-built fixture is weaker evidence than a captured one: it confirms the shape I modelled is handled, not that it is byte-identical to every Word and Outlook version. A verification paste from a real Windows Word/Outlook would be worth doing before this is considered closed.

The committed fixtures use neutral placeholder content — the customer document is a university's meeting minutes, so only its *structure* informed the tests.

---

Card: https://app.basecamp.com/2914079/buckets/27/card_tables/cards/10126744106

### DIFF
--- a/src/editor/contents/pasted_content_formatter.js
+++ b/src/editor/contents/pasted_content_formatter.js
@@ -1,3 +1,5 @@
+import OfficeFormatter from "./pasted_content_formatter/office_formatter"
+
 export default class PastedContentFormatter {
   constructor(doc) {
     this.doc = doc
@@ -7,7 +9,7 @@ export default class PastedContentFormatter {
     this.#stripStyleElements()
     this.#unwrapPlaceholderAnchors()
     this.#stripTableCellColorStyles()
-    this.#rebuildOfficeLists()
+    new OfficeFormatter(this.doc).format()
     this.#unwrapWrappedListChildren()
     this.#nestStrayListChildren()
     this.#stripStrayListChildren()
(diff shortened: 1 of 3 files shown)


src/editor/contents/pasted_content_formatter.js
```diff
@@ -49,6 +50,143 @@ export default class PastedContentFormatter {
...
+  // paragraphs that declare their depth in an inline `mso-list:l0 level2 lfo1`
+  // and carry the bullet or number glyph in a <span style="mso-list:Ignore">.
+  // Nothing downstream understands either, so the items land as plain paragraphs
+  // with the marker baked into the text. Rebuild real lists from the level each
+  // paragraph declares.
+  //
+  // Everything this reads lives in inline attributes, so it stays correct after
+  // #stripStyleElements has dropped Office's style sheet.
+  #rebuildOfficeLists() {
+    for (const run of this.#officeListRuns()) {
```
jorgemanrubia: Extract a helper object `/home/jorge/Work/basecamp/lexxy/src/editor/contents/pasted_content_formater/office_formatter.js` so that you instantiate `OfficeFormatter` and send it `.format()`, and it encapsulates all these methods.

jorgemanrubia: 🤖 Done in 6a9aaf93 — `src/editor/contents/pasted_content_formatter/office_formatter.js`.

Note I spelled the directory `pasted_content_formatter/` (two `t`s) to match the file it sits beside, rather than the `formater` in your comment. Matches the `node_inserter.js` + `node_inserter/` pattern. Say the word if you actually wanted the other spelling.

**Moved** (all nine Office methods): `#rebuildOfficeLists` — now `format()` itself, still just the loop over runs — plus `#officeListRuns`, `#nextOfficeListParagraph`, `#isIgnorableBetweenOfficeListParagraphs`, `#replaceWithLists`, `#lastItemOf`, `#officeListTagName`, `#createOfficeListItem`, `#officeListLevel`. They dropped the now-redundant `Office` prefix (`#officeListLevel` → `#listLevel`); bodies are unchanged.

**Stayed:** `#stripStyleElements`, `#unwrapPlaceholderAnchors` and `#stripTableCellColorStyles` are not Office-specific. Neither are the three stray-list repairs — they all start from `querySelectorAll("ol, ul")`, so they fire for Gmail and rendered-view pastes too. They also have to keep running *after* the rebuild, which is why the `new OfficeFormatter(this.doc).format()` call sits mid-pipeline rather than at the top.

**Equivalence evidence.** Snapshotted the pre-refactor class and diffed `format()` output against the refactored one over 35 clipboard fixtures — the 11 from this PR verbatim, 12 more Office cases (level skips, `lfo` switches, marker variants, comments between paragraphs, Office lists inside table cells), and 12 exercising the repairs that stayed. All 35 byte-identical. To check the harness was worth anything I ran 8 mutants through it: 7 caught, including removing the `baseLevel` anchoring and the same-level tag switch. The one that escaped is the `claimed` set in `#listRuns` — dropping it costs redundant work but `replaceWith` on an already-detached paragraph is a no-op, so it produces no output difference and diffing can't see it.

Suites: 119 vitest, 147 paste browser tests (chromium) including all 11 Office regressions, eslint clean. The regression tests go through the real editor rather than calling privates, so none needed touching. Harness was scratch — not committed.

PR #1219, opened 2026-08-02
macOS text replacements: catch the deliveries the first fix missed

rosa:
Follow-up to #1213

That PR (shipped in 0.9.28) intercepts `CONTROLLED_TEXT_INSERTION_COMMAND` and converts a `"\n"` found in the` InputEvent`'s `data` into line break nodes. We retested that version and confirmed it fixed the problem in Safari, but it continued failing in Chrome. Now, testing this via BrowserStack:

| Delivery | Payload | #1213 | This PR |
|---|---|---|---|
| Chrome `insertReplacementText` | `data` string | fires only on `\n` | converted |
| Chrome macOS replacement via `insertText` | bare **string** command payload | deliberately ignored | converted |
| Any payload carrying **U+2028/U+2029** instead of `\n` | all of the above | passes the `"\n"` check | normalized |
| Uncontrolled insertion reconciled from the DOM | no command at all | unreachable | transform backstop |

**Real Safari 17.3 on macOS Sonoma, genuine OS substitution (Live session):** typing a misspelling and letting real macOS autocorrect fire — WebKit routes autocorrect and Text Replacement through the same `SpellingCorrectionCommand` — produced:

```
beforeinput inputType=insertReplacementText data=[null] dataTransfer(text/plain)=[74 68 65] targetRanges=1
```

`data` is **null**; the replacement text rides in `dataTransfer`. So the shipped fix cannot fire in Safari, matching WebKit source (`SpellingCorrectionCommand::inputEventData()` returns the correction only for textarea/input; for contenteditable it goes into `inputEventDataTransfer()`).

**Real Safari 17.3 + real Chrome 150 on macOS Sonoma (Automate, against the built bundle):**
- Baseline (main): an `insertReplacementText` carrying `U+2028` — in `data` (Chrome shape) or `dataTransfer` (Safari shape) — left a literal U+2028 in the serialized value on **both** browsers: `<p> Thanks,U+2028Kyle...</p>`. With this branch, all shapes serialize to `<p> Thanks,<br>Kyle...</p>`.
- Rendering probes on both engines: a literal `\n` shows as a line break under `white-space: pre-wrap` (the editor) and collapses in normal rendering (the submitted message) — exactly the customer's "right while composing, one line after submit". A literal U+2028 renders as one line in both, and `\n` via Safari's dataTransfer became two `<p>`s even before #1213 — so the reported symptom can only come from the string/`data` shapes, which this PR now covers.

## The fix

- `LexicalEditorElement##insertionTextFrom` extracts the text from all three payload shapes; `dataTransfer` is only consulted for `insertReplacementText`, so drops and yanks keep Lexical's own handling.
- `Contents#insertTextWithLineBreaks` normalizes `U+2028`/`U+2029` alongside CRLF.
- New `LineSeparatorsExtension` registers a `TextNode` transform that converts any literal separator that lands in a text node — through uncontrolled insertions, pastes, or stored content — into a line break node. It also heals previously saved content on re-edit.

### DIFF
--- a/src/editor/contents.js
+++ b/src/editor/contents.js
@@ -60,20 +60,6 @@ export default class Contents {
     }, { tag })
   }
 
-  // A literal "\n" stays visible in the editor, which Lexical renders with white-space: pre-wrap,
-  // but collapses to a space once serialized to HTML, so it has to become line break nodes.
-  insertTextWithLineBreaks(text) {
-    const normalizedText = text?.replace(/\r\n?/g, "\n")
-    const selection = $getSelection()
-
-    if (normalizedText?.includes("\n") && $isRangeSelection(selection)) {
-      selection.insertRawText(normalizedText)
-      return true
-    } else {
-      return false
-    }
-  }
-
   insertAtCursor(...nodes) {
     const selection = this.#insertableSelection()
     const inserter = NodeInserter.for(selection)
--- a/src/extensions/line_separators_extension.js
+++ b/src/extensions/line_separators_extension.js
@@ -1,0 +1,35 @@
+import { $createLineBreakNode, TextNode } from "lexical"
+import LexxyExtension from "./lexxy_extension"
+
+const LINE_SEPARATORS = /\r\n|[\n\r\u2028\u2029]/
+
+// A literal line separator inside a text node renders as a break in the editor, or not at all,
+// but never survives serialization to HTML. macOS text replacements deliver U+2028 LINE SEPARATOR,
+// and other insertions can carry "\n", so however the text arrives, a transform converts whatever
+// lands in a text node into line break nodes.
+export class LineSeparatorsExtension extends LexxyExtension {
+  get lexicalExtension() {
+    return this.defineExtension({
+      name: "lexxy/line-separators",
+      register(editor) {
+        return editor.registerNodeTransform(TextNode, $replaceLineSeparatorWithLineBreak)
+      }
+    })
+  }
+}
+
+// Replaces the first separator only: the split leaves dirty text nodes behind, so Lexical re-runs
+// the transform until none remain.
+function $replaceLineSeparatorWithLineBreak(textNode) {
+  const match = textNode.getTextContent().match(LINE_SEPARATORS)
+
+  if (match) {
+    let separatorNode
+    if (match.index === 0) {
+      [ separatorNode ] = textNode.splitText(match[0].length)
+    } else {
+      [ , separatorNode ] = textNode.splitText(match.index, match.index + match[0].length)
+    }
+    separatorNode.replace($createLineBreakNode())
+  }
+}
--- a/src/elements/editor.js
+++ b/src/elements/editor.js
@@ -1,4 +1,4 @@
-import { $addUpdateTag, $createParagraphNode, $getRoot, $getSelection, $hasUpdateTag, $isElementNode, $isLineBreakNode, $isRangeSelection, $isTextNode, $onUpdate, CAN_REDO_COMMAND, CAN_UNDO_COMMAND, CLEAR_HISTORY_COMMAND, COMMAND_PRIORITY_NORMAL, CONTROLLED_TEXT_INSERTION_COMMAND, KEY_ENTER_COMMAND, PASTE_TAG, SKIP_DOM_SELECTION_TAG, TextNode } from "lexical"
+import { $addUpdateTag, $createParagraphNode, $getRoot, $getSelection, $hasUpdateTag, $isElementNode, $isLineBreakNode, $isRangeSelection, $isTextNode, $onUpdate, CAN_REDO_COMMAND, CAN_UNDO_COMMAND, CLEAR_HISTORY_COMMAND, COMMAND_PRIORITY_NORMAL, KEY_ENTER_COMMAND, PASTE_TAG, SKIP_DOM_SELECTION_TAG, TextNode } from "lexical"
 import { buildEditorFromExtensions } from "@lexical/extension"
 import { ListItemNode, ListNode, registerList } from "@lexical/list"
 import { AutoLinkNode, LinkNode } from "@lexical/link"
@@ -45,6 +45,7 @@ import { FormatEscapeExtension } from "../extensions/format_escape_extension.js"
 import { LinkOpenerExtension } from "../extensions/link_opener_extension.js"
 import { PreventLexicalTripleClickExtension } from "../extensions/prevent_lexical_triple_click_extension.js"
 import { CustomAttachmentDragAndDropExtension } from "../extensions/custom_attachment_drag_and_drop_extension.js"
+import { LineSeparatorsExtension } from "../extensions/line_separators_extension.js"
 import { nextFrame } from "../helpers/timing_helper.js"
 
 
@@ -206,7 +207,8 @@ export class LexicalEditorElement extends HTMLElement {
       FormatEscapeExtension,
       LinkOpenerExtension,
       PreventLexicalTripleClickExtension,
-      CustomAttachmentDragAndDropExtension
+      CustomAttachmentDragAndDropExtension,
+      LineSeparatorsExtension
     ]
   }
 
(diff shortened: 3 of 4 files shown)


src/extensions/line_separators_extension.js
```diff
@@ -0,0 +1,36 @@
...
+import LexxyExtension from "./lexxy_extension"
+
+const LINE_SEPARATORS = /\r\n|[\n\r\u2028\u2029]/
+
+// A literal line separator inside a text node renders as a break in the editor, or not at all,
+// but never survives serialization to HTML. Deliveries no command handler sees can leave one
+// behind — macOS text replacements carry U+2028 LINE SEPARATOR, and uncontrolled insertions that
+// Lexical reconciles from the DOM can carry "\n" — so a transform converts whatever slips in
+// into line break nodes.
+export class LineSeparatorsExtension extends LexxyExtension {
```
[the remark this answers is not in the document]
rosa: Matching paste is deliberate, decided in #1213's review round: pasting multi-line text into a single-line editor already produces a line break — verified empirically at current main, where pasting "Thanks,\nKyle" into the single-line fixture yields `<p>Thanks,<br>Kyle</p>`. Only interactive Enter is guarded by `supportsMultiLine`. Gating this transform would make replacements the one insertion path that flattens, diverging from paste rather than converging with it. Added the requested regression case pinning that parity: test/browser/tests/modes/single_line.test.js.

src/editor/contents.js
```diff
@@ -60,10 +60,11 @@ export default class Contents {
...
   }
 
-  // A literal "\n" stays visible in the editor, which Lexical renders with white-space: pre-wrap,
-  // but collapses to a space once serialized to HTML, so it has to become line break nodes.
+  // A literal newline stays visible in the editor, which Lexical renders with white-space: pre-wrap,
+  // but collapses to a space once serialized to HTML, so it has to become line break nodes. macOS
+  // multi-line text replacements carry U+2028 LINE SEPARATOR rather than "\n".
   insertTextWithLineBreaks(text) {
-    const normalizedText = text?.replace(/\r\n?/g, "\n")
+    const normalizedText = text?.replace(/\r\n?|[\u2028\u2029]/g, "\n")
```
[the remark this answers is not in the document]
rosa: Fixed — added test/system/text_replacement_round_trip_test.rb: dispatches a genuine insertReplacementText InputEvent carrying U+2028 into the editor, saves through the form, asserts the rendered post keeps the `<br>` (visible: :all — Capybara filters br elements as invisible), and re-edits asserting the editor value round-trips. Fails without the src fix (verified by rebuilding against main's src: the `p br` assertion is what breaks).

rosa:
Amended in place (eabfd999) rather than stacking commits, per Rosa:

- **webkit CI failure fixed**: the two dataTransfer-shape tests failed only on webkit because WebKit's InputEvent constructor ignores the `dataTransfer` option, so the synthetic event carried nothing and the editor inserted nothing. The test helper now reattaches it via `Object.defineProperty` when the constructor drops it (product code unchanged — real Safari events carry a real dataTransfer, verified on BrowserStack earlier). The fallback path itself is proven on chromium/firefox.
- **New single-line regression test** pinning that replacements match paste behavior (see review thread for the rationale).
- **New Action Text round-trip system test** (editor → save → rendered post → re-edit), which fails without the fix.
- Commit message rewritten shorter.

Local runs: vitest 137/137, affected browser specs green on chromium + firefox (webkit can't launch on this machine — CI is the webkit check), the new system test green and non-vacuous, eslint clean.

src/extensions/line_separators_extension.js
```diff
@@ -0,0 +1,36 @@
+import { $createLineBreakNode, TextNode } from "lexical"
+import LexxyExtension from "./lexxy_extension"
+
+const LINE_SEPARATORS = /\r\n|[\n\r\u2028\u2029]/
```
samuelpecher, line 4: <s>This regexp duplicates `contents.js` L66, is it worth extracting to a common helper?</s>

Not applicable since we should replace the empirical input cleaner with the declarative transform.

src/extensions/line_separators_extension.js
```diff
@@ -0,0 +1,36 @@
...
+// but never survives serialization to HTML. Deliveries no command handler sees can leave one
+// behind — macOS text replacements carry U+2028 LINE SEPARATOR, and uncontrolled insertions that
+// Lexical reconciles from the DOM can carry "\n" — so a transform converts whatever slips in
+// into line break nodes.
+export class LineSeparatorsExtension extends LexxyExtension {
+  get lexicalExtension() {
+    return this.defineExtension({
+      name: "lexxy/line-separators",
+      register(editor) {
+        return editor.registerNodeTransform(TextNode, $replaceLineSeparatorWithLineBreak)
```
samuelpecher, line 15: If you want, Lexical provides [`registerLexicalTextEntity`](https://lexical.dev/docs/concepts/transforms#registerlexicaltextentity) for this type of predictable `pattern => Node` transform.

```js
registerLexicalTextEntity(
  editor,
  (text) => text.match(LINE_SEPARATORS),
  LineBreakNode,
  $createLineBreakNode 
)
```

I'm envious, I kept not finding places to use it.

src/editor/contents.js
```diff
@@ -60,10 +60,11 @@ export default class Contents {
     }, { tag })
   }
 
-  // A literal "\n" stays visible in the editor, which Lexical renders with white-space: pre-wrap,
-  // but collapses to a space once serialized to HTML, so it has to become line break nodes.
+  // A literal newline stays visible in the editor, which Lexical renders with white-space: pre-wrap,
+  // but collapses to a space once serialized to HTML, so it has to become line break nodes. macOS
+  // multi-line text replacements carry U+2028 LINE SEPARATOR rather than "\n".
   insertTextWithLineBreaks(text) {
```
samuelpecher: This feels duplicated now the extension transform will handle the split. I'd favor the transform and just cut this: the transform is probably the better fix. It also removes the duplicated regexp pattern.

As long as the tests pass, it removes the whole chain including `this.#handleReplacementText()`. The transform guarantees that no matter how the input arrives, it results in `LineBreakNode` in the editor state.

samuelpecher (commented):
👏 Transforms are a better call for these fixes: no matter the input, we get the editor state we want. I'd go one further and only keep the tests from the previous PR, and reach for the in-build `registerLexicalTextEntity`.

src/extensions/line_separators_extension.js
rosa: Oh, nice!

rosa:
Thanks @samuelpecher! 

> Transforms are a better call for these fixes: no matter the input, we get the editor state we want.

Done! The `TextNode` transform is now the entire fix. 

> I'd go one further and only keep the tests from the previous PR, and reach for the in-build `registerLexicalTextEntity`.

I couldn't make this work. It calls `setFormat` on the created node and needs `{start, end}` from `getMatch`), and `LineBreakNode` isn't a `TextNode`. I ran your snippet, and it throws `nodeToReplace should not be undefined` and drops the insertion.  Not sure if I'm doing something wrong here.

samuelpecher:
Sorry! The position needs to be `{start:, end:}` formatted and `registerLexicalTextEntity` indeed wants to call `setFormat` on tis replacement. It looks like we could get away with having a noop `setFormat` on replacementNode, but that might not be worth the simplification.

rosa:
I think it's not worth it, yeah... The resulting code is not smaller, and I think it doesn't fit very well. I've learned more now about `registerLexicalTextEntity`, and it seems it's not the best fit for this. The reason is that this seems to be built for entities that might change and need to be watched in two directions. For example, hashtags (`#') or mentions (`@`). You type `#dogs`, and you get a hashtag node. Then, if you continue editing your text, and you remove the `#`, leaving just `dogs`, you need the hashtag node to be converted back into plain text because that's not a hashtag anymore. 

However, in this case, the line breaks we want to replace aren't this kind of entity. They can't be edited into an invalid state like `#dogs` to `dogs`. It feels like a different thing.

PR #1230, opened 2026-08-11
Keep an image's alternative text and intrinsic size in attachment content

jeremy:
**Consumer action:** none. This widens what an allowed element may carry; it
never admits an element. A differently cased `allowedElements` declaration now
applies where it used to half-apply — see below.

An `` inside a custom attachment is sanitized by Lexxy every time the
attachment renders in the editor, so what it may carry is Lexxy's decision rather
than the host app's. `alt` and `width`/`height` were all being dropped. `alt` is
what a screen reader reads; `width`/`height` are what stop the line reflowing
while the image loads. None of the three can execute anything.

`alt` goes in the blanket list, since it is inert on every element it can appear
on. `srcset` deliberately stays out — it carries URLs, so it belongs to a consumer
that declares it.

`width`/`height` are scoped to `img` instead, via a new per-tag mechanism, because
`ALLOWED_ATTR` is not per-tag: putting them in the blanket list would also permit
`<table width="100000">` and `<td height="500">` in attachment content, which is
layout the editor previously stripped. The widening only ever applies to tags the
caller already permits.

## `buildConfig` now normalizes case

Review catch, folded in here because half of it is new here. DOMPurify lowercases
`ALLOWED_TAGS` and calls the functional `ADD_ATTR` with the lowercased tag and
attribute names, while `buildConfig` kept the caller's casing. `allowedElements`
is documented public API, so a differently cased declaration was accepted and then
partly ignored, with no error:

- `[ "IMG" ]` allowed the element and silently dropped the `width`/`height` this
  PR adds — that half is new, since `DEFAULT_TAG_ATTRIBUTES` is new.
- `[ { tag: "img", attributes: [ "GID" ] } ]` silently dropped the attribute it
  declared — that half predates the series.

Both go through the same two lines. `tagAttributes` also moves to a null
prototype, so a declared tag can never read through to an `Object.prototype` key:
`tagAttributes["constructor"]` would answer with a function and `ADD_ATTR` would
call `.includes` on it. Not reachable today — such an element is dropped before
its attributes are consulted — so that part is hardening rather than a fix.

## What the allowlist actually gates

Lexxy's allowlist is a **display gate** for attachment content, not a persistence
gate. `alt`/`width`/`height` were never being lost from storage: `exportDOM` emits
the stored `innerHtml` verbatim, and the document-level sanitize treats `content`
as an opaque attribute value (verified byte-identical through `sanitize()` and
`insertAdjacentHTML`, even with raw angle brackets in a nested `alt`). They were
only being lost from what the editor rendered.

So of the assertion points in the system test, the two in the editor are the ones
sensitive to `dom_purify.js`. The saved value and the rendered page pin something
narrower but still worth pinning — that nothing else in the round trip drops the
content on the way past.

## Verification

- `test/actiontext/image_attributes_test.rb` reads Lexxy's allowlist **out of
  source** rather than copying it, so widening one side without the other fails
  there. It asserts on the image node and requires a surviving value: matching a
  substring of the whole rendered document passed the `class` case with no class
  on the image at all (the dummy layout wraps content in
  `<div class="lexxy-content">`), and read Action Text's `style=""` as agreement.
  Values are now per-attribute, because `style="1"` is not valid CSS.
- `test/system/attachments/attachment_content_images_test.rb` drives the dummy app
  end to end — editor DOM, submitted value, rendered page, re-edited document. That
  needed the dummy mention partial to carry an avatar, since the consumer shape
  this exists for does; a `data:` URI keeps it self-contained and off the network.
- Mutation-checked. Revert the config change and the `alt`/size and tag-scoping
  cases fail. Drop the lowercasing and the tag half of the new casing test fails;
  lowercase the tag but not the attributes and the attribute half fails. Remove
  `class` from Action Text's allowed attributes and the old agreement assertion
  passes on the wrapper's class while the new one fails; feed `style` a uniform
  `"1"` and the old one passes on `style=""` while the new one fails.

The tag helper test now asserts the mention's parts rather than the partial's
exact markup, so adding an avatar doesn't make it brittle.

---

**Part of a series** re-filing #1227 at reviewable scope, after #1227 was reverted
from `main` in 8c64aa45. Merge order: #1228 → #1229 → **this** → #1231 → #1232 →
#1226. Nothing here is released.

*Draft: needs human review and a soak period before merging.*


jeremy:
## Verification

- **Unit, mutation-checked:** revert the config change and the `alt`/size case and the tag-scoping case both fail. The other three keep passing, which is the point — they are the controls that stop this from being a blanket widening.
- **Rails, both CI legs green**, including the new `test/actiontext/image_attributes_test.rb` (3 runs, 16 assertions) and the new system test.
- **bc3:** 73 system tests and 132 rich-text/mention/attachment tests, 0 failures, against a build carrying this change.

Nothing here needed a browser beyond what CI already runs — the widening is decided entirely by `buildConfig`, and the Action Text test reads the allowlist out of source so the two sides can't drift apart silently.

CI: green (9/9).

jeremy:
Handoff note for the whole campaign, including QA results and what still needs a human decision: #1234

test/actiontext/image_attributes_test.rb
```diff
@@ -0,0 +1,52 @@
...
+    (tag_scoped + (blanket & %w[ alt ])).uniq
+  end
+
+  test "the source still declares the image attributes this test reads" do
+    # Guards the regexes above: if the declarations are renamed or restructured,
+    # IMAGE_ATTRIBUTES silently empties and every assertion below passes vacuously.
+    assert_includes IMAGE_ATTRIBUTES, "alt"
+    assert_includes IMAGE_ATTRIBUTES, "width"
+    assert_includes IMAGE_ATTRIBUTES, "height"
+  end
```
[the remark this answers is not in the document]
jeremy: Half fixed, half declined — and the half you found is the one that mattered.

**Fixed.** The `& %w[ alt ]` intersection meant a newly blanket-allowed attribute sat outside the check while the header claimed "widening one side without the other fails here". It would not have. It now takes the whole blanket list minus one documented exemption, and pins the exemption too so it can't quietly become a place to hide disagreements. Mutation-checked: adding `longdesc` to ALLOWED_HTML_ATTRIBUTES now fails, and did not before.

While there: the old comment said Action Text drops `contenteditable` and `style` "by design". It keeps `style`. Both sides agree on it, so it belongs in the check rather than in an exception list.

**Not doing: dropping the source read for an explicit contract.** The property is that two allowlists in two languages agree, and Ruby has no public interface onto the JavaScript one. An explicit Ruby list converts a drift this test catches into one it cannot see — the Ruby copy keeps passing while Lexxy widens underneath it, which is precisely the failure the test exists for. The Vitest cases verify buildConfig against *themselves*; they cannot know what Action Text does.

STYLE.md:319 is a real rule and this is a real exception to it, so the reasoning is now in the file rather than in this thread. The coupling is bounded: the vacuity guard fails loudly on a rename instead of passing on an empty list, which is the failure mode that would actually hurt.

Leaving unresolved — the exception to the style rule is a judgment call and a human may take the other side of it.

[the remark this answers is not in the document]
jeremy: Making the call and resolving. The substantive defect you found is fixed (the `& %w[ alt ]` intersection that let new blanket attributes escape the check). On the source-read itself I'm keeping it, and that's a judgment I'll own rather than defer: the property is that two allowlists in two languages agree, Ruby has no interface onto the JS one, and a hardcoded copy converts a drift this test catches into one it can't see. STYLE.md:319 is a real rule and this is a real, documented exception to it — the reasoning is in the file, and the vacuity guard fails loudly on a rename rather than passing on an empty list. If a future reviewer wants the two-explicit-lists shape instead, it's a cheap swap, but it's strictly weaker for the drift this is guarding and I don't think it's the right trade.

test/javascript/unit/sanitization/dom_purify_image_attributes.test.js
```diff
@@ -0,0 +1,50 @@
+import { describe, expect, test } from "vitest"
```
jorgemanrubia, line 1: there is now a unit/sanitization folder, move the test there

test/system/attachments/attachment_content_images_test.rb
```diff
@@ -0,0 +1,50 @@
...
+
+# Images inside a custom attachment's content are sanitized every time the
+# attachment renders in the editor, so what an <img> may carry is decided by
+# Lexxy rather than by the app that wrote the markup. That makes it a round-trip
+# property: the editor, the saved value, the rendered page and the re-edited
+# document all have to agree, and a difference only shows up after a reload.
+#
+# alt is what a screen reader reads; width and height are what stop the line
+# reflowing while the image loads. Both are inert, and both were being dropped.
+class AttachmentContentImagesTest < ApplicationSystemTestCase
```
jorgemanrubia, line 15: also move the system tests relative to attachments to an `attachments` subfolder

jorgemanrubia (approved):

test/javascript/unit/sanitization/dom_purify_image_attributes.test.js
jorgemanrubia: 🤖 Moved to `test/javascript/unit/sanitization/dom_purify_image_attributes.test.js` in https://github.com/basecamp/lexxy/commit/4de0a4a90d1b6420c2f91d46b37a2e8083e7ae89, next to `dom_purify_attachment_url.test.js`. Imports switched to the `src/…` alias, matching what the attachment url test does — the extra directory level changes the relative depth, so the alias avoids a third `../`.

Checked it is genuinely collected rather than silently dropped: the full `yarn test` run picks it up at the new path and all 5 of its tests execute (22 files, 144 tests, green).

One boundary call: `test/javascript/unit/dom_purify_stimulus.test.js` is also a sanitization test and is still sitting at the top of `unit/`. I left it alone since it isn't part of this branch, but say the word and I'll move it in a follow-up.

test/system/attachments/attachment_content_images_test.rb
jorgemanrubia: 🤖 Done in https://github.com/basecamp/lexxy/commit/aa4f63ac2c5134107eab397c85b79b5a2a94bdef. Pure `git mv`, no renames and no reindentation.

Moved — the three whose subject is attachments:

- `test/system/attachments/attachments_test.rb`
- `test/system/attachments/attachment_url_round_trip_test.rb`
- `test/system/attachments/attachment_content_images_test.rb`

Left where they are, taking the narrow reading. These reach for an attachment as the vehicle for something else, so the subject isn't attachments: `editor_to_string_test`, `editor_empty_blank_test`, `empty_status_test`, `events_test`, `action_text_load_test`, `prompts_test`. Three more are genuinely arguable and I'd rather you drew the line than I guessed: `gallery_test`, `upload_focus_test`, `mention_round_trip_test` — all attachment machinery, but each is about a gallery, about focus during upload, and about mentions respectively. Point at any of them and they follow in a second commit.

Verified the nested directory is still collected instead of quietly skipped: `bin/rails test:system` runs 75 tests / 287 assertions both before and after the move — identical, so nothing dropped out. `bin/rails test test/system/attachments` on its own runs the 13. Rubocop and eslint clean.

Two things I did not do, since you asked for a move: I didn't namespace the classes `Attachments::…` the way `test/system/trix` namespaces `Trix::…`, and I didn't shorten `attachments/attachments_test.rb` or `attachments/attachment_content_images_test.rb` now that the directory carries the word. Both are one commit away if you want the consistency.

jorgemanrubia:
🤖 **Force-pushed: restacked onto the current `dompurify-3-4-13`.**

This branch was one commit behind #1229 — it was missing https://github.com/basecamp/lexxy/commit/c423db760ebe58690ec52dd15a903f278a0cb8bf (the CI job caps and bounded `playwright install-deps` retry). Its base was the parent of that commit, so everything above it in the stack was resolving against a stale tree, and #1226's "Files changed" was showing commits belonging to lower PRs.

Rebased with no conflicts. `29adc365` is content-identical to the previous tip `aa4f63ac` — the only difference between the two trees is `.github/workflows/ci.yml`, picked up from the commit below, and this PR's own four commits are unchanged.

This was the bottom of a four-branch restack: #1230 → #1231 → #1232 → #1226, each onto the freshly-pushed tip of the one below. Every consecutive pair now reports up to date with 0 behind.

— @jorgemanrubia

jorgemanrubia:
🤖 Applied the confirmed review findings on this branch. Two commits, plus a body
rewrite.

**`buildConfig` didn't normalize case, so the public API silently half-applied** —
[`1efcd685`](https://github.com/basecamp/lexxy/commit/1efcd685c96d6c0389689e8488dba50aceb3b92e)

DOMPurify lowercases `ALLOWED_TAGS` and calls the functional `ADD_ATTR` with the
lowercased tag and attribute names, while `buildConfig` kept the caller's casing.
`allowedElements` is documented public API (`home/docs/extensions.md:42`), so a
differently cased declaration was accepted and then partly ignored:
`[ "IMG" ]` allowed the element and dropped the `width`/`height` this PR adds,
`[ { tag: "img", attributes: [ "GID" ] } ]` dropped the attribute it declared, and
`[ { tag: "IMG", attributes: [ "gid" ] } ]` failed on the tag key. `tagAttributes`
also moves to a null prototype, which removes an unreachable
`ADD_ATTR("width", "constructor")` `TypeError` — hardening, not a fix.

It landed **here rather than on #1231** because the `width`/`height` half is new
here: `DEFAULT_TAG_ATTRIBUTES` is what makes `[ "IMG" ]` lose attributes, so this is
the PR that would have shipped the new symptom. It also puts the fix below the two
PRs above it, so it flows up the stack once, and next to the test file that already
covers the widening. Heads up for the restack: #1231 moves this loop into
`allowlistFor()`, so that rebase will want a manual resolve — the change is the two
`const` lines and the `Object.create(null)`.

Mutation-checked: drop the lowercasing entirely and the tag half of the new test
fails; lowercase the tag but not the attributes and the attribute half fails.

**The Action Text agreement test passed for the wrong reason on `class` and
`style`** — [`506259d2`](https://github.com/basecamp/lexxy/commit/506259d26418aa8f6cb780c2a0ed1ebc28cc7bf1)

It substring-matched the whole rendered document, and the dummy layout wraps
content in `<div class="lexxy-content">` — so `ActionText::Content.new("").to_s`
already satisfied the `class` assertion with no class on the image at all. `style`
was worse: `style="1"` isn't valid CSS, comes back as `style=""`, and the substring
check read name-kept-value-destroyed as agreement. The two attributes where the
sanitizers differ in kind were the two the test couldn't see.

Now scoped to the image node and requiring a surviving value, with per-attribute
input values so `style` gets real CSS. Not asserting value equality — Action Text
normalizes `color: red` to `color:red;` — and `color`/`background-color`, the two
properties Lexxy allows, both round-trip. Mutation-checked both ways: remove
`class` from Action Text's allowed attributes and the old assertion passes on the
wrapper's class while the new one fails; feed `style` the uniform `"1"` and the old
one passes on `style=""` while the new one fails. No assertions removed.

**Half the "round trip" is insensitive to this PR** — same commit, comments and body
only.

`exportDOM()` emits `content: this.innerHtml` — the raw stored string, never the
sanitized DOM — and import is `parseAttachmentContent(getAttribute("content"))`
verbatim. The document-level sanitize at `src/elements/editor.js:364` treats
`content` as an opaque attribute value; I re-confirmed it byte-identical through
`sanitize()` → `insertAdjacentHTML` with raw angle brackets in a nested `alt`, zero
script nodes. So the helper comment claiming the submitted value is "re-serialized
**and sanitized**" was wrong on *sanitized*, and the body's four guards were two.
The assertion still earns its place — it pins that export doesn't lose the content —
and the body is reframed around what this actually is: Lexxy's allowlist is a
display gate for attachment content, not a persistence gate. `alt`/`width`/`height`
were never being lost from storage, only from what the editor rendered.

**Noted, not changed** (judgement calls, flagged rather than decided): unbounded
`width`/`height` surviving in attachment content — blast radius is editor-only,
since `content` isn't in `ActionText::Attachment::ATTRIBUTES` and mentions
re-render from the sgid — and the `[^\]]*` regex extraction in the Ruby test, which
can truncate the attribute list silently.

**Verified:** 145 vitest / 13 `bin/rails test` / 75 runs, 287 assertions
`bin/rails test:system` / 640 Playwright chromium, all green. WebKit cannot launch
on this machine (`libicu74` missing), so it is untested here, not passing.

PR #1231, opened 2026-08-11
Give Lexxy its own DOMPurify instance, and an allowlist per editor

jeremy:
**Consumer action:** none here **if the series lands in order.** #1232 adds a CSP
requirement that exists *because* of this change, and it is not optional for an
npm consumer running enforced Trusted Types: this PR creates a second DOMPurify
instance, whose lazily created policy contends for the `dompurify` name the host's
instance already claimed, and the `lexxy`-named policy that resolves it arrives in
#1232. So this must not be merged and soaked alone for those consumers. Read
#1232 with it.

Two faults, one mechanism: Lexxy was calling `setConfig()` on the DOMPurify
singleton.

## It disarmed the host app's sanitizer

That singleton is shared with the host app, and DOMPurify treats a persistent
config as **final** — once `setConfig()` has run, every later
`sanitize(html, config)` anywhere in the bundle silently ignores its own config
argument. An app sanitizing untrusted HTML with `{ ALLOW_DATA_ATTR: false }`
would keep passing that option and stop getting it the moment a Lexxy editor
connected. No error, nothing visible at the call site.

Calling dompurify's default export with a window returns a fresh instance. Ours
carries our hooks and config; neither side can reach the other.

**Scope: npm consumers only.** `rollup.config.mjs` leaves `dompurify` out of
`external`, so the Rails asset build inlines its own copy — there is no shared
module instance for Lexxy to reach, and nothing an app on that path could have had
disarmed. `rollup.config.npm.mjs` marks it external, which is where the sharing
comes from. The same boundary applies to the policy-name contention above: the
inlined singleton's policy is created lazily and that instance never sanitizes, so
it never contests the name either.

## It let the last editor to connect decide for all of them

One module-level config meant exactly that. Not cosmetic: an editor's `value` is
sanitized on read, so a rich editor sharing a page with a plain one silently
dropped its own headings, lists and links **from the value it submitted**. That
is data loss, and only on a page with more than one editor.

The config is now keyed by the Lexical editor — the identity both call sites
already have, as `this.editor` on the element and as `createDOM()`'s second
argument — and passed to each `sanitize()` call rather than set persistently, so
no global sanitizer state is left even on our own instance.

## And the fallback had the same fault

`CustomActionTextAttachmentNode` is exported from `src/index.js`, so it can be
registered in a Lexical editor Lexxy did not build. That editor never calls
`register()`, so every attachment it renders resolves through the fallback — which
was reassigned on every `register()`. Last-editor-wins, reintroduced on the one
path with no editor to key on and no way to observe which config it got. It is now
built once and never reassigned.

It was also `buildConfig([])`, which allows nothing, so markup a standalone
consumer kept before this series was silently stripped. It now declares no
allowlist at all, leaving DOMPurify's own default policy in place — our hooks and
`FORBID_ATTR` still apply, so it is a default, not a passthrough. `buildConfig`
therefore distinguishes "no allowlist" from "an empty one"; an editor registering
with an empty allowlist still gets the refusal it asked for.

## Not an API change

`createDOM(_config, editor)` is a Lexical override, and Lexical has always called
it with both arguments. Neither `sanitize` nor `setSanitizerConfig` is exported
from `src/index.js`. Nothing here is reachable from outside the package.

## Verification

Both levels, because neither catches the other. Unit tests assert isolation in
both directions against the real singleton; a Playwright fixture puts two
differently configured editors on one page — the reproduction that fails on
`main`.

**The regression that must keep passing:** two editors with differing attachment
configs, where the one denying `content` strips it and the one allowing it keeps
it.

The fallback path is covered through the exported node in a standalone Lexical
editor: it asserts the deterministic default rather than emptiness, and asserts the
result **does not change** after another editor registers. Mutation-checked in both
directions — restoring `buildConfig([])` fails the first and only the first,
restoring the reassignment fails the second and only the second.

---

**Part of a series** re-filing #1227 at reviewable scope, after #1227 was reverted
from `main` in 8c64aa45. Merge order: #1228 → #1229 → #1230 → **this** → #1232 →
#1226. Nothing here is released.

*Draft: needs human review and a soak period before merging.*


jeremy:
## bc3 QA: this fixes a live bug on a shipping path, and I can show it

The disarm this PR describes is not hypothetical for bc3. `app/javascript/rich_text/sanitize_attachment_content.js` imports the DOMPurify **singleton** and sanitizes with a per-call config:

```js
import DOMPurify from "dompurify"

const sanitizerConfig = {
  ALLOWED_TAGS: [ "bc-mention", "span", "img" ],
  ALLOWED_ATTR: [ "alt", "class", "gid", "height", "src", "title", "width" ],
  CUSTOM_ELEMENT_HANDLING: { tagNameCheck: /^bc-mention$/, ... }
}

export const sanitizeAttachmentContent = (html) => DOMPurify.sanitize(html, sanitizerConfig)
```

Its own comment says where it is used: **the chat pending-line controller and the Trix paste serializer** — "so both surfaces apply the same allowlist and rejection policy."

Under 0.9.29 they don't. Once a Lexxy editor connects, `setConfig()` has run, and DOMPurify ignores that config argument. Reproduced with bc3's real config and dompurify 3.4.0 (bc3's pinned version):

| | `<bc-mention>` | `<a href>` | `<p>` |
|---|---|---|---|
| Before any editor connects | kept | **stripped** | **stripped** |
| After a 0.9.29 editor connects | kept | **kept** | **kept** |
| After an editor from this PR connects | kept | **stripped** | **stripped** |

Row 2 is bc3 still passing `sanitizerConfig` at the call site, still believing it applies, and getting Lexxy's allowlist instead. Row 3 is this PR.

I have not assessed how far that widening can be pushed on those two paths — that needs someone who knows what bc3 does with the output downstream — but bc3 clearly intends `["bc-mention", "span", "img"]` on pasted and chat-composed attachment content and is not getting it today.

## Rest of the bc3 QA

Against `integration/qa-build` (`4b0c9b9d`), in a scratch bc3 worktree — nothing committed or pushed to bc3, and the worktree has been restored to `0.9.29`:

- **bc3 system suite: 73 tests, 0 failures.** Messages, comments, chat rooms, docs, to-dos, schedule, check-in answers.
- **Rich-text / mention / attachment tests: 132 tests, 0 failures.** Includes `rich_text_attachments_test`, `rich_text_helper_test`, `mentions_notifier_test`, `chats/lines_controller_api_test`, `messages_controller_test`.
- bc3 boots and the gem loads from a git ref.

## Browser check of the named regression

The regression this PR must not lose — two editors, differing attachment configs — verified in Chromium against the integration branch:

- editor allowing attachments keeps comment-bearing `content` ✅
- editor **denying** attachments strips it ✅
- rich editor keeps its own `<h1>`/`<ul>`/`<a>` with a plain editor also on the page ✅
- plain editor still strips what its own preset denies (control) ✅
- two editors on one page, no console errors ✅

---

CI: green (Rails-main and Rails-8.1 legs both pass; the one earlier red leg was a Selenium flake in `Trix::FromTrixToLexxyTest#test_gallery`, cleared on re-run).

jeremy:
Handoff note for the whole campaign, including QA results and what still needs a human decision: #1234

test/javascript/unit/helpers/sanitization_helper.test.js
```diff
@@ -0,0 +1,58 @@
...
+  expect(sanitize("<span>keep</span><script>evil()</script>", editor)).toBe("<span>keep</span>")
+})
+
+test("each editor keeps its own allowlist", () => {
+  const other = { name: "a second editor" }
+  setSanitizerConfig(other, [ "strong" ])
+
+  expect(sanitize("<span>a</span><strong>b</strong>", editor)).toBe("<span>a</span><strong>b</strong>")
+  expect(sanitize("<span>a</span><strong>b</strong>", other)).toBe("a<strong>b</strong>")
+})
```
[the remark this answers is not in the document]
jeremy: Fixed — you're right, and it was the regression the PR description explicitly names, which makes the gap worse than an ordinary missing case.

Added both directions of the `content` allow/deny pair, each registering the *other* editor last, which is how the module-level config actually failed. `content` is the attribute that matters because it carries the attachment's serialized markup — losing it destroys the attachment rather than trimming it.

Mutation-checked against the old behaviour: stop keying the config by editor and both new cases fail (alongside the tag-level one, which is the control).

src/elements/editor.js
```diff
@@ -361,7 +361,7 @@ export class LexicalEditorElement extends HTMLElement {
 
   #readSanitizedEditorValue() {
     return this.editor?.read(() => {
-      return sanitize($generateHtmlFromNodes(this.editor, null))
+      return sanitize($generateHtmlFromNodes(this.editor, null), this.editor)
```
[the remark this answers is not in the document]
jeremy: Not doing this — and it's a close enough call that I'm leaving the thread open.

The rule you're invoking is right in general, and I've just added exactly this kind of test on the sibling PR where it does buy something. Here I don't think it does, for one specific reason: **the server leg is editor-agnostic.**

The bug is that an editor's `value` depended on which editor connected last. Once the value is right, Loofah and Action Text see one HTML string and have no concept of how many editors produced it. A two-editor round-trip test would exercise the same server path as the single-editor round-trip tests already in `test/system/`, and to fail it would need a server stage that treats identical HTML differently based on editor count. There isn't one.

The multi-editor half — the part that is genuinely about the custom-element lifecycle, cached value reads and the bundled build — is covered end to end in a real browser by `test/browser/tests/editor/sanitizer_isolation.test.js`, which fails on `main`. Verified by hand in Chromium against the integration branch too.

So the coverage exists on both sides of the seam; what a Capybara multi-editor test would add is a second copy of the server leg, plus a second editor in the dummy app's form to maintain. If someone disagrees, the cheap version is adding a second editor to the existing round-trip fixture rather than a new test — but I'd want a named failure it catches first.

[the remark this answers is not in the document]
jeremy: Making the call and resolving. The rule you invoke is right in general — and I just added exactly this kind of test on #1229 where it buys something — but here I don't think it does, and that's a judgment I'll own rather than park.

The bug was that an editor's `value` depended on which editor connected last. Once the value is correct, Loofah and Action Text see one HTML string with no concept of how many editors produced it. A two-editor round-trip would exercise the same server path as the single-editor round-trip tests already in `test/system/`, and to fail it would need a server stage that treats identical HTML differently by editor count — there isn't one. The genuinely multi-editor half (custom-element lifecycle, cached value reads, the bundled build) is covered end to end in a real browser by `sanitizer_isolation.test.js`, which fails on `main`.

So both sides of the seam are covered; a Capybara multi-editor test would add a second copy of the server leg plus a second editor to maintain in the dummy form. If a concrete failure it would catch surfaces later, the cheap version is a second editor in the existing round-trip fixture — but I'm not adding it speculatively.

jorgemanrubia: Let's extract an `editor/sanitized.js` class (`SanitizedEditor`) that receives the editor as an instance and is exposed via `editor.sanitized` property. It is a property of the editor so that you can do `editor.sanitized.value`. That new class encapsulates the configuration handling we have now in `sanitization_helper`. Instead of the `setSanitizerConfig` you could do `editor.sanitized.allowed_tags=...`. This will leave the sanitization helpers module thinner and a proper entity encapsulation the sanitization logic, that is now editor-instance based.

jorgemanrubia: 🤖 Done in https://github.com/basecamp/lexxy/commit/ad9cf8a67aed3580abbf763e3dda6c4ed0137a40 — `src/editor/sanitized.js`, exposed as `editorElement.sanitized`.

```js
// src/elements/editor.js — from $initialEditorState
this.sanitized = new SanitizedEditor(editor)
this.sanitized.allowedTags = this.#getAllowedElements(editor)

// the value read, which was #readSanitizedEditorValue()
get value() {
  return this.cachedValue ??= this.sanitized?.value ?? null
}

// src/nodes/custom_action_text_attachment_node.js — createDOM(_config, editor)
figure.insertAdjacentHTML("beforeend", SanitizedEditor.for(editor).sanitize(this.innerHtml))
```

Public surface: `for(editor)`, `allowedTags=`, `value`, `sanitize(html)`. `#readSanitizedEditorValue` is gone — the getter says it in one line now.

**Where I drew the line.** The class took the per-editor state and nothing else: the `WeakMap`, the fallback, `buildConfig` and the `sanitize` call. What stayed in `sanitization_helper` is the attachment `url` hook and its constants — `attachmentUriFilterHook`, `isAttachmentTag`, the two URI regexes, `URI_BEARING_ATTACHMENT_ATTRIBUTES`. Those are pure functions of a tag and an attribute value with no editor in sight; making them methods would have bought a `this` they never use, and it's also the material you asked to be moved *into* that helper in #1229. The helper is now exactly one concern, which is the thinning you were after.

**`allowedTags`, not `allowed_tags`.** STYLE.md names `camelCase` explicitly and eslint enforces it, so I camelCased without asking. One word overrules me if you meant it literally. Worth flagging a second option while we're naming it: the value is element descriptors, not tags — strings *or* `{ tag, attributes }` — and it comes from `#getAllowedElements`, so `allowedElements` would be the more honest name. I kept your word rather than layering a second rename on top of the first; say which and I'll switch it.

**`setSanitizerConfig` removed, not delegated.** Checked before deleting: neither `sanitize` nor `setSanitizerConfig` is exported from `src/index.js` or the `lexxy_helpers` chunk, and the only callers anywhere were `elements/editor.js`, `custom_action_text_attachment_node.js`, and this PR's own unit test — nothing in `home/docs`, nothing in `test/dummy`. `setSanitizerConfig` predates this PR (it's on `main`), but only as a module-internal function, so there is no consumer to keep a shim for.

**One deviation.** The instance takes the *Lexical* editor rather than the element, unlike `Contents`/`Selection`/`Clipboard`. The allowlist is set from `$initialEditorState`, which runs inside `buildEditorFromExtensions` — before `this.editor` is assigned — and `#loadInitialValue` runs immediately after it in the same callback. Constructing from the element would mean either registering by an identity that doesn't exist yet, or moving configuration after `this.editor = …`, which would sanitize the initial value's attachment content with the previous editor's allowlist. It's also the identity the `WeakMap` already keyed on and the only one a node has, so `for()` stays a direct translation rather than a lookup through `getRootElement().closest()`.

Two smaller notes:

- The fallback for an editor that never declared an allowlist is kept, but it's now an editor-less instance carrying only the most recent allowlist. Parking the last configured *instance* there would pin its Lexical editor, and so its DOM, for the life of the page.
- After this, an element's `value` reads through its own instance and can't pick up a neighbour's allowlist by construction; the one place per-editor resolution can still go wrong is `for()`. So the moved unit tests (`test/javascript/unit/editor/sanitized.test.js`, was `unit/helpers/sanitization_helper.test.js`) resolve through `for()` where per-editor behaviour is the point. Mutation-checked: make `for()` always return the fallback and 4 tests fail — the three per-editor cases plus the attachment case in `unit/editor/sanitizer_isolation.test.js`.

**Bonus:** the helper no longer imports `config/dom_purify`, so the circular import I flagged in #1229 is gone. `npx rollup -c` warns `src/config/dom_purify.js -> src/helpers/sanitization_helper.js -> src/config/dom_purify.js` at `2fec9929` and is silent at this commit.

**Tests.** `npx vitest run`: 154 passed, 24 files. Browser, chromium: all 642 green (two needed a retry — the leak case below and one plain-text paste case). `sanitizer_isolation` passes at both levels, and `sanitizer_isolation` + `paste/xss_sanitization` + `paste/table_cell_styles` pass on firefox too. Two caveats, stated rather than buried: **webkit cannot launch on this machine** (`browserType.launch: Host system is missing dependencies`), so nothing here is webkit-verified; and `editor/leak.test.js` is flaky — it failed 1 of 3 runs here, and 2 of 8 runs at `2fec9929` with the same fixture (including the *listener* case, which this change can't touch), so it's pre-existing rather than something I introduced. I did check the obvious way this refactor could have leaked, and that's what the editor-less fallback above is for.

src/editor/sanitized.js
```diff
@@ -0,0 +1,66 @@
...
+//    how every other editor on the page sanitized. That is not cosmetic: an
+//    editor's `value` is sanitized on read, so a rich editor sharing a page with
+//    a plain one would silently drop its own headings, lists and links from the
+//    value it submits.
+//
+// An instance is built around the Lexical editor because that is the identity
+// both call sites have to hand: the element holds it as `this.editor` and
+// exposes its sanitizer as `editorElement.sanitized`, while nodes only receive
+// it as the second argument to createDOM() and resolve the sanitizer with for().
+export default class SanitizedEditor {
```
jorgemanrubia: let's rename to EditorSanitizer

src/editor/sanitized.js
```diff
@@ -0,0 +1,66 @@
...
+  constructor(editor) {
+    this.#editor = editor
+  }
+
+  set allowedTags(allowedTags) {
+    this.#config = buildConfig(allowedTags)
+    SanitizedEditor.#register(this)
+  }
+
+  get value() {
```
jorgemanrubia: lets remove the .value method and make that an editor sanitizer only exposes the `sanitize(html)` method. It is weird that it does both things (serving value and sanitizing arbitrary strings).

src/elements/editor.js
```diff
@@ -755,7 +749,8 @@ export class LexicalEditorElement extends HTMLElement {
   }
 
   #configureSanitizer(editor) {
-    setSanitizerConfig(this.#getAllowedElements(editor))
+    this.sanitized = new SanitizedEditor(editor)
```
jorgemanrubia: call it `this.sanitizer`

jorgemanrubia: Add a public static method `this.sanitizer = EditorSanitizer.register(editor, this.extensions.allowedElements))`.

Also, let's make `allowedTags` an internal property. Move the logic to determine those to the EditorSanitizer class, out of editor.js.

```
#getAllowedElements(editor) {
    return this.#getImportableTags(editor).concat(this.extensions.allowedElements)
  }

  #getImportableTags(editor) {
    const tags = Array.from(editor._htmlConversions.keys())
    return tags.filter(tag => !tag.startsWith("#"))
  }
```

src/editor/sanitized.js
jorgemanrubia: 🤖 Renamed in https://github.com/basecamp/lexxy/commit/b01d59ddf8dd48fb9ade4888d70aabff2e0e1ab3.

The file went with it — `src/editor/sanitizer.js` — following `configuration.js` / `EditorConfiguration`, and so did the test, `test/javascript/unit/editor/sanitizer.test.js`. `EditorSanitizer` reads better than what I had anyway: it's the editor's sanitizer, not a kind of editor.

src/editor/sanitized.js
jorgemanrubia: 🤖 Removed in https://github.com/basecamp/lexxy/commit/b01d59ddf8dd48fb9ade4888d70aabff2e0e1ab3. The whole public surface is now `sanitize(html)`, plus the two statics.

You were right that it was doing two things, and dropping it paid off more than I expected: **`value` was the only member that needed a Lexical editor.** With it gone the sanitizer holds no editor at all — just a config — and the editor is used at registration and nowhere else, as the `WeakMap` key and to derive importable tags. Which in turn means the fallback can be the most recently registered sanitizer, instead of the editor-less stand-in it had to be last round to avoid pinning a disconnected editor's DOM. That whole paragraph of justification is gone from the file.

`editor.js` composes it again, which is what it did before this class existed:

```js
get value() {
  return this.cachedValue ??= this.#readSanitizedEditorValue()
}

#readSanitizedEditorValue() {
  return this.editor?.read(() => {
    return this.sanitizer.sanitize($generateHtmlFromNodes(this.editor, null))
  }) ?? null
}
```

`cachedValue ??=` is untouched, and the `?? null` still covers a read before the editor exists. `this.sanitizer` is left unguarded on purpose — it and `this.editor` are set in the same breath, so if one is there the other is.

src/elements/editor.js
jorgemanrubia: 🤖 Done in https://github.com/basecamp/lexxy/commit/b01d59ddf8dd48fb9ade4888d70aabff2e0e1ab3 — `this.sanitizer`, alongside `this.contents` / `this.selection` / `this.clipboard`.

jorgemanrubia: 🤖 Done in https://github.com/basecamp/lexxy/commit/b01d59ddf8dd48fb9ade4888d70aabff2e0e1ab3. Both halves — `register` and the allowlist derivation moving in. Final shape:

```js
// elements/editor.js — #configureSanitizer is now one line
this.sanitizer = EditorSanitizer.register(editor, this.extensions.allowedElements)

// editor/sanitizer.js
static register(editor, allowedElements) {
  const sanitizer = new EditorSanitizer(this.#allowedElementsFor(editor, allowedElements))
  this.#instances.set(editor, sanitizer)
  this.#fallback = sanitizer

  return sanitizer
}

static #allowedElementsFor(editor, allowedElements) {
  return this.#importableTags(editor).concat(allowedElements)
}
```

`#getAllowedElements` and `#getImportableTags` moved across unchanged, as private statics, and are gone from `editor.js`.

**`register` and `for` both stay, and neither is dead.** They are different operations at different call sites: an element registers once while its editor is being built, and every attachment node resolves on each `createDOM()`. Collapsing them would mean either a `register` that nodes call with an allowlist they don't have, or a `for` that silently creates. The class reads as write-once, read-many, so I kept both names.

**The `_htmlConversions` timing holds — probed, not assumed.** Registration happens where the old `#getImportableTags` call did, inside `$initialEditorState`, so I instrumented the derivation and counted what was actually there: **35 importable tags for a rich editor, 17 for a plain one**, at the moment `register()` runs. Lexical has the conversions in place by then. That difference is also *why* `sanitizer_isolation` can tell the two editors apart, so the existing tests are already leaning on it.

**Naming.** The allowlist is internal now, which settles the open question from the last round in favour of the truthful name: it holds element descriptors, strings *or* `{ tag, attributes }`, so it is `allowedElements` throughout — matching `extensions.allowedElements` and `buildConfig`'s own parameter. Nothing outside the class sets or reads it, so there is no setter left behind; `register` is the only way in.

**One behaviour change, in the safer direction.** The unused fallback now starts as `buildConfig([])` rather than `{}`, so before any editor has registered it allows nothing instead of falling through to DOMPurify's defaults. That is closer to what the fallback comment always claimed it was for. After the first `register` it is the most recent allowlist either way.

**Verification.** `npx vitest run`: **155 passed / 24 files** (one new case, that an editor keeps the tags it can import). Playwright chromium, full suite: **all 642 green**. Firefox subset — `sanitizer_isolation`, `paste/xss_sanitization`, `paste/table_cell_styles` — 18 passed. Rollup is still silent, so the circular import stays gone.

Mutation check, re-run against the registration path since that is what changed: force `for()` to always miss and **4 tests fail** — the three per-editor cases in `unit/editor/sanitizer.test.js` plus the attachment case in `unit/editor/sanitizer_isolation.test.js`.

Two caveats, stated rather than buried. **WebKit still cannot launch on this machine** (`browserType.launch: Host system is missing dependencies`), so nothing here is webkit-verified. And two chromium tests needed a retry: `editor/leak.test.js`, flaky as before, and `attachments/attachments.test.js` — I chased that second one rather than shrugging at it, because attachment `createDOM` is exactly what this touches. It is pre-existing flakiness under high parallelism at the same rate on both sides: **4 failures in 304 runs on this branch, 4 in 304 at `2fec9929`**, and it does not fail at all when the worker count is capped.

src/nodes/custom_action_text_attachment_node.js
```diff
@@ -72,11 +72,13 @@ export class CustomActionTextAttachmentNode extends DecoratorNode {
...
 
-  createDOM() {
+  createDOM(_config, editor) {
     const figure = createElement(this.tagName, { "content-type": this.contentType, "data-lexxy-decorator": true, draggable: true })
     figure.dataset.lexicalNodeKey = this.__key
 
-    figure.insertAdjacentHTML("beforeend", sanitize(this.innerHtml))
+    // Resolved from the editor so this content is sanitized with its own
+    // allowlist rather than whichever editor connected most recently.
+    figure.insertAdjacentHTML("beforeend", EditorSanitizer.for(editor).sanitize(this.innerHtml))
```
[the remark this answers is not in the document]
jorgemanrubia: 🤖 You're right on both counts, and the second one is the worse of the two. Fixed in https://github.com/basecamp/lexxy/commit/570d604b8d546ab359fb910c0bbacf0912d4eaa4.

The fallback was wrong in two independent ways:

1. **Reassigned on every `register()`**, so an unregistered consumer sanitized with whichever Lexxy editor initialised most recently. That is last-editor-wins — the bug this class exists to remove — reintroduced on the one path that has no editor to key on and no way to observe which config it got. It is now fixed: built once, never reassigned.
2. **`buildConfig([])` allows nothing**, so markup a standalone consumer kept before this series was silently stripped.

Rather than documenting the footgun, the fallback now declares **no allowlist at all**, leaving DOMPurify's own default tag and attribute policy in place — which is what that consumer had when `sanitize()` was a bare `DOMPurify.sanitize(html)`. Our hooks and `FORBID_ATTR` still apply, so it is a default, not a passthrough.

So `buildConfig` now distinguishes "no allowlist" from "an empty one": called with none it leaves `ALLOWED_TAGS`, `ALLOWED_ATTR` and `ADD_ATTR` out rather than emptying them. An editor that registers an empty allowlist still gets the refusal it asked for. Keeping this inside `buildConfig` rather than assembling a separate default config matters one PR up: #1232 puts `TRUSTED_TYPES_POLICY` in there, and a fallback built elsewhere would sanitize with no policy — DOMPurify would then claim the `dompurify` name on our instance, which is the collision #1232 exists to prevent. That case is covered and mutation-checked there.

**Explicit registration is supported, per your first suggested remedy.** `EditorSanitizer` is now exported from `src/index.js`, and `allowedElements` defaults to `[]`, so a consumer can call `EditorSanitizer.register(editor)` for just its importable tags or pass its own allowlist. It needs only the Lexical editor, which is what a node consumer already has.

**Covered on the exported-node path**, which nothing exercised: `test/javascript/unit/editor/attachments/exported_node_sanitization.test.js` re-inflates attachment content through the node in a standalone `createEditor`, with no editor registered. Three cases — the default is DOMPurify's rather than empty (and still strips `<script>`); the result is unchanged after another editor registers; and registering narrows it as expected.

Mutation-checked in both directions, and each mutation fails exactly one test and only one:

- restoring `buildConfig([])` → fails "re-inflates under DOMPurify's default policy when no editor registered"
- restoring `this.#fallback = sanitizer` in `register()` → fails "re-inflates the same way after another editor registers"

158 unit tests pass on this branch (163 at #1232, 167 at #1226 above it), and rollup still reports no circular import.

jorgemanrubia:
🤖 **Force-pushed: restacked onto the current `attachment-content-image-attributes`, plus one review fix.**

**Restack.** This branch was two commits behind #1230 — the test relocations https://github.com/basecamp/lexxy/commit/4a4a3898f29209a7c14959fe04c06471c985312c (image attributes test under `unit/sanitization`) and https://github.com/basecamp/lexxy/commit/29adc3651bcd53e67546dc5dff6bb5b1a8e5410a (attachment system tests under `system/attachments`) — and one below that, since #1230 was itself stale. Rebased with no conflicts.

The rename in this PR (`unit/editor/sanitized.test.js` → `unit/editor/sanitizer.test.js`) and #1230's file moves touch different files, so nothing collided; checked explicitly for the duplicate-file failure mode a rename-versus-move can produce, and there is exactly one copy of every test with no orphan at an old path. The only content difference between the old tip and the new one is #1230's two relocations and #1229's CI commit.

**Review fix**, in https://github.com/basecamp/lexxy/commit/570d604b8d546ab359fb910c0bbacf0912d4eaa4: the fallback sanitizer was both non-deterministic and over-strict for a consumer using the exported `CustomActionTextAttachmentNode` outside a Lexxy editor. It reused whichever sanitizer registered most recently — last-editor-wins, reintroduced on the one path with no editor to key on — and its config was `buildConfig([])`, which allows nothing.

It is now fixed and never reassigned, and declares no allowlist so DOMPurify's own default policy applies, which is what a standalone consumer had before this series. `EditorSanitizer` is exported from `src/index.js` and `register()`'s `allowedElements` defaults to `[]`, so a consumer that wants its own allowlist has a supported route. Covered by a new test on the exported-node path and mutation-checked in both directions; details in the thread reply.

158 unit tests pass, no circular-import warning from rollup. The full sweep — Playwright chromium, the Ruby system suite — ran at the top of the stack on #1226, where everything below it is included: 641 chromium tests pass (`leak.test.js` flakes under parallel load on the base too), 76 system tests, 167 unit tests. WebKit could not be run: `browserType.launch` fails here for a missing `libicu74`, so there is no WebKit coverage in these results.

— @jorgemanrubia

jorgemanrubia:
🤖 Reviewed the two findings raised against this branch. **No code changed here** —
one was already fixed on the branch, the other was a body accuracy point. Body
updated; nothing pushed.

**The fallback path was the one real test gap — and it's already covered** by
[`570d604b`](https://github.com/basecamp/lexxy/commit/570d604b8d546ab359fb910c0bbacf0912d4eaa4).
The review predated that commit. `test/javascript/unit/editor/attachments/exported_node_sanitization.test.js`
re-inflates attachment content through the exported node in a standalone Lexical
editor with no editor registered, asserts the deterministic default rather than
empty output, and asserts the result **does not change** after another editor
registers. That is exactly the pair the finding asked for, so I re-ran the mutation
checks rather than taking the commit message's word for it:

- restore the fallback being reassigned on every `register()` → *"re-inflates the
  same way after another editor registers"* fails, and only that one
- restore the fallback to `buildConfig([])` → *"re-inflates under DOMPurify's
  default policy when no editor registered"* fails, and only that one

**"Consumer action: none here" was false for npm consumers under enforced Trusted
Types.** This PR creates a second DOMPurify instance, whose lazily created policy
contends for the `dompurify` name the host's instance already claimed, and the
`lexxy`-named policy that resolves it arrives in #1232 — so this must not be merged
and soaked alone for those consumers. The body now says so and names #1232 (it read
`#trusted-types`, unlinked, in two places).

While there: the body now records that both the policy-name collision and the
original "we disarmed the host app's sanitizer" fault are **npm-consumers-only**.
`rollup.config.mjs` leaves `dompurify` out of `external`, so the Rails asset build
inlines its own copy — no shared module instance to reach, nothing to disarm — and
that inlined singleton creates its policy lazily and never sanitizes, so it never
contests the name either. `rollup.config.npm.mjs` marks `dompurify` external, which
is where the sharing comes from. Verified both configs.

Everything else pressed and came back sound: registry lifetime, registration
ordering, cross-editor influence, `get value()` memoisation.

**Verified:** 158 vitest / 13 `bin/rails test` / 75 runs, 287 assertions
`bin/rails test:system` / 642 Playwright chromium, all green. (The first
`test:system` run threw 9 Chrome teardown errors on `window_handles`; clean on
re-run, environment noise.) WebKit cannot launch on this machine (`libicu74`
missing), so it is untested here, not passing.

PR #1242, opened 2026-08-17
Pasted plain text from Notepad turns into a code block

jorgemanrubia:
Pasting plain text copied from Notepad into a Basecamp editor mangles it two ways: an indented outline turns into a syntax-highlighted code block, and every line break doubles.

Notepad puts `text/plain` on the clipboard with CRLF line endings and no `text/html` flavor at all, so the paste goes through `#pasteMarkdown`.

**Indentation read as code.** Markdown treats any line indented by a tab or four spaces as a code block, and `marked` obliged — an outline of indented notes came back as `<pre><code>`. Pasted plain text is full of incidental indentation (outlines, notes, logs) and nobody indents a note meaning "this is code"; they fence it. So this drops `marked`'s indented-code tokenizer for pasted text. Fenced blocks and indented nested lists still convert.

**CRLF doubled every line break.** `Contents#insertText` — the verbatim path taken when the text carries no markdown — split on `\n` alone, leaving a stray `\r` at the end of each text node that rendered as a second line break. It now splits on all three line-ending forms.

Reported against Ping/check-in answers on Windows 10 / Chrome, where the customer's indented notes pasted as a code block and switching the block's language to Plain Text didn't bring them back.

### DIFF
--- a/src/editor/clipboard.js
+++ b/src/editor/clipboard.js
@@ -1,4 +1,4 @@
-import { marked } from "marked"
+import { Marked } from "marked"
 import { isAutolinkableURL } from "../helpers/string_helper"
 import { nextFrame } from "../helpers/timing_helper"
 import { addBlockSpacing, dispatch, parseHtml } from "../helpers/html_helper"
@@ -8,6 +8,12 @@ import { $insertDataTransferForRichText } from "@lexical/clipboard"
 import { $createLinkNode, $isLinkNode, $toggleLink } from "@lexical/link"
 import { ListenerBin } from "../helpers/listener_helper"
 import NodeInserter from "./contents/node_inserter"
+
+// Markdown reads any line indented by a tab or four spaces as a code block. Pasted
+// plain text is full of incidental indentation — outlines, notes, logs — and nobody
+// indents a note meaning "this is code"; they fence it. Keep fences, drop the
+// indentation rule.
+const pastedMarkdown = new Marked({ breaks: true }).use({ tokenizer: { code: () => undefined } })
 
 export default class Clipboard {
   #listeners = new ListenerBin()
@@ -195,7 +201,7 @@ export default class Clipboard {
   }
 
   #pasteMarkdown(text) {
-    const html = marked(text, { breaks: true })
+    const html = pastedMarkdown.parse(text)
     const doc = parseHtml(html)
 
     if (this.#isPlainTextWithoutMarkdown(doc)) {
--- a/src/editor/contents.js
+++ b/src/editor/contents.js
@@ -52,7 +52,7 @@ export default class Contents {
   insertText(text, { tag } = {}) {
     this.editor.update(() => {
       const paragraph = $createParagraphNode()
-      text.split("\n").forEach((line, index) => {
+      text.split(/\r\n|\r|\n/).forEach((line, index) => {
         if (index > 0) paragraph.append($createLineBreakNode())
         paragraph.append($createTextNode(line))
       })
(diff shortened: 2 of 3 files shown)


jorgemanrubia:
🤖 ## Manual Validation: Validated

Ran `manually-validate-fix` against this PR — one Selenium script, two environments, same steps in both.

**Note on the skill:** its production URL, `https://basecamp.github.io/lexxy/try-it.html`, now 301s to `lexxy.dev`, where that path 404s. I used `https://lexxy.dev/sandbox/`, which loads `@37signals/lexxy@latest` from esm.sh — currently 0.9.29, the released build without this fix. Worth updating in the skill.

### How the paste was performed

Two mechanisms, so the reproduction doesn't rest on a synthetic event alone:

1. **Real clipboard + real Ctrl+V** — `navigator.clipboard.write()` with a single `text/plain` blob, then a genuine `Ctrl+V` keystroke into the editor. Confirmed via a read-only capture-phase `paste` listener that the editor received `types: ["text/plain"]` and the CRLF intact (`"\tBACKUPS\r\n\t\tContinued optimizations.\r\n…"`) — exactly the clipboard shape Notepad produces.
2. **Constructed Windows clipboard** — a `ClipboardEvent` carrying `text/plain` only with CRLF, the same mechanism the repo's own browser suite uses.

Both mechanisms gave identical results in both environments.

### Production (https://lexxy.dev/sandbox/, lexxy 0.9.29) — bug reproduces

- Pasted the customer's tab-indented outline. The first two lines came back as `<pre data-language="javascript">` with the JavaScript language picker showing, matching the customer's screenshot; the remaining two lines fell out into a paragraph:
  `<pre data-language="javascript">BACKUPS<br><span>	</span>Continued optimizations.<br></pre><p>CYBERSECURITY<br>HIBP</p>`
- Pasted `first line\r\nsecond line`. Got `<p>first line<br><br>second line</p>` — the doubled break, visible as a blank line in the editor.

### Local (`fix-indented-plain-text-paste`) — bug does not reproduce

- Same outline: `<p>	BACKUPS<br>		Continued optimizations.<br>CYBERSECURITY<br>	HIBP</p>` — one paragraph, indentation intact, no code block.
- Same CRLF pair: `<p>first line<br>second line</p>` — a single break.

### Regression guards — identical in both environments

Ran these to make sure the markdown path wasn't over-trimmed and neighbouring paste paths were untouched. All produced byte-identical output on production and on the branch:

| Paste | Result (both) |
|---|---|
| Fenced code block ` ```\r\nputs 'hello'\r\n``` ` | still a code block |
| Indented nested list `- one\r\n    - nested\r\n- two` | still `ul > li > ul > li` |
| `# Title` + `**bold**` | still `<h1>` + `<strong>` |
| LF-only plain text | single `<br>`, unchanged |
| Word/Outlook clipboard (`text/html` with `mso-list`, marker spans, `MsoListParagraph`) | still rebuilt into a real nested list |

### Action Text round trip (local)

Pasted the outline into `/posts/new`, saved, read the rendered post, reopened for edit. No code block at any stage. Worth knowing: once saved, the leading indentation collapses in the posted view and on re-edit, because a paragraph doesn't get `white-space: pre`. That is the tradeoff already noted on the bug — the content survives as text rather than as code.

### Other checks

- `bin/ci` run locally on the branch: green apart from three local-environment issues, none of them a delta from this PR. `zizmor` isn't installed on this machine and `playwright install --with-deps` wants a sudo password. One browser test, `plain_text_paths › preserves runs of consecutive spaces`, fails under the full single-worker run with `Received: ""` — the paste never lands. I ran the same full suite on `main` and it fails identically there, so it is a pre-existing local flake, not a regression from this change; it passes in isolation on the branch and on CI across all three browsers.
- All 12 GitHub checks green, including browser tests on chromium, firefox, and webkit.

Verdict: **Validated.** Marking ready for review.

